PR #589, opened 2015-05-01
Minor changes from devel-mark

marksantcroos:
Hmmm, additional commits to the branch automagically gets added to the PR.

### DIFF
--- a/src/radical/pilot/agent/radical-pilot-agent-multicore.py
+++ b/src/radical/pilot/agent/radical-pilot-agent-multicore.py
@@ -1417,7 +1417,6 @@ class LaunchMethodMPIRUN(LaunchMethod):
     def _configure(self):
         self.launch_command = self._find_executable([
             'mpirun',            # General case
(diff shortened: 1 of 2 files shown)


andre-merzky:
We are mixing job submission and file system endpoint schemas -- but I don't have a better proposal which would not increase complexity.  Lets change that when it becomes a problem...

andre-merzky:
Thanks!

marksantcroos:
> We are mixing job submission and file system endpoint schemas

Isnt that a feature instead of a problem?

andre-merzky:
Yes, maybe -- until I want to add gsissh & go, then it will become confusing... gsigo? ;)

marksantcroos:
I would probably suggest `gogogadgetsftp`, but otherwise I agree.

The ultimate solution once this becomes a problem is probably "simple", exposing them separately.
Which is only a small change in the resource config and in most cases people will stick with the defaults anyway.

andre-merzky:
Agree - we can do that once it's worthwhile to do so...

Issue #572, opened 2015-04-21
LSDMap stage fails on stampede -- possibly due to compiler mismatch during mpirun_sh


vivek-bala:
LSDMap is installed using the intel/14* compiler (and cannot be run with intel/13* due to some specific python error). RP executes mpirun_sh from the intel/13* compiler.

The shell script looks like,

```
# Change to working directory for unit
cd /work/02734/vivek91/radical.pilot.sandbox/rp.session.ip-10-184-31-85.vivek.016546.0000-pilot.0000/unit.000017
# Pre-exec commands
module load gromacs
python pre_analyze.py 16 tmp.gro
echo 2 | trjconv -f tmp.gro -s tmp.gro -o tmpha.gro
module load -intel +intel/14.0.1.106
export PYTHONPATH=/home1/03036/jp43/.local/lib/python2.7/site-packages
module load python/2.7.6
export PATH=/home1/03036/jp43/.local/bin:$PATH
# The command to run
/opt/apps/intel13/mvapich2/1.9/bin/mpirun_rsh -np 16 c558-802 c558-802 c558-802 c558-802 c558-802 c558-802 c558-802 c558-802 c558-802 c558-802 c558-802 c558-802 c558-802 c558-802 c558-802 c558-802 c558-802 c558-802 c558-802 c558-802 c558-802 c558-802 c558-802 c558-802 c558-802 c558-802 c558-802 c558-802 c558-802 c558-802 c558-802 c558-802 LD_LIBRARY_PATH=$LD_LIBRARY_PATH PATH=$PATH PYTHONPATH=$PYTHONPATH /opt/apps/intel14/mvapich2_2_0/python/2.7.6/lib/python2.7/site-packages/mpi4py/bin/python-mpi "lsdm.py" "-f" "config.ini" "-c" "tmpha.gro" "-n" "neighbors.nn" "-w" "weight.w"
```
The corresponding STDERR,

```
...
...
..
The following have been reloaded with a version change:
  1) intel/13.0.2.146 => intel/14.0.1.106  3) python/2.7.9 => python/2.7.6
  2) mvapich2/1.9a2 => mvapich2/2.0b

[c558-802.stampede.tacc.utexas.edu:mpispawn_0][spawn_processes] Failed to execvp() 'c558-802':  (2)
[c558-802.stampede.tacc.utexas.edu:mpispawn_0][spawn_processes] Failed to execvp() 'c558-802':  (2)
[...]

PR #590, opened 2015-05-01
Fix for #572

marksantcroos:
Didn't intend to include the other two commits on this branch, so that has been reverted.


andre-merzky:
The test is supposed to run with 16 cores on the headnode of stampede (schema=ssh)?  I guess the CU workload is small enough to do so?

marksantcroos:
> The test is supposed to run with 16 cores on the headnode of stampede (schema=ssh)?

Can I assume you wrote this in a hurry? :)

andre-merzky:
No, actually not - I looked back and forth a couple of times.  That does not mean I didn't miss anything obviously :P  Was looking out for SAGA schemas I guess...

PR #599, opened 2015-05-04
Fix/pip unalias

andre-merzky:
also removes an unneccessary 'unset' and fixes some doc strings.  

Tests: http://ci.radical-project.org/job/radical.pilot.devel/branch/hotfix%2F587/

### DIFF
--- a/src/radical/pilot/bootstrapper/default_bootstrapper.sh
+++ b/src/radical/pilot/bootstrapper/default_bootstrapper.sh
@@ -568,16 +568,11 @@ virtenv_activate()
     fi
     # NOTE: some resources define a function pip() to implement the same cacert 
     #       fix we do above.  On some machines, that is broken (hello archer), 
-    #       thus we undefine that function here.  To be on the save side, we 
-    #       also undefine aliases
-    unalias  pip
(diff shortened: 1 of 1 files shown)


marksantcroos:
Does the test actually remove the existing ve? Otherwise its of little value.

andre-merzky:
Each branch has its own workspace and ve, so yes, that is a new ve.

marksantcroos:
Obviously I'm talking about the remote ve.

marksantcroos:
And can we get rid of the code duplication in the process or you dont feel comfortable doing that now?

andre-merzky:
Ah, sorry, got it...  Well, pip is also used in the update mode, and when the radical stack is installed.  That holds at least for some of the target resources (like, local).  For the others the tests show at least that the branch still works with the current configs...

code duplication: that is the 'rehash' on pip which is done in several places?  We could do that in a method, sure...

marksantcroos:
Thanks for the de-dup.
While reviewing, I found that virtenv_activate is always called with an argument, but that doesnt seem to be used?

marksantcroos:
Passes visual inspection and tests.

PR #602, opened 2015-05-04
Feature/schedule queue protocol

andre-merzky:
This commit set cleans the communication between different agent components and the agent scheduler.   It is otherwise semantically void.

### DIFF
--- a/src/radical/pilot/agent/radical-pilot-agent-multicore.py
+++ b/src/radical/pilot/agent/radical-pilot-agent-multicore.py
@@ -708,6 +708,8 @@ class Scheduler(threading.Thread):
 
                 if command == COMMAND_RESCHEDULE:
 
+                    # reschedule is done over all units in the wait queue
+                    assert (cu == None) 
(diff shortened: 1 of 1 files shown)


andre-merzky:
Tests: http://ci.radical-project.org/job/radical.pilot.devel/branch/feature%2Fschedule-queue-protocol/

src/radical/pilot/agent/radical-pilot-agent-multicore.py
```diff
@@ -5283,7 +5303,7 @@ def run(self):
 
         # to make sure that threads are not stuck waiting on a queue, we send
         # a signal on each queue
-        self._schedule_queue.put (None)
+        self._schedule_queue.put (COMMAND_WAKEUP)
```
marksantcroos: The wakeup command is not processed at the other end.

andre-merzky: Ah, nice catch!  Fixed.

PR #603, opened 2015-05-04
fix agent argument handling for empty parameters

andre-merzky:
very simple fix for a problem which got triggered by some swift integration test.

### DIFF
--- a/src/radical/pilot/agent/radical-pilot-agent-multicore.py
+++ b/src/radical/pilot/agent/radical-pilot-agent-multicore.py
@@ -3939,6 +3939,9 @@ class ExecWorker_SHELL(ExecWorker):
             ret = list()
             for arg in args :
 
+                if not arg:
+                    continue
+
                 # if string is between outer single quotes,
                 #    pass it as is.
                 # if string is between outer double quotes,


marksantcroos:
Whats the condition, `for arg in [None]:`?

andre-merzky:
https://github.com/radical-cybertools/radical.pilot/blob/fix/agent-argument-handling-2/src/radical/pilot/agent/radical-pilot-agent-multicore.py#L3995 makes sure that args is not `None`.  The added check should work ok for `args=[None]` -- that would be ignored, and an empty arg list would be returned, which later results in an empty arg string (`''`).

andre-merzky:
FWIW, there is no check if `args` is a list -- that should be caught by the attrib interface on the CU description on the submission side...

marksantcroos:
That was my question, how does the [None] end up there? User specified?

andre-merzky:
Yes, user specified -- actually as `['-cf', '', '-par', '', ...]` -- the `arg[0]` stumbled over that...

marksantcroos:
That doesn't necessarily give the results that the user expects then, does it?

marksantcroos:
But also not sure what was intended.

andre-merzky:
The former popen spawner handled it the same way: https://github.com/radical-cybertools/radical.pilot/blob/master/src/radical/pilot/agent/radical-pilot-agent-multicore.py#L3525

 We probably should add some documentation, but the policy really is: 'handle arguments as strings, if the arg value starts and ends with single quote, then pass it in single quote, otherwise pass it in double quotes.  So:

`args = ['a', '"b"', "c", "'d'", '', ""] ->  "a" "b" "c" 'd' '' ''`

marksantcroos:
I mean, is there any use for "empty" arguments, and if not, shouldnt we warn and/or fail early?

andre-merzky:
That actually changes what the executable sees as positional parameters.  If that is good practice is a different question, but I don't want to judge user code on that level.  It is technically legal to specify empty parameters:

```
$ echo test | cut -f 1 -d '' | grep ''
test
```

marksantcroos:
Right, but I do we maintain that positionality now?

andre-merzky:
yes -- an empty arg ends up as `''` on the command line.

marksantcroos:
Ok, cool.

PR #604, opened 2015-05-04
Fix mysterious stalling local tests.

marksantcroos:


andre-merzky:
The jenkins host has only one core - that is why we had it at '1' IIRC.  It will work with `4` now, because we set debug, but it might make Jenkins slow(er)...  Anyway, will merge, if it becomes unbearable we should increase the VM size...

PR #605, opened 2015-05-04
add a comment on the archer config change

andre-merzky:
also reset default target for example script


marksantcroos:
> also reset default target for example script

The name of the script is not very appropriate anymore is it? Do we not need a local and remote script then?

andre-merzky:
No, we don't need separate scripts anymore.  Changing that needs updates in the documentation though, so a quick delete+rename is not sufficient.

PR #610, opened 2015-05-05
disable agent oversubscription on backfilling scheduler

andre-merzky:
that optimization pessimizes some use cases, amongst others Matteos experiments.  This needs to be rediscussed after refactoring.

### DIFF
--- a/src/radical/pilot/scheduler/backfilling.py
+++ b/src/radical/pilot/scheduler/backfilling.py
@@ -21,7 +21,7 @@ from radical.pilot.states              import *
 
 # to reduce roundtrips, we can oversubscribe a pilot, and schedule more units
 # than it can immediately execute.  Value is in %.
-OVERSUBSCRIPTION_RATE = 200
+OVERSUBSCRIPTION_RATE = 0
 
 # -----------------------------------------------------------------------------
 # 


marksantcroos:
I didn't even know we had that feature. Was that in response to something or pro-active?
I will have to look up the code that uses this value to see whether the commit is ok.

marksantcroos:
It seems the whole code is disabled, but one can only guess as I could say nasty things about documentation and readability there :-)

andre-merzky:
It is used here: https://github.com/radical-cybertools/radical.pilot/blob/master/src/radical/pilot/scheduler/backfilling.py#L288

It was in response to the backfilling scheduler being slow, as it suffers from callback latency.  That is caused by the roundtrip obviously, but also by the poor time resolution of the current cu state pull implementation on the client end.  This pre-fetching counters that to some extent, as it overlaps the callback latency with agent activity.

It obviously works great for a single pilot, and performance gets close to direct scheduler in our test and example cases (modulo slow-start).  It screws up when the prefetch-size is close to the total workload size *and* multiple pilots are involved *and* the second pilot comes up after the prefetch has assigned to the first pilot.  Which is exactly what Matteo hits (his workload fits exactly into the prefetch-size).

andre-merzky:
> It seems the whole code is disabled, but one can only guess as I could say nasty things about documentation and readability there :-)

I am not sure I understand that -- the if gets passed over when the value is set to `0` -- but I would expect the oversubscription to be passed when oversubscription is set to `0%`.  What am I missing?

marksantcroos:
That a non-local variable is set for undocumented reasons, so I dont know whether there are side-effects. (In this case the side-effects of not setting it at all)

andre-merzky:
So you are saying the comment in L22 is insufficient, and there should be a similar comment where the var is used?

marksantcroos:
No, I was referring to `self.pilots[pid]['over']` (and `'caps'` for that matter).

andre-merzky:
`caps` is defined -- the comment above the block of definitions says in the first line `get initial information about the pilot capabilities`.  

`over` can be a local var, that is true.  I expected that to be something which eventually becomes dynamic -- at this point its superfluous.

I still don't really understand what you mean with
> It seems the whole code is disabled, but one can only guess as I could say nasty things about documentation and readability there :-)
What code is disabled?  Is it really just that this var is kept in the state dict and not a local var?  I'll fix that then.

andre-merzky:
Pushed.  I added that to the existin pull request, for simplicity, even if it is off-topic...

PR #611, opened 2015-05-05
add slots to unit frames

### LINKED ISSUE #548: cu slot information should be added to unit dataframe
Also, the raw data in the database looks wrong.

andre-merzky:
Fixes #548

### DIFF
--- a/src/radical/pilot/session.py
+++ b/src/radical/pilot/session.py
@@ -261,7 +261,7 @@ class Session (saga.Session, Object):
 
     #---------------------------------------------------------------------------
     #
-    def close(self, cleanup=True, terminate=True, delete=None):
+    def close(self, cleanup=None, terminate=None, delete=None):
         """Closes the session.
 
         All subsequent attempts access objects attached to the session will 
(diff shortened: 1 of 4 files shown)


src/radical/pilot/session.py
```diff
@@ -285,11 +285,15 @@ def close(self, cleanup=True, terminate=True, delete=None):
             logger.error("Session object already closed.")
             return
 
+        # set defaults
+        if cleanup   == None: cleanup   = True
+        if terminate == None: terminate = True
+
```
marksantcroos: What problem is this solving?

andre-merzky: It's not solving any problem - I find it easier to read.  That is the way python needs default parameter values for mutable types, and I find it easier to use the same way for all types really, then one way for one type class, another way for another type class (specifically since python does not even define the types, and we have many places where we pass either an instance or a list).

andre-merzky: FWIW, I can stop converting those things, no need to start another style discussion...

marksantcroos:
Thanks

src/radical/pilot/session.py
marksantcroos: I see the argument of the difference with mutable parameters, although I'm happy to consider those exceptions.

Anyway, I don't feel strong about it, I was really wondering whether there was a subtle difference in semantics, which you have clarified to be not the case.

PR #618, opened 2015-05-05
SuperMUC resource config changes.

marksantcroos:
- Change mongodb tunnel endpoint.
  As connections can procede from the login nodes to the mongodb, we no longer
  need a static tunnel. (I'm pretty sure I made this mod before ...)
- Create and point to new populated virtenv.
  This one has pymongo==2.8 apache-libcloud colorama python-hostlist,
  the RADICAL stack is flown in real-time.
- The latter is enforced by setting rp_version to "debug".
- Remove the sourcing of the VE from the pre_bootstrap .. don't know why it was
  there in the first place ...


andre-merzky:
Thanks!

PR #635, opened 2015-06-03
Really disable timeout mechanism.

marksantcroos:
Given that this was not used anyway, and causes problems on BW, there is no
reason not to do this.

Symptoms seen are that the timeout that sends the alarm causes the script to
exit. Which make me think that that my '/bin/sleep' might do something funky
with the SIGALRM handler.

Ultimately it would be good to better understand it, but this is not the time
and place.


andre-merzky:
Thanks!

I'll check  with SHantenu on BW access -- could not manage to reproduce the problem at all, but would really be interested in finding out what happens.  The code depends on signals in several places, I don't like any funkiness there.  Also, we have been discussing auto-reconnect for pty lately, and timeouts would start to make sense again in that context (to avoid zombies).

PR #643, opened 2015-06-21
Devel andre

andre-merzky:
A couple of smaller commits begin to pile up, so I think its time for a merge before things pile up more.  

BTW: this is the only drawback in our current pull-request based merge model: you have to merge all commits from a branch or none, even the commits you pushed after the pull request was created.  Either that, or one needs to sort the commits manually into different branches, which is somewhat of an overhead...


marksantcroos:
> or one needs to sort the commits manually into different branches, which is somewhat of an overhead...

Thats a feature, not a bug. You can't have your cake and eat it.

PR #654, opened 2015-07-19
Feature/prepost timing

andre-merzky:
This also adds an option to `prof` to specify a timestamp, so that we can add timings which are not collected life.


examples/getting_started_local.py
```diff
@@ -150,13 +150,15 @@ def wait_queue_size_cb(umgr, wait_queue_size):
...
+            cud.pre_exec      = ["sleep 5"]
+            cud.post_exec     = ["sleep 5"]
             cud.arguments     = ["5"]
             cud.cores         = 1
     
             cuds.append(cud)
     
         # Submit the previously created ComputeUnit descriptions to the
-        # PilotManager. This will trigger the selected scheduler to start
+        # PilotManager. This will triggee the selected scheduler to start
```
marksantcroos: typo

src/radical/pilot/agent/radical-pilot-agent-multicore.py
```diff
@@ -3768,6 +3784,16 @@ def _check_running(self):
                 self._cus_to_watch.remove(cu)
                 self._schedule_queue.put ([COMMAND_UNSCHEDULE, cu])
 
+                if os.path.isfile("%s/PROF" % _cu['workdir']):
+                    with open("%s/PROF" % _cu['workdir'], 'r') as prof_f:
```
marksantcroos: Can this contain unicode?

marksantcroos:
Looks good. If you address the two comments I'll merge.

src/radical/pilot/agent/radical-pilot-agent-multicore.py
andre-merzky: No -- its the result of a shell echo which cannot  contain unicode, AFAICS.

marksantcroos: So the unicode stuff below can go?

andre-merzky: I see.   Yes, it can.  I  copied that from somewhere...  Thanks.

andre-merzky: done

marksantcroos:
I had to make a commit to have the agent not crash on me. How stable is this supposed to be? It seems a bit haphazard ...

andre-merzky:
Thanks, that change certainly makes sense.  Haphazard it might be to some extent, but it is also relatively simple.  If we decide at some point we don't want  this timed anymore, it should be easy to remove.

marksantcroos:
Now lets audit this and all values should be ~0!

PR #655, opened 2015-07-20
Feature/generic staging

marksantcroos:
These changes generalize the file staging which allows it to work with non-ssh transfers, e.g. Globus Online, and it also leads the way for OSG development.


src/radical/pilot/configs/nersc.json
```diff
@@ -76,7 +80,7 @@
         "default_remote_workdir"      : "$SCRATCH",
         "valid_roots"                 : ["/global", "/scratch1", "/scratch2"],
         "bootstrapper"                : "default_bootstrapper.sh",
-        "rp_version"                  : "debug",
+        "rp_version"                  : "local",
```
andre-merzky: Is this intentional?

src/radical/pilot/controller/input_file_transfer_worker.py
```diff
@@ -174,22 +171,16 @@ def run(self):
...
                                 else:
                                     copy_flags = 0
 
-                                try :
-                                    input_file.copy(target, flags=copy_flags)
-                                except Exception as e :
-                                    logger.exception (e)
-                                input_file.close()
+                                try:
+                                    saga_dir.copy(input_file_url, target, flags=copy_flags)
```
andre-merzky: input staging now is done on the saga_dir object -- shouldn't the same be applied to the output staging, or is that not needed there / already there differently?

andre-merzky:
I did not test the branch, but otherwise this looks ok to me (modulo the comment on output-FTW.

src/radical/pilot/configs/nersc.json
marksantcroos: As discussed, sort of, as its the default anyway.

andre-merzky:
Many thanks, Mark! :D

marksantcroos:
Do I get a cookie now? :)

PR #659, opened 2015-07-20
Complete support for Comet

marksantcroos:


andre-merzky:
Thanks for diving down the rabbit hole :)

PR #666, opened 2015-07-27
Disable writing to /tmp.

marksantcroos:
This should not be enabled until it is thought out properly.


andre-merzky:
Fair point about the hop, obviously.  One obvious approach is to only use /tmp/ on non-hopping launch methods I guess...

andre-merzky:
Fair point about the hop, obviously.  One obvious approach is to only use /tmp/ on non-hopping launch methods I guess...

marksantcroos:
The location is (currently) not launch-method related though.

PR #668, opened 2015-07-28
Move to new version of Intel compiler and MPI lib.

marksantcroos:
Among other things this fixes XSEDE ticket 33533.


andre-merzky:
Excellent!

PR #675, opened 2015-08-06
Feature/remote orte

andre-merzky:
This pull request covers a number of configuration and bootstrap changes, which are preparatory for te planned agent splitup.


marksantcroos:
I think we concluded we could kill this branch and PR, right?

andre-merzky:
Let me check if anything is missing in split_agent -- but I think not.

andre-merzky:
yeah, I think its ok to remove it, and to do the PR from split_agent to devel...

marksantcroos:
Ack. I don't think I have objection against doing that (now).

PR #683, opened 2015-08-15
printf vs. echo  fixes

andre-merzky:
`printf` interprets `%` which is not always wanted, for example when a shell script is passed as argument.    Echo is not always interpreting `\n` correctly, thus equally corrupting scripts passed as arguments.  We thus wrap `printf` in a function which makes sure that `%` are escaped correctly, thus maintaining `\n` interpretation, and avoiding the printf expansion.

### DIFF
--- a/src/radical/pilot/agent/radical-pilot-spawner.sh
+++ b/src/radical/pilot/agent/radical-pilot-spawner.sh
@@ -46,6 +46,15 @@ fi
 
 # --------------------------------------------------------------------
 #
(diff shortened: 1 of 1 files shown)


marksantcroos:
This is the same fix I verified in a different branch already, right?

andre-merzky:
Yes, I think so.

marksantcroos:
Ok, then I consider them tested.

marksantcroos:
Thanks

PR #684, opened 2015-08-15
Feature/session recording

andre-merzky:
When debugging RP applications, I would like to see the pilot and unit descriptions as they arrive on the RP layer.  This PR adds some lines of code which capture those and dump json files to a given dir (specified in `RADICAL_PILOT_RECORD_SESSION`).

Once one has that, one can now also read those json files and play them back via RP, like a session-recording and session-playback mechanism.  That has very limited capabilities for sure: time ordering or causal ordering is not captured, and most importantly, data for staging ops are not captured, but its still quite useful to inspect what is happening on the RP layer w/o the need to go through all the kernel configs and ENMD cluster somethings...

### DIFF
--- a/bin/radicalpilot-run-session
+++ b/bin/radicalpilot-run-session
@@ -1,0 +1,154 @@
+#!/usr/bin/env python
+
+import os
+import sys
+import glob
+import pprint
+import datetime
+import logging
+import pymongo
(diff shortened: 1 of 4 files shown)


marksantcroos:
Thanks

PR #687, opened 2015-08-18
Feature/agent split

andre-merzky:
This pull request is created to foster the discussion of the commit set which ultimately results in the agent component split-up, but which also touches numerous other parts of RP.

Note that due to the log and prof restructuring, this branch depends on the devel-andre branches in saga and utils.


andre-merzky:
wohoo! :)

PR #699, opened 2015-08-26
added agent_variables in example config

JensTimmerman:
I don't fully know what these variables are supposed to do, but it doesn't work without them.

### DIFF
--- a/docs/source/machconf.rst
+++ b/docs/source/machconf.rst
@@ -140,6 +140,9 @@ A configuration file has to be valid JSON. The structure is as follows:
                                              "source /home/hpc/pr87be/di29sut/pilotve/bin/activate"
                                             ],
             "valid_roots"                 : ["/home", "/gpfs/work", "/gpfs/scratch"],
+            "agent_type"                  : "multicore",
+            "agent_scheduler"             : "CONTINUOUS",
+            "agent_spawner"               : "POPEN",
             "pilot_agent"                 : "radical-pilot-agent-multicore.py"
         },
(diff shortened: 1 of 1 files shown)


marksantcroos:
Hi Jens, are you on a documentation quest or actually trying to use our software? ;)

We have been making some changes in this area lately and these are not yet represented in the documentation, thanks for pointing that out.

To answer your question:
- `"agent_type":  "multicore"`
 This specifies which agent to use, currently we only have one, the `multicore` agent.

- `"agent_scheduler": "CONTINUOUS"`
 This specifies the scheduler that the agent uses to map tasks to cores on the resource. We have a couple of them, but `CONTINUOUS` is the most generic one.

- `"agent_spawner": "POPEN"`
   This specifies the mechanism to execute tasks on the compute nodes. We currently have two `POPEN`, which was the original, and a more recent `SHELL` which uses a shell based execution "service".

Regards,

Mark

ps. Say hi to Fotis ;-)

JensTimmerman:
Hi Mark, 

I'm actually trying to use the software. By looking at the configuration files you ship I was able to figure out there are other settings that aren't documented yet, and started using them. 
It would be nice if they were all documented, but I noticed these were actually vital for your software to run.

I'll say hi when I see him ;-)

Regards,
Jens Timmerman

andre-merzky:
Mark, this one looks good actually -- will you merge it?

A.

marksantcroos:
Yes, its on my todo.

PR #700, opened 2015-08-26
Release/0.35

andre-merzky:


shantenujha:
http://radicalpilot.readthedocs.org/en/latest/ still points to 0.33

Can we automate so that documentation points to latest version?

andre-merzky:
Thanks for the reminder - the doc update has been pushed.

andre-merzky:
Mark -- our release procedure is silent about the fate of the release branches after release.  I would suggest we delete them, as I don't think we need more than the tags to base eventual hotfixes on a release?

marksantcroos:
Tags are enough, branches can go.

PR #701, opened 2015-08-31
Feature/collected procs

andre-merzky:
In some circumstances, popen left zombies behind.  The addition of the `wait()` calls should avoid that.


marksantcroos:
Hmmm, are you sure about this?
This work was done to fix something and it didn't do that. So does it serve other purposes?

andre-merzky:
I am relatively sure that the wait leads to a correct collection of the processes, even if those abort abnormally.  What still leads to zombies is an abnormal agent abort, as neither kill nor wait to the CUs is issued then.  At that point we will have to rely on the resource to clean up the process tree, ie. the session.  This should not be a problem for the batch systems we use, but is probably unclean (at best) for `fork://`.  But I don't think it is worth to clean up that specific case, really -- that is not the target use case for RP after all.

Or am I missing the point wrt. your question?

marksantcroos:
Did you actually encounter the problem that you are fixing?

andre-merzky:
I can't recall the exact context where this happened, but it popped up when a CU was killed by some signal.  The `poll()` figured out that the process was  gone, but did not reap the process entry -- specifically, returncode got set incorrectly.  

Thinking about it: this is probably also addressed by the popen fix we have elsewhere, which simply avoids that condition in the first place.  But either way, I checked the subprocess impl, and it seems that the required waitpid is always called on `proc.wait()`, so I am sure it will reap the process then, and it seems like the clean thing to do.  `waitpid()` on dead processes has no significant overhead, so I am not  concerned about performance of the additional call.

FWIW, in the cancel case, the `kill/wait` combo is also what python uses internally (https://github.com/python/cpython/blob/master/Lib/subprocess.py#L704)

andre-merzky:
Soo, whats up with this PR? :)  

Look at it this way: It ensures a `waitpid()` is called on popen'ed processes.  In the worst case this is not needed -- and will then result in a NOOP which is not costly (one syscall which does a process table lookup and returns nothing).

marksantcroos:
Happy to look after it once the other branches are merged.
And the worst case is that it introduces a regression. You are describing the best case.

src/radical/pilot/agent/radical-pilot-agent-multicore.py
```diff
@@ -3926,6 +3926,8 @@ def _check_running(self):
                     # We got a request to cancel this cu
                     action += 1
                     cu['proc'].kill()
+                    cu['proc'].wait() # make sure proc is collected
+
```
marksantcroos: So this one I can see happening.

src/radical/pilot/agent/radical-pilot-agent-multicore.py
```diff
@@ -3938,6 +3940,9 @@ def _check_running(self):
             else:
                 self._prof.prof('exec', msg='execution complete', uid=cu['_id'])
 
+                # make sure proc is collected
+                cu['proc'].wait()
```
marksantcroos: You think this can ever be required?

andre-merzky: yes, this is actually the one I stumbled over.  It is probably redundant to the popen fix now though,  granted.  This one does not fix the popen problem we had in its own, but made sure that the pid is collected.  

Since it is redundant, I don't mind removing it, but I would also not mind leaving it for its explicity.

marksantcroos: If this can actually be required after a successful poll then I'm fine with keeping it.

PR #702, opened 2015-08-31
Fix/issue 695

andre-merzky:
This PR applies the same error capture logic to the agent staging which is already in place for the umgr staging, making sure that the unit is failed immediately upon failing staging ops.


marksantcroos:
This one needs re-merging devel before it can be considered.

src/radical/pilot/agent/radical-pilot-agent-multicore.py
```diff
@@ -4849,53 +4850,60 @@ def work(self, cu):
...
+            self._log.exception("staging input failed -> unit failed")
+            staging_ok = False
 
-        self._prof.prof('log', msg="toward agent scheduling", uid=cu['_id'])
 
-      # self.advance(cu, rp.AGENT_SCHEDULING_PENDING, publish=True, push=True)
-        self.advance(cu, rp.ALLOCATING_PENDING, publish=True, push=True)
+        # Agent input staging is done (or failed)
+        if staging_ok:
+          # self.advance(cu, rp.AGENT_SCHEDULING_PENDING, publish=True, push=True)
```
marksantcroos: Whats up with this line?

andre-merzky: that's a preparation for the last set of pending state renames.  `ALLOCATING` is not yet in sync with the target state model.  There are some more, similar comments somewhere through the agent I believe...

PR #705, opened 2015-09-02
cleanup of agent termination

andre-merzky:


andre-merzky:
I think this should be considered for merge at this point.

The commit set intents to implement the following functionality:

(i) on component startup, each component sends an `alive` message on the command pubsub channel.  An agent will collect the `alive` messages for all components it owns, and will in turn then also send an `alive` message.  Agents will also collect `alive` message for all subagent they spawned, affectively leading to a startup barrier in `agent.0` which will only proceed execution once the complete sub-agent tree with all components  are found to be alive.

Note that no timeout mechanism is in currently in place, we may want to add one at some point.

Agent and component bootstrap times can vary  wildly on different machines, between different runs, and can extent to the order of minutes.

(ii) the main loop for all components is now excepting for runtime errors, and will send a `shutdown` command on the command pubsub channel if any is caught.  That command is interpreted by all components and agents, and will lead to an immediate agent shutdown.  No attempt is made to salvage units in other components, nor is there any attempt made to recover the failing components.

`agent.0` will attempt to update the pilot state in MongoDB, any unit state progression or rescheduling is up to the application's or unit manager's interpretation of the pilot state.

src/radical/pilot/utils/component.py
```diff
@@ -361,6 +371,11 @@ def declare_publisher(self, topic, pubsub):
         self._log.debug('declared publisher : %s : %s : %s' \
                 % (topic, pubsub, q.name))
 
+        # FIXME: I am not exactly sure why this is 'needed', but declaring
+        #        a published as above and then immediately publishing on that
+        #        channel seems to sometimes lead to a loss of messages.
+        time.sleep(1)
+
```
marksantcroos: I get a bit nervous when we add another 500 lines to fix synchronisation and then ultimately it only works if we add a sleep ...

marksantcroos:
Other than my comment which will likely not get addressed I would be ok with merging this after pre_bootstrap2 is merged.

src/radical/pilot/utils/component.py
andre-merzky: Yeah, me too.  Its a `FIXME` because I very much would like to understand it.  If you feel strong about it, I'll dig into it before merge -- but I am rather confident that no messages get lost at this point.

PR #706, opened 2015-09-04
Feature/plotting

marksantcroos:
Despite the name this turned out to be more of an IPDPS branch.
So agent profiling changes and general agent improvements towards support for Cray.


src/radical/pilot/agent/radical-pilot-agent-multicore.py
```diff
@@ -2038,7 +2045,7 @@ def construct_command(self, task_exec, task_args, task_numcores,
         hosts_string = ",".join([slot.split(':')[0] for slot in task_slots])
         export_vars  = ' '.join(['-x ' + var for var in self.EXPORT_ENV_VARIABLES if var in os.environ])
 
-        orte_command = '%s --hnp "%s" %s -np %s -host %s %s' % (
+        orte_command = '%s --debug-devel --hnp "%s" %s -np %s -host %s %s' % (
             self.launch_command, dvm_uri, export_vars, task_numcores, hosts_string, task_command)
```
andre-merzky: should we make the debug dependent on the pilot debug level (which is in `self._cfg['debug']`)?

marksantcroos: We could, although the debug level itself is no longer an option.

andre-merzky: debug is  synced with the debug level on module side:
https://github.com/radical-cybertools/radical.pilot/blob/devel/src/radical/pilot/controller/pilot_launcher_worker.py#L616

marksantcroos: Pfff https://github.com/radical-cybertools/radical.pilot/blob/devel/src/radical/pilot/bootstrapper/default_bootstrapper.sh#L1324

andre-merzky: Yeah, those should go into the SAGA job environment I guess...

PR #708, opened 2015-09-07
Feature/prebootstrap2

marksantcroos:
This is required to get ExecWorkers going on Cray compute nodes.


andre-merzky:
why not double quotes?  Because this is usually specified in a json?  I don't like that re reinterpret shell syntax and bend over backwards just to not escape quotes ... :/  Anyway, if you insist, then that's it...  Otherwise the PR looks fine, I'll give it a spin later today!

marksantcroos:
> why not double quotes?

What does this refer to now?

andre-merzky:
```
>> so, not using ' but " is not an option?
> Correct.
why not double quotes? 
```

marksantcroos:
To prevent (too) early expansion.

marksantcroos:
This is now implicitly merged via the de-tour.

PR #712, opened 2015-09-09
Feature/netifaces

marksantcroos:
This tries to tackle the ip/hostname troubles.
Hopefully the comments are self-explanatory.


marksantcroos:
Note that I broke this out of a different three, I hope it actually works in the context of devel.

src/radical/pilot/agent/radical-pilot-agent-multicore.py
```diff
@@ -2311,64 +2310,59 @@ def _configure(self):
...
-                s = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
-                s.setblocking(False)
-                s.settimeout(1.0)
-                s.connect(('8.8.8.8', 53))
-                return s.getsockname()[0]
-            except Exception as e:
-                if logger:
-                    logger.exception("use fallback IP detection (%s)" % e)
+        # Known intefaces in preferred order
+        sorted_preferred = ['ipogif0', 'eth0']
```
andre-merzky: this is quite some magic right here... :P

marksantcroos: You dont like the heuristic?

andre-merzky: I like them if there are no alternatives.  In this case, I like 'ipogif0' as config entry even better :)  Is that hard-coded heuristics then still useful (ie. does it apply on any other host)?

marksantcroos: If we only go for configuration, then I agree. Having some "eth0 is better than lo0" when a config entry is lacking is probably still a sane thing to do.

PR #715, opened 2015-09-10
More precise time stamps in profiling using gtod.

marksantcroos:
- Save some trees in the long run also!
- Also fixed the return value with SHELL spawner if there is post_exec, similar
  to a fix recently made to POPEN.


andre-merzky:
Thanks, looks good to me.  I won't be able to test today I think, but will merge ASAP...

andre-merzky:
Thanks!

PR #717, opened 2015-09-11
Feature/micro benchmarks

andre-merzky:
Hey Mark, 

this is another kitchen sink I'm afraid, it just grew to be that... :/  I can pluck it apart into different branches/PRs if its too unwieldy.  But if you want to go through it, you may want to look at the individual commits.  That is in total likely more lines, but should be easier to parse, as I tried to keep the commits small-scoped...

thanks!

PS.: please check specifically at 3c08458 !


marksantcroos:
Thanks!

PR #720, opened 2015-09-13
fix hanging pilot bootstrap

andre-merzky:
This PR goes toward `feature/micro_benchmarks`, but it may make sense to delay it and then later merge directly to devel.


marksantcroos:
Works on Titan too.

andre-merzky:
wohoo!

PR #724, opened 2015-09-15
Feature/local cpn

andre-merzky:
setting `cores_pre_node=2` in the localhost config now results in

```
agent.0.bootstrap_3.log:2015-09-15 15:27:05,983: agent.0.bootstrap_3 : MainProcess : MainThread : \
DEBUG   : configure localhost to have 4 nodes a 2 cores
```

### DIFF
--- a/src/radical/pilot/agent/radical-pilot-agent-multicore.py
+++ b/src/radical/pilot/agent/radical-pilot-agent-multicore.py
@@ -3483,10 +3483,17 @@ class ForkLRMS(LRMS):
                 self._log.warn("more cores available: using requested %d instead of available %d.",
                         selected_cpus, detected_cpus)
 
-        self.node_list = list()
-        for i in range(selected_cpus):
+        # if cores_per_node is set in the agent config, we slice the number of
+        # cores into that many virtual nodes.  cpn defaults to selected_cpus,
+        # to preserve the previous behavior.
+        self.cores_per_node = self._cfg.get('cores_per_node', selected_cpus)
+        requested_nodes = int(math.ceil(float(selected_cpus) / float(self.cores_per_node)))
+        self.node_list  = list()
+        for i in range(requested_nodes):
             self.node_list.append("localhost")
-        self.cores_per_node = 1
+
+        self._log.debug('configure localhost to have %s nodes a %s cores',
+                len(self.node_list), self.cores_per_node)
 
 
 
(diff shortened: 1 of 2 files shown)


src/radical/pilot/agent/radical-pilot-agent-multicore.py
```diff
@@ -3483,8 +3483,20 @@ def _configure(self):
...
+        self.cores_per_node = self._cfg.get('cores_per_node')
+        if not self.cores_per_node:
+            self.cores_per_node = selected_cpus
+
+        requested_nodes = int(math.ceil(float(selected_cpus) / float(self.cores_per_node)))
+        self.node_list  = list()
+        for i in range(requested_nodes):
+            self.node_list.append("localhost")
+
+        self._log.debug('configure localhost to have %s nodes a %s cores',
```
marksantcroos: "a"?

marksantcroos:
Thanks, I made some modifications.

src/radical/pilot/agent/radical-pilot-agent-multicore.py
andre-merzky: That might be German -- 'à' means 'per' but inverses the direction:

  5 cores per node, for 2 nodes
  ==
  2 nodes a 5 cores
  ==
  2 nodes with 5 cores each (which you now used)

Germans took that over from the French, and I always assumed that it
works in English, too.  I checked now, and it doesn't :/  Its not used
in French anymore either, or so it seems :P

Thanks!

On Wed, Sep 16, 2015 at 10:07 AM, Mark Santcroos
<notifications@github.com> wrote:
> In src/radical/pilot/agent/radical-pilot-agent-multicore.py:
>
>> @@ -3483,8 +3483,20 @@ def _configure(self):
>>                  self._log.warn("more cores available: using requested %d
>> instead of available %d.",
>>                          selected_cpus, detected_cpus)
>>
>> -        self.node_list = ["localhost"]
>> -        self.cores_per_node = selected_cpus
>> +        # if cores_per_node is set in the agent config, we slice the
>> number of
>> +        # cores into that many virtual nodes.  cpn defaults to
>> selected_cpus,
>> +        # to preserve the previous behavior.
>> +        self.cores_per_node = self._cfg.get('cores_per_node')
>> +        if not self.cores_per_node:
>> +            self.cores_per_node = selected_cpus
>> +
>> +        requested_nodes = int(math.ceil(float(selected_cpus) /
>> float(self.cores_per_node)))
>> +        self.node_list  = list()
>> +        for i in range(requested_nodes):
>> +            self.node_list.append("localhost")
>> +
>> +        self._log.debug('configure localhost to have %s nodes a %s
>> cores',
>
> "a"?
>
> —
> Reply to this email directly or view it on GitHub.



-- 
99 little bugs in the code.
99 little bugs in the code.
Take one down, patch it around.

127 little bugs in the code...

PR #728, opened 2015-09-16
Feature/archer

marksantcroos:
Functionally working ORTE on ARCHER. Practically unusable because of file system suckiness on ARCHER.


src/radical/pilot/agent/radical-pilot-agent-multicore.py
```diff
@@ -2071,7 +2071,14 @@ def construct_command(self, task_exec, task_args, task_numcores,
...
 
         # Construct the hosts_string, env vars
-        hosts_string = ",".join([slot.split(':')[0] for slot in task_slots])
+        # On some Crays, like on ARCHER, the hostname is "archer_N".
+        # In that case we strip off the part upto and including the underscore.
+        #
+        # TODO: If this ever becomes a problem, i.e. we encounter "real" hostnames
+        #       with underscores in it, or other hostname mangling, we need to turn
+        #       this into a system specific regexp or so.
+        #
```
andre-merzky: We could eventually also make it a callout to 'host' ir 'nslookup', as the short names should always be resolvable on system level.  But right now I think this is the appropriate (if ugly) approach.

src/radical/pilot/configs/resource_epsrc.json
```diff
@@ -26,6 +25,42 @@
...
+        {
+            "job_manager_endpoint"    : "pbspro+ssh://login.archer.ac.uk/",
+            "filesystem_endpoint"     : "sftp://login.archer.ac.uk/"
+        },
+        "default_queue"               : "standard",
+        "lrms"                        : "PBSPRO",
+        "agent_type"                  : "multicore",
+        "agent_config"                : "cray",
+        "agent_scheduler"             : "CONTINUOUS",
+        "agent_spawner"               : "POPEN",
```
andre-merzky: Does the shell spawner also fail on Archer?  Not asking because I think we should switch, just wonder if you happen to have checked...

marksantcroos: Didn't try, but given the state on ARCHER thats likely not very conclusive anyway.

src/radical/pilot/agent/radical-pilot-agent-multicore.py
marksantcroos: What does the "short names" refer to? And what do you suggest?

andre-merzky: Maybe I still don't really understand what this is supposed to do, but I though this would translate hostname aliases as given by the nodelist into hostnames which are usable to connect to the hosts on IP/TCP level, ie. valid hostnames.  One would think that the things given in the hostlist would be translatable via the 'host' command?

marksantcroos: Nope, the hosts reported are just funny names it seems.

andre-merzky: pfft - ok then! :)

marksantcroos:
And does it work for you too?

andre-merzky:
yep, so it seems...

PR #735, opened 2015-09-18
Feature/fetchprofiles

marksantcroos:
Besides general improvements this is essential to get profiles off of machines where you run experiments from the headnode, and therefore can't use the URIs from the session as is.


andre-merzky:
Mark, could you please test this commit with `go://`?  It makes local URLs explicitly local.  FWIW, for sftp this needs saga-python/hotfix/issue_452 to function correctly.

andre-merzky:
See radical-cybertools/saga-python/pull/455

andre-merzky:
Thanks!

PR #736, opened 2015-09-18
Added scheme check for non-queuing adaptor mechanisms

iparask:


marksantcroos:
Can you add the other comments too that I wrote in my snippet?

iparask:
Done

marksantcroos:
Thanks!

PR #738, opened 2015-09-20
Feature/agent termination 2

andre-merzky:
I kind of forgot how hard it is to get threads and procs teared down in correct
order if shutdown can be triggered *anywhere*.  Python's strange object
destruction policies don't make it any simpler...  Anyway, I hope this is it
now:

  - separate out initialization from `__init__`
  - clearer separation of parent/child initialization / finalization
  - termination now always causes finalization, so that sigterm can trickle down
  - termination does not depend on messaging anymore
  - startup errors are considerd fatal
  - runtime errors are considered fatal, but can be made non-fatal by a switch
    in the Component base class
  - add a timeout to the startup barrier, to detect stuck sub-agents (timeout is not yet configurable!)
  - add locking to callback invocation, which specifically avoids races and deadlocks during shutdown

### DIFF
--- a/src/radical/pilot/agent/radical-pilot-agent-multicore.py
+++ b/src/radical/pilot/agent/radical-pilot-agent-multicore.py
@@ -5329,6 +5329,9 @@ class AgentWorker(rpu.Worker):
                 self._log.info('terminate: sub-agent %s got finalized' % arg)
                 self.close()
 
+            # other finalization messages are ignored -- we'll get our signal
+            # when its time...
+
 
     # --------------------------------------------------------------------------
     #
@@ -5876,6 +5879,8 @@ def bootstrap_3():
 
         _, mongo_db, _, _, _  = ru.mongodb_connect(cfg['mongodb_url'])
         mongo_p = mongo_db["%s.p" % cfg['session_id']]
+        pilot_id = cfg['pilot_id']
+
 
         # set up signal and exit handlers
         def exit_handler():
@@ -6017,7 +6022,6 @@ if __name__ == "__main__":
     print "---------------------------------------------------------------------"
     print
 
-    dh = ru.DebugHelper()
     bootstrap_3()
 
 #
--- a/src/radical/pilot/agent/radical-pilot-agent-multicore.py
+++ b/src/radical/pilot/agent/radical-pilot-agent-multicore.py
@@ -5558,7 +5558,7 @@ class AgentWorker(rpu.Worker):
 
                 with open (ls_name, 'w') as ls:
                     ls.write('#!/bin/sh\n\n')
-                    ls.write("%s\n" % cmd)
+                    ls.write("exec %s\n" % cmd)
                     st = os.stat(ls_name)
                     os.chmod(ls_name, st.st_mode | stat.S_IEXEC)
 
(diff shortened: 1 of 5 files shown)


andre-merzky:
Apologies that I already pulled proctitle PR in here -- I kind of needed that for debugging shutdown, and since feature/proctitle looks mergeable to Mark, I went the simple route...

src/radical/pilot/agent/radical-pilot-agent-multicore.py
```diff
@@ -5429,60 +5445,114 @@ def initialize(self):
...
-    def finalize(self):
+    def alive_barrier(self):
+
+        # FIXME: wait for bridges, too?  But we need pubsub for counting... Duh!
+        total = len(self._bridges)    + \
+                len(self._components) + \
+                len(self._workers   ) + \
+                len(self._sub_agents)
+        start   = time.time()
+        timeout = 60
```
andre-merzky: Hmm, probably not a good value on BW?  Any suggestion?

The timeout should actually not strictly be necessary anymore, as we watch the process handles now  and will detect if any component fails before it can send an alive message.  That does not hold for the bridges though (we need the channels to send alive in the first place... :P).  So, we could reduce the timeout scope to the bridges -- which makes this code somewhat more convoluted though...

src/radical/pilot/agent/radical-pilot-agent-multicore.py
```diff
@@ -5558,7 +5628,7 @@ def start_sub_agents(self):
 
                 with open (ls_name, 'w') as ls:
                     ls.write('#!/bin/sh\n\n')
-                    ls.write("%s\n" % cmd)
+                    ls.write("exec %s\n" % cmd)
```
andre-merzky: This is  also just tightening the process tree, and gives more direct cancellation.

src/radical/pilot/agent/radical-pilot-agent-multicore.py
```diff
@@ -5570,10 +5640,15 @@ def start_sub_agents(self):
             self._log.info ("create sub-agent %s: %s" % (sa, cmdline))
             sa_out = open("%s.out" % sa, "w")
             sa_err = open("%s.err" % sa, "w")
-            sa_proc = subprocess.Popen(args=cmdline, shell=True, stdout=sa_out, stderr=sa_err)
+            sa_proc = subprocess.Popen(args=cmdline, stdout=sa_out, stderr=sa_err)
```
andre-merzky: We are running a shell script, no need to use the shell twice.

marksantcroos:
Not again ...

src/radical/pilot/agent/radical-pilot-agent-multicore.py
marksantcroos: So what is the timeout for then?

andre-merzky: Well, it still does make  sure we are not caught pants down with no bridges: if the components come up, but the bridges not (or can't be used because of misconfiguration), then we won't get an alive message.  Which is good, because we would not get any other messages  either, such as units.  The timeout will prevent that scenario from resulting in a deadlock.  I don't think it serves any other purpose right now, but that is good enough until we have some more active bridge health monitoring (if ever).

marksantcroos:
Thanks!

PR #739, opened 2015-09-21
Feature/dependencies

marksantcroos:
Note that I also wouldnt mind adding setproctitle to the mix in the bootstrapper.


andre-merzky:
if you feel comfortable enough with the procnames, I would also not mind adding that.  Great, I'll do that then!

andre-merzky:
One thing I just realized: `VIRTENV_RADICAL_DEPS` is only interpreted when creating a new virtenv, not on updating it.  So this would now break any run on an existing ve.

How are we going to handle this?  I see two options: (i) keep the modules in `VIRTENV_RADICAL_DEPS` __and__ in setup.py, which is safe but redundant, or (ii) also interprete `VIRTENV_RADICAL_DEPS` in `update_virtenv`, which creates some seconds bootstrap overhead for each pilot.  I would actually opt for (i).

marksantcroos:
> One thing I just realized: VIRTENV_RADICAL_DEPS is only interpreted when creating a new virtenv, not on updating it.

How about https://github.com/radical-cybertools/radical.pilot/blob/devel/src/radical/pilot/bootstrapper/default_bootstrapper.sh#L812 ?

andre-merzky:
Hmm, I am confused. You are right, but then I don't understand what I just saw...   Let me look close to understand this... - thanks!

andre-merzky:
Ah - I had the mode on 'create' (not 'update as assume).  Create will would simply skip all pip installs if the ve existed already.  Well, that  makes the point weaker, but not invalid: we would break all    VEs which exist and where mode is 'create' -- which are rather many.

marksantcroos:
> we would break all VEs which exist and where mode is 'create' -- which are rather many.

Correct, but in my book that is more a reason to reconsider "create" as the de-facto standard.
Putting everything in setup.py and having it installed time and again in rp-install is not the solution either IMHO.

andre-merzky:
Well, it only installs what is not in the ve.  So having as much as possible in the ve on create is the right thing to do, which reduces the rp_install part to the radical stack, as it should.

I would not mind 'update' as default mode necessarily, but two points: it does add some bootstrap time, and it does add some potential for instability.  There is simply no real need to update 3rd part modules once a ve  is created and working...

We could slightly change the `create` semantic to 'check if ve exists *and has all modules needed*', which would basically include a `pip install` for all modules in `VIRTENV_RADICAL_DEPS` (but not `pip install --upgrade`)

marksantcroos:
Alternatively we could change the semantics of `update` to "update". I.e. add missing packages, but don't upgrade existing.

andre-merzky:
Are you sleep-walking? ;)

> Alternatively we could change the semantics of update to "update". I.e. add missing packages, but don't upgrade existing.

Well, maybe - but then we'd still have 'create' leaving things potentially unusable, and we loose the option to have a proper upgrade.  I don't think that buys us much compared to the fix on 'create'...

[I don't mind renaming update to upgrade, FWIW, but I actually think this means the same already.  But, whats in a name? :) ]

marksantcroos:
Is setproctitle required on the client side?
Depending on the answer I would either change the conditional installing in setup.py to unconditional or remote the block altogether.

marksantcroos:
> Well, maybe - but then we'd still have 'create' leaving things potentially unusable, and we loose the option to have a proper upgrade. I don't think that buys us much compared to the fix on 'create'...

Hmm, I was thinking about an additional target, but that doesn't really solve the problem.
Not that the problem has one solution anyway.

I think I'm leaning towards a default update/upgrade target that installs whats missing and upgrades whatever is upgradable (if we didn't specify a fixed version). I'm tempted to say that will be the most robust default. If people want more performance, they have the knobs to skip some of the aspects.

andre-merzky:
Again, I would agree with the default of upgrade -- but that still leaves create (as it is right now) in a potentially broken state...  

Asking  differently: where do you see the drawbacks of installing missing pieces during 'create'?

andre-merzky:
setproctitle: the module refactor is scheduled for the November timeframe.  At that point, most of the agent dependencies will also become module dependencies (zmq, ntplib, setproctitle),  with the exception of python-hostlist and netifaces (for the time being).  For that reason I don't think it makes much sense to optimize the separation between setup.py and bootstrapper.  The optimization should be to pre-install all dependencies in the VE during 'create', to update all dependencies during 'update', and to have only the radical modules under rp_install.  That worked good so far, modulo updates to the module list, and I think the diff between agent and module is too small to justify more finegrained control.

marksantcroos:
> Again, I would agree with the default of upgrade -- but that still leaves create (as it is right now) in a potentially broken state...

My suggestion would more or less replace create, at least as the default.

> Asking differently: where do you see the drawbacks of installing missing pieces during 'create'?

Version mismatches.

marksantcroos:
setproctitle: I asked, because doing a conditional install isnt the right thing to do for something that can change at runtime.

andre-merzky:
Maybe we should again step back a little?  Lets drop the names for the moment.

I think what we need to cover semantically are the following cases:

  * bare machine: a resource has not been used before by that user, no state ve exists.  We need to end up with a workable VE
  * used machine: a resource is in use, and has all packages installed in the VE already.
  * outdated machine: a resource had been used, but the installed ve has modules which need updating in order to obtain a usable VE
  
Additionally, we have two concerns which are somewhat orthogonal I think: static VEs and development mode, the latter meaning that we want want to  force an RP stack install with staged tarballs.

Is that set complete, in your opinion?

A.

marksantcroos:
> Maybe we should again step back a little? Lets drop the names for the moment.
> 
> I think what we need to cover semantically are the following cases:
> 
> bare machine: a resource has not been used before by that user, no state ve exists. We need to end up with a workable VE
> used machine: a resource is in use, and has all packages installed in the VE already.
> outdated machine: a resource had been used, but the installed ve has modules which need updating in order to obtain a usable VE
> Additionally, we have two concerns which are somewhat orthogonal I think: static VEs and development mode, the latter meaning that we want want to force an RP stack install with staged tarballs.
> 
> Is that set complete, in your opinion?
> 
> A.

Shall we move that to a ticket and consider the actual fix ready for merge now?

andre-merzky:
No, it does not fix the create - this breaks for me on many machines.  I'd rather leave setup.py as it is, which installs stuff in the rp_install/ trees, which is inefficient, but works.  I actually would always consider that useful, as a fallback.  Happy to merge the addition to DEPS though, to get new VEs created with the full set of deps outside of rp_install.  How about that?

marksantcroos:
> this breaks for me on many machines.

Can you explain me how this can fail as an explicit dependency and succeed as an implicit dependency?

marksantcroos:
Hmmm, I guess you refer to the create with an existing one. That sounds like circular reasoning, but I'll not resist to that now.
So I can live with the intermediate solution of only adding it to the DEPS.

andre-merzky:
circular in the context that 'update' should be the default?  Probably yes.  But it isn't, and even if we change this immediately, things should still work  with the non-default 'create'.  Either way, that compromise looks good to me, until we approach this more thoroughly... Thanks!

marksantcroos:
The diff has become really small now :)

andre-merzky:
Much smaller than the discussion would suggest :D

Thanks!

marksantcroos:
> Much smaller than the discussion would suggest :D

It speaks to the usefulness of discussing multiple commits in the context of one feature/PR.

andre-merzky:
Yep!

andre-merzky:
This specific PR is merged, closed in favor of #751

PR #744, opened 2015-09-23
Fix/shutdown race

andre-merzky:
There are two ways to shut down a component: calling stop() on it, which kills
the process, and sending a shutdown command over pubsub.  When implementing the
process based shutdown (stop()), I left the command based one in place as
fallback -- but in some cases (when a callback action takes some time for
example), the main loop's detection of the shutdown command is delayed, and the
process based one kicks in.  Which is fine, except that the AgentWorker expects
the component to listen to the shutdown at that point, and thus gets confused by
the disappearing child.

Semantically that does not matter much: the component process is gone either
way.  But I now dissabled the command based shutdown for all but the AgentWorker
(which gets that command from the main agent (HeartbeatMonitor).  The makes the
 shutdown sequence more predictable and avoid some inconsequential, but annoying
 stack traces during component finalization.

This PR also raises an error on agent config parsing problems (instead of failing later 
with an undefined symbol).


marksantcroos:
> This PR also raises an error on agent config parsing problems (instead of failing later 
> with an undefined symbol).

I dont see that part.

andre-merzky:
Me neither.  But this PR is not yet up for merging anyway I believe...

marksantcroos:
This has become a catch-all, but bottom line its better than devel.

andre-merzky:
thanks!

PR #749, opened 2015-09-23
Disable ntp for now.

marksantcroos:
The timeout on ntplib.NTPClient().request() is only for the socket call, not for the name lookup.
In practice this comes down to 40 seconds per Profiler() instantiation when there is no outbound network connectivity.

You do the math ...

On the Crays the time of the systems seem to be in sync.

### DIFF
--- a/src/radical/pilot/bootstrapper/default_bootstrapper.sh
+++ b/src/radical/pilot/bootstrapper/default_bootstrapper.sh
@@ -1305,6 +1305,12 @@ $converted_entry"
 done
 IFS=$OLD_IFS
 
+
+# we can't always lookup the ntp pool on compute nodes -- so do it once here,
+# and communicate the IP to the agent.  The agent may still not be able to
+# connect, but then a sensible timeout will kick in on ntplib.
+RADICAL_PILOT_NTPHOST=`dig +short 0.pool.ntp.org | grep -v "\.$" | head -n 1`
+
 # Before we start the (sub-)agent proper, we'll create a bootstrap_2.sh script
 # to do so.  For a single agent this is not needed -- but in the case where
 # we spawn out additional agent instances later, that script can be reused to
@@ -1338,6 +1344,9 @@ export SAGA_VERBOSE=DEBUG
 export RADICAL_VERBOSE=DEBUG
 export RADICAL_UTIL_VERBOSE=DEBUG
 export RADICAL_PILOT_VERBOSE=DEBUG
+
+# avoid ntphost lookups on compute nodes
+export RADICAL_PILOT_NTPHOST=$RADICAL_PILOT_NTPHOST
 
 # pass environment variables down so that module load becomes effective at
 # the other side too (e.g. sub-agents).
--- a/src/radical/pilot/utils/prof_utils.py
+++ b/src/radical/pilot/utils/prof_utils.py
@@ -114,9 +114,9 @@ class Profiler (object):
         # We first try to contact a network time service for a timestamp, if that
         # fails we use the current system time.
         try:
-            assert True == False
             import ntplib
-            response = ntplib.NTPClient().request('0.pool.ntp.org')
+            ntphost  = os.environ.get('RADICAL_PILOT_NTPHOST', '0.pool.ntp.org')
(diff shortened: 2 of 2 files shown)


marksantcroos:
I think the state of that NTP library is very dubious. For one thing, it doesnt even work on my laptop.

andre-merzky:
What happens on your laptop?  I actually tested MacOS for once, it worked back then, and still does:

```
(ve)cameo:~ merzky$ uname -a
Darwin cameo.fritz.box 14.4.0 Darwin Kernel Version 14.4.0: Thu May 28 11:35:04 PDT 2015; root:xnu-2782.30.5~1/RELEASE_X86_64 x86_64

(ve)cameo:~ merzky$ python
Python 2.7.6 (default, Sep  9 2014, 15:04:36) 
[GCC 4.2.1 Compatible Apple LLVM 6.0 (clang-600.0.39)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import ntplib
>>> from time import ctime
>>> c = ntplib.NTPClient()
>>> response = c.request('pool.ntp.org')
>>> print ctime(response.tx_time)
Wed Sep 23 19:47:14 2015
```

I am hesitant to just remove / disable it unconditionally.  It was added to solve a real problem, that of event ordering on combined profiles.  Otherwise I see negative durations on stampede, for example, which kind of screw with the whole analysis.

Let me look into ntplib timeouts.  I don't mind falling back to system time for nodes which can't use it, obviously...

marksantcroos:
This is on my laptop.

```
$ python
Python 2.7.10 (default, Sep 22 2015, 10:25:02) 
[GCC 4.2.1 Compatible Apple LLVM 7.0.0 (clang-700.0.72)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import ntplib
>>> from time import ctime
>>> c = ntplib.NTPClient()
>>> response = c.request('pool.ntp.org')
>>> Traceback (most recent call last):
  File "/Users/mark/.virtualenv/rp/lib/python2.7/site-packages/ntplib.py", line 316, in request
    raise NTPException("No response received from %s." % host)
ntplib.NTPException: No response received from pool.ntp.org.

>>> 
```

Using ntpdate works fine, so its not a networking issue.
```
$ ntpdate -q pool.ntp.org
server 83.98.201.134, stratum 2, offset 0.012140, delay 0.03311
server 188.166.42.87, stratum 3, offset -0.000524, delay 0.05829
server 91.148.192.49, stratum 2, offset 0.012354, delay 0.03331
server 195.191.112.251, stratum 16, offset -0.033335, delay 0.09229
23 Sep 20:49:10 ntpdate[9577]: adjust time server 83.98.201.134 offset 0.012140 sec
```

marksantcroos:
I agree that we need some synchronisation.

NTP is not a general viable option, and I would prefer to have one solution that works everywhere.

Wouldn't it be relatively easy to do some kind of broadcast over zmq and create an offset from that?

andre-merzky:
synching times is a hard problem -- NTP is made to do it.  I would only look for other solutions where NTP does not work due to connectivity.  The main problem we are trying to solve is syncing between RP module and RP agent, and between different agents.  Compute nodes are likely not far off from head node clocks (although I have seen that actually happen to cause event reordering).

Using ZMQ would mean we implement our own, ntp like protocol, because you need to remove the communication delays from the equation.  rtt/2 may or may not be good enough.  Also, that works only within the agent at the moment, so cannot be an universal solution (yet).

I do have a solution to avoid the loopkup timeouts without changing much - I am testing that right now on BW.  That might not be an ultimate solution, but should take the urgency out of the discussion, I hope...

andre-merzky:
Hey Mark, 

as discussed, here is the temporary workaround to avoid the accumulating dns lookup delays.  The ntp request delays are now limited to 1sec/call, which still adds up to several seconds per agent, depending on number of components.

The time source (`ntp` or `sys`) is now recorded in the initial profile timestamp, to make it simpler to revert ntp after data collection, if we so want.

I tested local and stampede, we probably should make sure that this does not trip anything  strange on BW.

marksantcroos:
Looks good. Didnt actually run it (yet), as there are too many individual branches that I need.

PR #754, opened 2015-09-24
Feature/orteconfigs

marksantcroos:
Modulo the weirdness we see specifically on BW its ready for prime time.


andre-merzky:
Wohoo! :)

Issue #747, opened 2015-09-23
Missing module encountered while executing getting started local example


vivek-bala:
Elena from Edinburgh ran into the following issue.

Pilot fails right after going into PendingActive. Looking at the agent.err file tells that module **hostlist**, which is required by the agent(https://github.com/radical-cybertools/radical.pilot/blob/master/src/radical/pilot/agent/radical-pilot-agent-multicore.py#L150), is missing. 

This is on python-2.7.5. I'm not sure if this is not available in 2.7.5  or something specific to that installation, but maybe this could be documented (installation section/ FAQ).

Complete thread: https://bitbucket.org/extasy-project/extasy/issues/6/readthedocs-for-extasy-01-running-01

marksantcroos:
I dont think this is a documentation issue.
It is supposed to get installed automagically, so the user doesn't need to do anything.

The fact that the installation apparently failed, but that the agent still started is weird though.

Would need to see full logs to judge better.

vivek-bala:
The verbose log is available at https://gist.github.com/anonymous/7841805eebb19ead9c71

andre-merzky:
Would you also have the logs from the agent side?  Specifically the one which shows the missing module message / stacktrace, and bootstrap_1.err/out.

vivek-bala:
This is using rp 0.35. Are bootstrap_1.err/out produced ?

Agent.err:
```
more agent.err Python 2.7.5 Traceback (most recent call last): File "/Users/elenabreitmoser/radical.pilot.sandbox/rp.session.mbp-eb.epcc.ed.ac.uk.elenabreitmoser.016701.0003-pilot.0000/rp_install/bin/radical-pilot-agent-multicore.py", line 150, in <module> import hostlist ImportError: No module named hostlist
```

andre-merzky:
Ah, thanks for the version info! (please always include that in the ticket message -- but I should have seen that in the gist log you provided...

This looks like the pilot is running on localhost?  Please ask Elena to remove `$HOME/radical.pilot.sandbox/ve_local/` (and any other virtualenv in `$HOME/radical.pilot.sandbox/`),  and then try again.  I think this is a problem with an old, incomplete virtualenv and the default local resource config, which is not set to update that VE.

If that fails, can you please also send Agent.out then, which contains the installation log?

Thanks!

vivek-bala:
Elena Breitmoser reporter

I do the following

1) Fresh virtualenv
```
[...]

PR #764, opened 2015-09-29
use POPEN for localhost

andre-merzky:
There is a lingering issue in the shell spawner which makes SHELL
unstable on localhost.  The fix is in devel, but instead of starting
to cherry-pick, it seems easier to switch to POPEN for the time being.

This relates to #747.


marksantcroos:
Agreed on the hotfix approach.

marksantcroos:
Will you roll a release?

andre-merzky:
Yes.  We don't really have a hotfix-release procedure.  I'll just push that to pypi, ok?

PR #771, opened 2015-09-29
move time source hint out of the parsers way

andre-merzky:


marksantcroos:
Closing, and will re-open towards devel.

PR #778, opened 2015-09-30
Hotfix/zmq autoport

andre-merzky:
Queue and Pubsub bridges now find free ports on their own

This implies that we need to inspect the bridges for the used addresses afterwards, and that we need to communicate both ends (sink and source) to the agent components individually.

Further, as the addresses are now only known *after* starting the bridges, we need to move the bridge startup (and shutdown) into bootstrap_3, before creating the AgentWorker, as that already tries to connect to the bridges.

### DIFF
--- a/src/radical/pilot/agent/radical-pilot-agent-multicore.py
+++ b/src/radical/pilot/agent/radical-pilot-agent-multicore.py
@@ -5349,7 +5349,6 @@ class AgentWorker(rpu.Worker):
 
         # keep track of objects we need to stop in the finally clause
         self._sub_agents = dict()
-        self._bridges    = dict()
         self._components = dict()
         self._workers    = dict()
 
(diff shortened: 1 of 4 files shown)


src/radical/pilot/configs/resource_local.json
```diff
@@ -24,7 +24,8 @@
         "mpi_launch_method"           : "MPIEXEC",
         "rp_version"                  : "debug",
         "virtenv"                     : "%(global_sandbox)s/ve_localhost",
-        "virtenv_mode"                : "create"
+        "virtenv_mode"                : "create",
+        "cores_per_node"              : 1
```
marksantcroos: Can you undo this bit?

andre-merzky: Ah, right,  sorry for that...  I'll undo (even if I find it useful, as that is the only way to have non-trivial agent configs on localhost.  But I should put that into my private configs, apologies!)

marksantcroos: Agreed. I have similar local changes.

andre-merzky: Done.

marksantcroos:
Thanks!

PR #779, opened 2015-09-30
Record the pilot id in profiles.

marksantcroos:
Andre, did you remove these intentionally?


andre-merzky:
I don't think so?  At some point, the Profiler c'tior accepted a uid -- maybe that is a  remnant of that time, or those entries never had a uid to begin with?  Not sure...

marksantcroos:
> At some point, the Profiler c'tior accepted a uid

Could be indeed, I actually considered doing that now, but didnt bother.

Thanks for the quick merge.

PR #783, opened 2015-10-02
Fix/archer python

marksantcroos:
I've tested it myself and it works.


andre-merzky:
Works like a charm - thanks!

PR #784, opened 2015-10-02
Fix/bootstrap locking

andre-merzky:
This should fix the VE locking problems on Archer and SuperMUC.

### DIFF
--- a/src/radical/pilot/bootstrapper/default_bootstrapper.sh
+++ b/src/radical/pilot/bootstrapper/default_bootstrapper.sh
@@ -415,17 +415,14 @@ virtenv_setup()
     virtenv="$2"
     virtenv_mode="$3"
 
-    virtenv_create=TRUE
-    virtenv_update=TRUE
-
-    lock "$pid" "$virtenv" # use default timeout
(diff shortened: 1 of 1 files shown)


src/radical/pilot/bootstrapper/default_bootstrapper.sh
```diff
@@ -433,8 +430,9 @@ virtenv_setup()
 
     elif test "$virtenv_mode" = "update"
     then
-        test -f "$virtenv/bin/activate" || virtenv_create=TRUE
+        virtenv_create=FALSE
         virtenv_update=TRUE
+        test -f "$virtenv/bin/activate" || virtenv_create=TRUE
```
marksantcroos: This will lead to a create and then an update. (Which I think also happened before the change)

marksantcroos:
Other than that comment its good to go.

src/radical/pilot/bootstrapper/default_bootstrapper.sh
andre-merzky: That is true, on both accounts.  Let me remove that pessimisation...

PR #787, opened 2015-10-02
muhahaha

andre-merzky:
;)

### DIFF
--- a/src/radical/pilot/agent/radical-pilot-agent-multicore.py
+++ b/src/radical/pilot/agent/radical-pilot-agent-multicore.py
@@ -2273,7 +2273,7 @@ class LRMS(object):
 
         if self.agent_nodes:
             self._log.info('Reserved agent node(s): %s' % self.agent_nodes.values())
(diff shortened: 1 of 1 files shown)


marksantcroos:
I will only accept this once it aligns perfectly on pixel level for every number of nodes for each of the sets.

andre-merzky:
working on it...

andre-merzky:
hahaha - and now you can't tell if I am kidding, right?? :D

src/radical/pilot/agent/radical-pilot-agent-multicore.py
```diff
@@ -2273,7 +2273,7 @@ def __init__(self, cfg, logger):
 
         if self.agent_nodes:
             self._log.info('Reserved agent node(s): %s' % self.agent_nodes.values())
-            self._log.info('Agent(s) running on node(s): %s' % self.agent_nodes.keys())
+            self._log.info('Utilized agent node(s): %s' % self.agent_nodes.keys())
```
marksantcroos: Too bad the new text is wrong.

PR #795, opened 2015-10-05
Feature/clone drop cb

andre-merzky:
When cloned units are pushed through the scheduler, but are dropped right afterwards, the scheduled cores remain allocated forever.  This PR adds a callback mechanism to deallocate those cores when  dropping the clones.


andre-merzky:
The code in this PR has been extensively used by the RP micro benchmarks, and I thus considere it ready for merge.

PR #796, opened 2015-10-05
Fix/init sequence

andre-merzky:
change the sequence of init defs so that early exceptions don't get undefs

Also makes sure that clean termination happens on SIGTERM


marksantcroos:
I only see the addition of the SIGTERM, I think you piggy-backed the bulk of this PR already via some other route.

andre-merzky:
Yes, the definition move got cherry-picked somewhere already.  Thanks!

PR #798, opened 2015-10-05
Feature/proftarball

marksantcroos:
With a large number of sub-agents and instances, the number of profile transfers grew out of hand.

### DIFF
--- a/src/radical/pilot/bootstrapper/default_bootstrapper.sh
+++ b/src/radical/pilot/bootstrapper/default_bootstrapper.sh
@@ -1414,6 +1414,22 @@ profile_event 'cleanup done'
 echo "#"
 echo "# -------------------------------------------------------------------"
 
+echo
+echo "# -------------------------------------------------------------------"
+echo "#"
+echo "# Tarring profiles ..."
+PROFILES_TARBALL="profiles.tgz"
+tar -czf $PROFILES_TARBALL *.prof
+ls -l $PROFILES_TARBALL
+echo "#"
+echo "# -------------------------------------------------------------------"
+
+echo
+echo "# -------------------------------------------------------------------"
+echo "#"
+echo "# Done, exiting!"
+echo "#"
+echo "# -------------------------------------------------------------------"
+
 # ... and exit
 exit $AGENT_EXITCODE
-
--- a/src/radical/pilot/utils/session.py
+++ b/src/radical/pilot/utils/session.py
@@ -2,6 +2,7 @@ import os
 import sys
 import glob
 import saga
+import tarfile
 
 import radical.utils as ru
 from   radical.pilot.states import *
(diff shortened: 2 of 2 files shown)


src/radical/pilot/bootstrapper/default_bootstrapper.sh
```diff
@@ -1414,6 +1414,22 @@ profile_event 'cleanup done'
 echo "#"
 echo "# -------------------------------------------------------------------"
 
+echo
+echo "# -------------------------------------------------------------------"
+echo "#"
+echo "# Tarring profiles ..."
+PROFILES_TARBALL="profiles.tgz"
+tar -czf $PROFILES_TARBALL *.prof
```
andre-merzky: Uh, `-czf`?  You live a dangerous life... ;)  Lets see if you know what I mean...  The line is correct though...

andre-merzky:
Nice PR, thanks.  Let me test before merge later today...

src/radical/pilot/bootstrapper/default_bootstrapper.sh
marksantcroos: Let me try to poke your dirty little brain: POSIX, i.e. "tar czf"?

andre-merzky: Yes, exactly :) `tar zcf file.tgz ` an `tar zfc file.tgz` are equivalent.  OTOH, `tar -zcf file.tgz` and `tar -zfc file.tgz` are not.  But as said the line is correct, and the un-dashed version has one of the wierdest arg conventions, ever... :P

src/radical/pilot/utils/session.py
```diff
@@ -93,6 +94,49 @@ def fetch_profiles (sid, dburl=None, client=None, tgt=None, access=None, session
...
+            tarball.extractall(tgt_url.path)
+
+            profiles = glob.glob("%s/*.prof" % tgt_url.path)
+            print "Tarball %s extracted to '%s'." % (ftgt.path, profiles)
+            ret.extend(profiles)
+
+            # If extract succeeded, no need to fetch individual profiles
+            continue
+
+        # If we dont have a tarball (for whichever reason), fetch individual profiles
```
marksantcroos: Given that the lack of a .tgz is likely a non-proper shutdown, should we actually get rid of the fallback?

andre-merzky: In principle, yes.  In practice, I can think of more reasons why a tarball would not get created.  The agent being force-killed due to walltime limits being the obvious one.

marksantcroos: 1. That would not make the profiles more useful?
2. That would still account as a non-proper shutdown as the agent follow instructions from the lrms to shutdown.

andre-merzky: 1. The profiles may well contain all information you may want to have, as we try to flush the profiles on exit
2. I know someone who maintains the agent is in `DONE` state when being killed due to walltime limits.  That sounds contradictory to me wrt. to  your statement (unless I misparse it?)

marksantcroos: 1. Ok, thats fine. Is there a clear way to find out that all is there?
2. I guess so. Walltime end should lead to a clear shutdown. Has nothing to do with the actual final state.

andre-merzky: 1. https://github.com/radical-cybertools/radical.pilot/blob/devel/src/radical/pilot/agent/radical-pilot-agent-multicore.py#L6100 and https://github.com/radical-cybertools/radical.pilot/blob/devel/src/radical/pilot/utils/component.py#L338 is probably what you would look for.  We can add a `flushed` message in `prof.flush()` to make this simpler.  But even if not all data are flushed, it depends on what you want to look for in the profiles, they still may be useful, and still may contain all   unit information, for example, only missing shutdown stuff.  I am all for checking and verifying - but I am hesitant to throw away partial data just because they are not useful for a specific case...

2. I guess that depends how the batch system enforces shutdown.  I don't know if there is a generic rule, but there is not much   to be don on a SIGKILL, for example, specifically no clean shutdown with some minutes of FS interactions (archer).

marksantcroos: 1. Agreed.

2. Pretty universal: first TERM, then KILL. Most systems also allow you do specify the delay between the two.

andre-merzky: What is the default delay between the two?  Would it be (reasonably) safe to assume that tar (and other shutdown activities such as profile and log flushing, mongodb state updates) are not interrupted?

andre-merzky: 2. And, should we request a reasonable large delay?

marksantcroos: > What is the default delay between the two?

Manpage says its a system setting. I recall 10 seconds.

> Would it be (reasonably) safe to assume that tar (and other shutdown activities such as profile and log flushing, mongodb state updates) are not interrupted?

Assuming that we implement handling of SIGTERM in the bootstrapper you mean?

marksantcroos: > And, should we request a reasonable large delay?

Once we have proper signal handling in place, we could indeed play with this value if defaults are too short.

PR #802, opened 2015-10-06
only profile pre and post exec if profiling is enabled

andre-merzky:


src/radical/pilot/agent/radical-pilot-agent-multicore.py
```diff
@@ -3769,9 +3769,11 @@ def spawn(self, launcher, cu):
         with open(launch_script_name, "w") as launch_script:
             launch_script.write('#!/bin/bash -l\n\n')
 
-            launch_script.write("echo script start_script `%s` >> %s/PROF\n" % (cu['gtod'], cu_tmpdir))
+            if 'RADICAL_PILOT_PROFILE' in os.environ:
```
marksantcroos: Don't we have a flag we can check? Do we know this environment variable actually reaches so far?

andre-merzky: Good question.  No, we don't have a flag.  We may want to consider adding the value to the agent config, which is the canonical place for these things anyway.

Reach:  it seemed to  work on BW, avoiding the `PROF` parsing errors when profiling was disabled.  I am not perfectly sure if env is forwarded on all agent launch methods.

marksantcroos: > We may want to consider adding the value to the agent config, which is the canonical place for these things anyway.

+1

marksantcroos: > Reach: it seemed to work on BW, avoiding the PROF parsing errors when profiling was disabled. I am not perfectly sure if env is forwarded on all agent launch methods.

Thats at least something, as its "rather far".

marksantcroos: Are you working on this? Otherwise I'll do it, I kind of need it "right now".

andre-merzky: No, not really working on it.  But I also don't think we do need a  cfg flag right now? (I think this is what you refer to?)

marksantcroos: You mean that we can work around it?

andre-merzky: No, I meant that using the env seems to work, at least for what I tested.  If you observed otherwise,  then the config flag would be the right thing to do.  Did you?

andre-merzky: FWIW: the profile itself also depends on the env var, and that seemed to   work reliably in sub-agents so far...

marksantcroos: Hop ...

andre-merzky: Mark, can  you be verbose, please?  As said, if  you observe the env var not being forwarded, than a config flag is the right thing to do.  That would also affect profiling I assume.  No, I am not working on it right now (didn't see the need).

marksantcroos: > FWIW: the profile itself also depends on the env var, and that seemed to work reliably in sub-agents so far...

What do you mean with this? We were talking about the profile(r) all the time already?

marksantcroos: > Mark, can you be verbose, please? As said, if you observe the env var not being forwarded, than a config flag is the right thing to do.

It's (was) not passed on an SSH hop.

marksantcroos: > That would also affect profiling I assume.

What do you mean with this? Performance impact?

marksantcroos: > FWIW: the profile itself also depends on the env var, and that seemed to work reliably in sub-agents so far...

I think why it worked depends on a number of circumstances. For local it inherits the shell. For ORTE, the orte-dvm actually got the environment variable, so it magically used that for all orte-submit's too.

marksantcroos:
It works, thanks!

src/radical/pilot/agent/radical-pilot-agent-multicore.py
andre-merzky: > I think why it worked depends on a number of circumstances. For local it inherits the shell. For ORTE, the orte-dvm actually got the environment variable, so it magically used that for all orte-submit's too.

That makes sense I guess.  Thanks for catching it!

PR #809, opened 2015-10-07
Feature/ssh envs

marksantcroos:
In full disclosure, I actually dont know why we dont put it in the script itself, but that is a more major operation. Also, there actually might be a reason why we dont do that :-)


andre-merzky:
Agree that it probably makes sense to move that to the script...

PR #810, opened 2015-10-08
Add 'br0' to list of preferred interface names.

### LINKED ISSUE #789: v0.36RC1 fails on SuperMic
I am getting this error in the bootstrap_3.log:

```
2015-10-04 23:07:32,498: agent_0.bootstrap_3 : MainProcess                     : MainThread     : ERROR   :   File "/worka/work/iparaske/radical.pilot.sandbox/ve_supermic/rp_install/bin/radical-pilot-agent-multicore.py", line 6050, in bootstrap_3
    nodeip = LRMS.hostip(cfg.get('network_interface'))
  File "/worka/work/iparaske/radical.pilot.sandbox/ve_supermic/rp_install/bin/radical-pilot-agent-multicore.py", line 2398, in hostip
    ip = netifaces.ifaddresses(pref)[af][0]['addr']
```

marksantcroos:
Fixes #789 and possibly some more.

### DIFF
--- a/src/radical/pilot/agent/radical-pilot-agent-multicore.py
+++ b/src/radical/pilot/agent/radical-pilot-agent-multicore.py
@@ -2364,7 +2364,11 @@ class LRMS(object):
         black_list = ['lo', 'sit0']
 
         # Known intefaces in preferred order
-        sorted_preferred = ['ipogif0', 'eth0']
+        sorted_preferred = [
+            'ipogif0', # Cray's
+            'br0', # SuperMIC
+            'eth0'
+        ]
         
         # Get a list of all network interfaces
         all = netifaces.interfaces()


andre-merzky:
If I understand the code correctly, it should have found that interface eventually, even if it was not preferred?  Or did another, invalid one take precedence?  In general, is that something which should move to the resource config (ie. to a per-host setting)?

marksantcroos:
> Or did another, invalid one take precedence?

This.

> In general, is that something which should move to the resource config (ie. to a per-host setting)?

We could. Lets see how it goes, adding a couple of more to the list is fine.

There is also some room for additional heuristics, but if that becomes required I would agree that configuration would be the next step.

andre-merzky:
Ack

PR #821, opened 2015-10-09
Feature/log.tgz

andre-merzky:
This PR creates a logfile tarball for easier ticket attachement.  No tool for downloading that tarball is yet included.  Some small  changes to profile fetching are included in the PR, too, to avoid file name collisions on multi-pilot runs.

### DIFF
--- a/src/radical/pilot/bootstrapper/default_bootstrapper.sh
+++ b/src/radical/pilot/bootstrapper/default_bootstrapper.sh
@@ -117,7 +117,7 @@ profile_event()
     fi
 
     printf "%.4f,%s,%s,%s,%s,%s\n" \
-        "$NOW" "bootstrap_1" "$PILOT_ID" "ACTIVE" "$event" "$msg" \
+        "$NOW" "bootstrap_1" "$PILOTID" "ACTIVE" "$event" "$msg" \
         >> "$PROFILE"
 }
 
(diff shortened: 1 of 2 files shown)


src/radical/pilot/bootstrapper/default_bootstrapper.sh
```diff
@@ -1414,15 +1414,31 @@ profile_event 'cleanup done'
...
-echo
-echo "# -------------------------------------------------------------------"
-echo "#"
-echo "# Tarring profiles ..."
-PROFILES_TARBALL="profiles.tgz"
-tar -czf $PROFILES_TARBALL *.prof
-ls -l $PROFILES_TARBALL
-echo "#"
-echo "# -------------------------------------------------------------------"
+if ! test -z "`ls *.prof 2>/dev/null`"
```
marksantcroos: Use the return value instead of string parsing?

src/radical/pilot/utils/session.py
```diff
@@ -123,14 +123,19 @@ def fetch_profiles (sid, dburl=None, client=None, tgt=None, access=None, session
...
+            os.mkdir("%s/%s" % (tgt_url.path, pilot['_id']))
+        except OSError:
+            pass
+
         # We now have a local tarball
         if tarball_available:
             print "Extracting tarball %s into '%s'." % (ftgt.path, tgt_url.path)
             tarball = tarfile.open(ftgt.path)
-            tarball.extractall(tgt_url.path)
+            tarball.extractall("%s/%s" % (tgt_url.path, pilot['_id']))
```
marksantcroos: Ah right, good catch to handle multiple pilots. This likely requires changes in the consumption parts too.

src/radical/pilot/bootstrapper/default_bootstrapper.sh
andre-merzky: *shrug* Don't care much.  Its not really parsing though, is it?

src/radical/pilot/utils/session.py
andre-merzky: Yeah, sorry for that, thought so... :/

marksantcroos: I'll look over those.

src/radical/pilot/bootstrapper/default_bootstrapper.sh
marksantcroos: If all equal I couldn't care much either, I agree. But I was wondering whether this could break with gazillion of files.

andre-merzky: No, I don't think so -- at least not in a way that using the return val would make a difference.  Like, when max command line length is hit we are screwed either way...

src/radical/pilot/bootstrapper/default_bootstrapper.sh
```diff
@@ -1414,15 +1414,31 @@ profile_event 'cleanup done'
...
+then
+    echo
+    echo "# -------------------------------------------------------------------"
+    echo "#"
+    echo "# Tarring profiles ..."
+    PROFILES_TARBALL="$PILOTID.prof.tgz"
```
marksantcroos: Why do you actually choose to add the pilotid to the filename remotely already? Isnt that clear from the pilot directory?

andre-merzky: because that makes out-of-band download simpler.  That is probably more important for the log tarball, which we expect people to provide for tickets, but then I like symmetry as you know, so did that for both tarballs, as the prof one needed renaming on the other end anyway...

marksantcroos: Hmm, ok, not sure I see that argument really, but not a big deal.

src/radical/pilot/utils/session.py
marksantcroos: Nothing in RP required further changes and downstream only one extra line so far.

PR #829, opened 2015-10-12
Wait for its child.

marksantcroos:
Have the parent component wait for the child during termination.


andre-merzky:
Yes, the fix looks   good - many thanks!

PR #833, opened 2015-10-14
make localhost (ForkLRMS) behave like a resource with an infinite number of cores

andre-merzky:
infinite is very difficult to spell...


andre-merzky:
wrong target...

PR #835, opened 2015-10-14
Feature/radical undebug

andre-merzky:
remove RADICAL_DEBUG, use proper logger in queue, pubsub

### DIFF
--- a/src/radical/pilot/utils/pubsub.py
+++ b/src/radical/pilot/utils/pubsub.py
@@ -7,6 +7,7 @@ import errno
 import pprint
 import Queue           as pyq
 import multiprocessing as mp
+import radical.utils   as ru
 
 # --------------------------------------------------------------------------
 # defines for pubsub roles
--- a/src/radical/pilot/utils/pubsub.py
+++ b/src/radical/pilot/utils/pubsub.py
@@ -73,18 +73,15 @@ class Pubsub(object):
         self._role       = role
         self._addr       = address
         self._debug      = False
-        self._logfd      = None
+        self._log        = ru.get_logger('rp.bridges')
         self._name       = "pubsub.%s.%s" % (self._channel, self._role)
         self._bridge_in  = None           # bridge input  addr
         self._bridge_out = None           # bridge output addr
 
-        if 'msg' in os.environ.get('RADICAL_DEBUG', '').lower():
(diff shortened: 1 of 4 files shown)


src/radical/pilot/utils/pubsub.py
```diff
@@ -73,18 +74,15 @@ def __init__(self, flavor, channel, role, address=None):
         self._role       = role
         self._addr       = address
         self._debug      = False
-        self._logfd      = None
+        self._log        = ru.get_logger('rp.bridges')
```
marksantcroos: Where is this supposed to end up? I dont see the messages.

andre-merzky: I wanted to not put this into the `radical.pilot.util` logger space, as that would be enabled by `RADICAL_PILOT_VERBOSE`, but logging on the queues can seriously impact performance.   So that would only get enabled on `RP_BRIDGES_VERBOSE=DEBUG` (which defaults, as all other levels, to `WARNING`).  It goes to stderr by default, but it probably would make sense to specify a `[self.name].log` as we do for the agent.

That RP namespace stuff might also point to the need to rethink how we handle logger hierarchies, maybe we want to do that different than suggested  by Python's logging module.  But either way, both (log target and log level spec) is for later -- the purpose of this PR is really mainly to get rid of `RADICAL_DEBUG`, w/o breaking things...

marksantcroos: > I wanted to not put this into the radical.pilot.util logger space, as that would be enabled by RADICAL_PILOT_VERBOSE, but logging on the queues can seriously impact performance. So that would only get enabled on RP_BRIDGES_VERBOSE=DEBUG (which defaults, as all other levels, to WARNING).

Ok.

> It goes to stderr by default, but it probably would make sense to specify a [self.name].log as we do for the agent.

Indeed.

> That RP namespace stuff might also point to the need to rethink how we handle logger hierarchies, maybe we want to do that different than suggested by Python's logging module. But either way, both (log target and log level spec) is for later -- the purpose of this PR is really mainly to get rid of RADICAL_DEBUG, w/o breaking things...

Ok.

marksantcroos:
Thanks. If you could take of routing the output from stderr to some file at some point that would be great.

PR #836, opened 2015-10-14
make localhost (ForkLRMS) behave like a resource with an inifnite number of cores

andre-merzky:
infinite is very difficult to spell...


marksantcroos:
Thanks

PR #842, opened 2015-10-15
Bring code in line with documentation :-)

marksantcroos:
Allows for an individual log level on the agent side of things.


andre-merzky:
:)

PR #844, opened 2015-10-17
Feature/yarn

iparask:


src/radical/pilot/agent/radical-pilot-agent-multicore.py
```diff
@@ -152,12 +152,17 @@ class UpdateWorker
...
+import json
+import urllib2 as ul
 
 import saga                as rs
 import radical.utils       as ru
 import radical.pilot       as rp
 import radical.pilot.utils as rpu
 
+from datetime import datetime
+
```
andre-merzky: This is not used anymore I believe?

src/radical/pilot/agent/radical-pilot-agent-multicore.py
```diff
@@ -1170,6 +1179,190 @@ def corner2offset(self, block, corner):
...
+    # --------------------------------------------------------------------------
+    #
+    def _configure(self):
+
+        try:
+            #-----------------------------------------------------------------------
+            # Find out how many applications you can submit to YARN. And also keep
+            # this check happened to update it accordingly
+
```
andre-merzky: What do you mean with 'keep this check happen[sic]'?  Where is it kept?

src/radical/pilot/agent/radical-pilot-agent-multicore.py
```diff
@@ -1170,6 +1179,190 @@ def corner2offset(self, block, corner):
...
+            self._mem_size = metrics['clusterMetrics']['allocatedMB']
+
+            self.avail_app = {'apps':max_num_app - num_app,'timestamp':sample_time}
+            self.avail_cores = self._mnum_of_cores - self._num_of_cores
+            self.avail_mem = self._mmem_size - self._mem_size
+
+
+            return True
+        except:
+            return False
```
andre-merzky: The return value of `_configure` is not evaluated - you may want to let the exception fall through...

src/radical/pilot/agent/radical-pilot-agent-multicore.py
```diff
@@ -1170,6 +1179,190 @@ def corner2offset(self, block, corner):
...
+        yarn_schedul_json = json.loads(yarn_status.read())
+
+        max_num_app = yarn_schedul_json['scheduler']['schedulerInfo']['queues']['queue'][0]['maxApplications']
+        num_app = yarn_schedul_json['scheduler']['schedulerInfo']['queues']['queue'][0]['numApplications']
+        if (self.avail_app['timestamp'] - sample)>60 and \
+           (self.avail_app['apps'] != max_num_app - num_app):
+            self.avail_app['apps'] = max_num_app - num_app 
+            self.avail_app['timestamp']=sample
+
+        return 'Can accept up to {0} applications per user. Free cores {1} Free Mem {2}'.format(self.avail_app['apps'],self.avail_cores,self.avail_mem)
```
andre-merzky: line is too long :)

src/radical/pilot/agent/radical-pilot-agent-multicore.py
```diff
@@ -1170,6 +1179,190 @@ def corner2offset(self, block, corner):
...
+        with self._slot_lock :        
+
+            self._log.info(self.slot_status())
+            self._log.debug('YARN Service and RM URLs: {0} - {1}'.format(self._service_url,self._rm_url))
+                    
+            cu['opaque_slots']={'lm_info':{'service_url':self._service_url,
+                                            'rm_url':self._rm_url,
+                                            'nodename':self._client_node},
+                                'task_slots':[cu['description']['cores'],2048]
+                                            }
```
andre-merzky: 2048 should be enough for everybody? ;)

src/radical/pilot/agent/radical-pilot-agent-multicore.py
```diff
@@ -1170,6 +1179,190 @@ def corner2offset(self, block, corner):
...
+
+        return True
+
+    # --------------------------------------------------------------------------
+    #
+    def work(self, cu):
+
+      # self.advance(cu, rp.AGENT_SCHEDULING, publish=True, push=False)
+        self._log.info("Overiding Parent's class method")
+        self.advance(cu, rp.ALLOCATING      , publish=True, push=False)
```
andre-merzky: Who would have though I'd ever say this: there are too many spaces!  You may want to distribute them elsewhere... ;)

src/radical/pilot/agent/radical-pilot-agent-multicore.py
```diff
@@ -1299,6 +1494,48 @@ def lrms_shutdown_hook(cls, name, cfg, lrms, lm_info, logger):
...
+          # LAUNCH_METHOD_RUNJOB        : LaunchMethodRUNJOB,
+          # LAUNCH_METHOD_SSH           : LaunchMethodSSH
+        }.get(name)
+
+        if not impl:
+            logger.info('no LRMS config hook defined for LaunchMethod %s' % name)
+            return None
+
+        logger.info('call LRMS config hook for LaunchMethod %s: %s' % (name, impl))
+        return impl.lrms_final_hook(lm_info, logger)
```
andre-merzky: Does the finalization hook work out for you?  Also, you can remove one of the empty lines below...

src/radical/pilot/agent/radical-pilot-agent-multicore.py
```diff
@@ -1326,6 +1563,7 @@ def _find_executable(cls, names):
         for name in names:
             ret = cls._which(name)
             if ret is not None:
+                self._log.info("Found %s"%ret)
```
andre-merzky: As a general rule: for logging commands, please use
```
logger.info('msg: %s', msg)
```
instead of
```
logger.info('msg: %s' % msg)
```
This avoids Python's string expansion to kick in if logging is disabled (and the strings would thus not be used anyway).  We don't do that everywhere either, but in general try to adhere to that rule...  Thanks!

src/radical/pilot/agent/radical-pilot-agent-multicore.py
```diff
@@ -1972,8 +2211,8 @@ def lrms_config_hook(cls, name, cfg, lrms, logger):
                the DVM.
         """
 
-        dvm_command = cls._which('orte-dvm')
-        if not dvm_command:
+        dvm_cmd = cls._which('orte-dvm')
+        if not dvm_cmd:
```
andre-merzky: Uhm, why the rename?  Now `dvm_command` is not set further below...  I don't mind the shorter name at all, but then it needs to be changed everywhere...

src/radical/pilot/agent/radical-pilot-agent-multicore.py
```diff
@@ -2204,6 +2443,381 @@ def construct_command(self, task_exec, task_args, task_numcores,
...
+
+        #-----------------------------------------------------------------------
+        # Create YARN script
+        # This funcion creates the necessary script for the execution of the
+        # CU's workload in a YARN application. The function is responsible
+        # to set all the necessary variables, stage in, stage out and create
+        # the execution command that will run in the distributed shell that
+        # the YARN application provides. There reason for staging out is
+        # because after the YARN application has finished everything will be
+        # deleted.
```
andre-merzky: I may be missing it, but is the CU `environment` set anywhere in the resulting script?

src/radical/pilot/agent/radical-pilot-agent-multicore.py
```diff
@@ -2344,6 +2959,25 @@ def __init__(self, cfg, logger):
...
+        #
+        # That list may turn out to be insufficient for some schedulers.  Yarn
+        # for example may need to communicate YARN service endpoints etc.  an
+        # LRMS can thus expand this dict, but is then likely bound to a specific
+        # scheduler which can interpret the additional information.
+        self.lrms_info['lm_info']        = self.lm_info
+        self.lrms_info['node_list']      = self.node_list
+        self.lrms_info['cores_per_node'] = self.cores_per_node
+        self.lrms_info['agent_nodes']    = self.agent_nodes
+
```
andre-merzky: I am not sure I understand this  diff.  (a): all those details are already on the `self._cfg` I believe, why is it stored again?   (b) where is self.lrms_info used?  I can't find it referenced...

I think this setting is not needed, since (as you say yourself) the information is passed via the cfg?

src/radical/pilot/agent/radical-pilot-agent-multicore.py
```diff
@@ -3769,8 +4464,12 @@ def _populate_cu_environment(self):
     #
     def work(self, cu):
 
-      # self.advance(cu, rp.AGENT_EXECUTING, publish=True, push=False)
-        self.advance(cu, rp.EXECUTING, publish=True, push=False)
+        if self._task_launcher.name !='LaunchMethodYARN':
+          # self.advance(cu, rp.AGENT_EXECUTING, publish=True, push=False)
+            self.advance(cu, rp.EXECUTING,       publish=True, push=False)
```
andre-merzky: spaces :)

src/radical/pilot/agent/radical-pilot-agent-multicore.py
```diff
@@ -3873,8 +4580,19 @@ def spawn(self, launcher, cu):
 
             # The actual command line, constructed per launch-method
             try:
-                launch_command, hop_cmd = \
-                    launcher.construct_command(cu['description']['executable'],
+                if launcher.name == 'LaunchMethodYARN':
```
andre-merzky: We should  go through all parts of the code where 'YARN' is checked -- but this one specifically: the construct_command signature should be kept consistent.  Lets discuss this with Mark.

src/radical/pilot/agent/radical-pilot-agent-multicore.py
```diff
@@ -3908,9 +4626,18 @@ def spawn(self, launcher, cu):
                 if 'RADICAL_PILOT_PROFILE' in os.environ:
                     launch_script.write("echo post stop  `%s` >> %s/PROF\n" % (cu['gtod'], cu_tmpdir))
 
+            # YARN pre execution folder permission change
+            # TODO: This needs to move inside the construct command when the launcher
+            #       takes only the CU description
+            if launcher.name == 'LaunchMethodYARN':
```

src/radical/pilot/agent/radical-pilot-agent-multicore.py
```diff
@@ -3908,9 +4626,18 @@ def spawn(self, launcher, cu):
...
+                launch_script.write('\n## Changing Working Directory permissions for YARN\n')
+                launch_script.write('chmod $old_perm .\n')
+
             launch_script.write("# Exit the script with the return code from the command\n")
             launch_script.write("exit $RETVAL\n")
 
+
```
andre-merzky: ?

src/radical/pilot/agent/radical-pilot-agent-multicore.py
```diff
@@ -4006,65 +4733,109 @@ def _check_running(self):
         action = 0
 
         for cu in self._cus_to_watch:
+            #-------------------------------------------------------------------
+            # This code snippet reads the YARN application report file and if
+            # the application is RUNNING it update the state of the CU with the
+            # right time stamp. In any other case it works as it was.
+            if self._task_launcher.name == 'LaunchMethodYARN' \
```
andre-merzky: Same here: we should rethink our abstractions if Yarn requires so massive changes to the spawner.  Too much code redundency, and too much dependency no the LaunchMethod.  Maybe an evaluation hook on the launcher would help?  Like, `lm.create_command()`  before spawning, `lm.evaluate_result()` after collecting the job?

src/radical/pilot/agent/radical-pilot-agent-multicore.py
```diff
@@ -6111,11 +6918,7 @@ def sigalarm_handler(signum, frame):
             lrms = LRMS.create(name   = cfg['lrms'],
                                cfg    = cfg,
                                logger = log)
-            cfg['lrms_info'] = dict()
-            cfg['lrms_info']['lm_info']        = lrms.lm_info
-            cfg['lrms_info']['node_list']      = lrms.node_list
-            cfg['lrms_info']['cores_per_node'] = lrms.cores_per_node
-            cfg['lrms_info']['agent_nodes']    = lrms.agent_nodes
+            cfg['lrms_info'] = lrms.lrms_info
```
andre-merzky: Ah, that is what you moved upward I guess...  That may make sense...

src/radical/pilot/agent/radical-pilot-agent-multicore.py
andre-merzky: Note to self: this is a merge problem, AM to address.

src/radical/pilot/agent/radical-pilot-agent-multicore.py
andre-merzky: AM: check where this comes from...

iparask:
Separate YARN spanwer from POPEN

src/radical/pilot/agent/radical-pilot-agent-multicore.py
iparask: Note to me. Find how the CU environment passes to the YARN application.

src/radical/pilot/agent/radical-pilot-agent-multicore.py
iparask: This is not mine. Unless you mean the empty line.

andre-merzky: Yep, the empty one :)

src/radical/pilot/agent/radical-pilot-agent-multicore.py
```diff
@@ -152,12 +152,17 @@ class UpdateWorker
 import traceback
 import subprocess
 import multiprocessing
+import commands
```
marksantcroos: This is deprecated.

src/radical/pilot/agent/radical-pilot-agent-multicore.py
```diff
@@ -1170,6 +1179,190 @@ def corner2offset(self, block, corner):
...
+
+            self._log.info('Checking rm_ip %s' % self._cfg['lrms_info']['lm_info']['rm_ip'])
+            self._rm_ip = self._cfg['lrms_info']['lm_info']['rm_ip']
+            self._service_url = self._cfg['lrms_info']['lm_info']['service_url']
+            self._rm_url = self._cfg['lrms_info']['lm_info']['rm_url']
+            self._client_node = self._cfg['lrms_info']['lm_info']['nodename']
+
+            sample_time = rpu.timestamp()
+            yarn_status = ul.urlopen('http://{0}:8088/ws/v1/cluster/scheduler'.format(self._rm_ip))
+
```
marksantcroos: Can this url be parametrized in some way?

src/radical/pilot/agent/radical-pilot-agent-multicore.py
```diff
@@ -1170,6 +1179,190 @@ def corner2offset(self, block, corner):
...
+            yarn_schedul_json = json.loads(yarn_status.read())
+
+            max_num_app = yarn_schedul_json['scheduler']['schedulerInfo']['queues']['queue'][0]['maxApplications']
+            num_app = yarn_schedul_json['scheduler']['schedulerInfo']['queues']['queue'][0]['numApplications']
+
+            #-----------------------------------------------------------------------
+            # Find out the cluster's resources
+            cluster_metrics = ul.urlopen('http://{0}:8088/ws/v1/cluster/metrics'.format(self._rm_ip))
+
```
marksantcroos: Same here.

src/radical/pilot/agent/radical-pilot-agent-multicore.py
```diff
@@ -1170,6 +1179,190 @@ def corner2offset(self, block, corner):
...
+        # As it seems this part of the Scheduler is not according to the assumptions
+        # made about slot status. Keeping the code commented just in case it is
+        # needed later either as whole or art of it.
+        sample = rpu.timestamp()
+        yarn_status = ul.urlopen('http://{0}:8088/ws/v1/cluster/scheduler'.format(self._rm_ip))
+        yarn_schedul_json = json.loads(yarn_status.read())
+
+        max_num_app = yarn_schedul_json['scheduler']['schedulerInfo']['queues']['queue'][0]['maxApplications']
+        num_app = yarn_schedul_json['scheduler']['schedulerInfo']['queues']['queue'][0]['numApplications']
+        if (self.avail_app['timestamp'] - sample)>60 and \
```
marksantcroos: 60?

src/radical/pilot/agent/radical-pilot-agent-multicore.py
```diff
@@ -2204,6 +2443,381 @@ def construct_command(self, task_exec, task_args, task_numcores,
...
+        """
+        FIXME: this config hook will inspect the LRMS nodelist and, if needed,
+               will start the YRN cluster on node[0].
+        """
+
+        logger.info('Hook called by YARN LRMS with the name %s'%lrms.name)
+
+        def config_core_site(node):
+
+            core_site_file = open(os.getcwd()+'/hadoop/etc/hadoop/core-site.xml','r')
```
marksantcroos: Parametrize.

src/radical/pilot/agent/radical-pilot-agent-multicore.py
```diff
@@ -2204,6 +2443,381 @@ def construct_command(self, task_exec, task_args, task_numcores,
...
+            lines = core_site_file.readlines()
+            core_site_file.close()
+
+            prop_str  = '<property>\n'
+            prop_str += '  <name>fs.default.name</name>\n'
+            prop_str += '    <value>hdfs://%s:54170</value>\n'%node
```
marksantcroos: Is that a default port or something?

src/radical/pilot/agent/radical-pilot-agent-multicore.py
```diff
@@ -2204,6 +2443,381 @@ def construct_command(self, task_exec, task_args, task_numcores,
...
+            prop_str += '</property>\n'
+
+            prop_str += '<property>\n'
+            prop_str += '  <name>dfs.data.dir</name>\n'
+            prop_str += '    <value>file:///tmp/hadoop/hadoopdata/hdfs/datanode</value>\n'
+            prop_str += '</property>\n'
+
+            lines.insert(-1,prop_str)
+
+            hdfs_site_file = open(os.getcwd()+'/hadoop/etc/hadoop/hdfs-site.xml','w')
```
marksantcroos: Parametrize.

src/radical/pilot/agent/radical-pilot-agent-multicore.py
```diff
@@ -2204,6 +2443,381 @@ def construct_command(self, task_exec, task_args, task_numcores,
...
+            rm_url         = "%s:%s" % (lrms.rm_ip, lrms.rm_port)
+            rm_ip          = lrms.rm_ip
+            launch_command = cls._which('yarn')
+
+        else:
+            # Here are the necessary commands to start the cluster.
+            if lrms.node_list[0] == 'localhost':
+                #Download the tar file
+                node_name = lrms.node_list[0]
+                stat = os.system("wget http://apache.claz.org/hadoop/common/hadoop-2.6.0/hadoop-2.6.0.tar.gz")
```
marksantcroos: Parametrize.

src/radical/pilot/agent/radical-pilot-agent-multicore.py
```diff
@@ -2204,6 +2443,381 @@ def construct_command(self, task_exec, task_args, task_numcores,
...
+                stat = os.system('tar xzf hadoop-2.6.0.tar.gz;mv hadoop-2.6.0 hadoop;rm -rf hadoop-2.6.0.tar.gz')
+            else:
+                node = commands.getstatusoutput('/bin/hostname')
```
marksantcroos: Module commands is deprecated.

src/radical/pilot/agent/radical-pilot-agent-multicore.py
```diff
@@ -2204,6 +2443,381 @@ def construct_command(self, task_exec, task_args, task_numcores,
...
+                logger.info('Entered Else creation')
+                node_name = node[1]
+                stat = os.system("wget http://apache.claz.org/hadoop/common/hadoop-2.6.0/hadoop-2.6.0.tar.gz")
+                stat = os.system('tar xzf hadoop-2.6.0.tar.gz;mv hadoop-2.6.0 hadoop;rm -rf hadoop-2.6.0.tar.gz')
```
marksantcroos: The return values of the above calls are not checked.

src/radical/pilot/agent/radical-pilot-agent-multicore.py
```diff
@@ -2204,6 +2443,381 @@ def construct_command(self, task_exec, task_args, task_numcores,
...
+            logger.info('Stoping YARN')
+            os.system(lm_info['hadoop_home'] + '/sbin/stop-yarn.sh')
+
+            logger.info('Stoping DFS.')
+            os.system(lm_info['hadoop_home'] + '/sbin/stop-dfs.sh')
+
+            logger.info("Deleting HADOOP files from temp")
+            os.system('rm -rf /tmp/hadoop*')
+            os.system('rm -rf /tmp/Jetty*')
+            os.system('rm -rf /tmp/hsperf*')
```
marksantcroos: You might want to be a bit more specific with these rm's.

src/radical/pilot/agent/radical-pilot-agent-multicore.py
```diff
@@ -3845,6 +4544,14 @@ def spawn(self, launcher, cu):
                 if 'RADICAL_PILOT_PROFILE' in os.environ:
                     launch_script.write("echo pre  stop `%s` >> %s/PROF\n" % (cu['gtod'], cu_tmpdir))
 
+            # YARN pre execution folder permission change
+            # TODO: This needs to move inside the construct command when the launcher
+            #       takes only the CU description
+            if launcher.name == 'LaunchMethodYARN':
+                launch_script.write('\n## Changing Working Directory permissions for YARN\n')
+                launch_script.write('old_perm="`stat -c %a .`"\n')
+                launch_script.write('chmod 777 .\n')
```
marksantcroos: 777 is probably not the right choice.

iparask: Here the reason: The YARN container that executes the workload is under the user that Hadoop and Yarn has. It is different from the user that launches the Application. Somehow that user should be able to write to the folder of each CU. That means 2 things. First that the YARN user should have access to the node that the application was launched and 2nd access to the folder of the CU.

There is no guarantee that the user who will run the YARN application will be even in the same group as the YARN user.

Also this 777 changes to the old permission number.

Of course if you have another suggestion I am happy to use it and see how it goes

src/radical/pilot/agent/radical-pilot-agent-multicore.py
iparask: After the YARN and Hadoop cluster is closed it leaves behind all these folders. the reason for the rms is mainly to clean the node that the YARN cluster was created and run.

This is in the case where RP launches YARN and Hadoop.

src/radical/pilot/agent/radical-pilot-agent-multicore.py
iparask: In case they did not run?

Thank you, I will include them

src/radical/pilot/agent/radical-pilot-agent-multicore.py
iparask: Ok, but I need the value the hostname command returns.

That applies to anywhere else I have used commands

src/radical/pilot/agent/radical-pilot-agent-multicore.py
iparask: Except from the actual URL everything else is standard in order to access the status of the YARN scheduler. Maybe, the port can be parametrized, but I have to check first.

src/radical/pilot/agent/radical-pilot-agent-multicore.py
iparask: Is there an alternative?

src/radical/pilot/agent/radical-pilot-agent-multicore.py
iparask: As above

src/radical/pilot/agent/radical-pilot-agent-multicore.py
iparask: I wanted to introduce a delay between updates without using sleep.

If there is another idea, I am all ears!

src/radical/pilot/agent/radical-pilot-agent-multicore.py
iparask: So, the hadoop tar ball has been downloaded and extracted. The xml files that are written are necessary to set the environment for spawning the YARN cluster.

These path are very specific. I do not see a way how to parametrize it.

src/radical/pilot/agent/radical-pilot-agent-multicore.py
iparask: Something. It needs to be a free port. I selected it just because it is not the default and I am afraid that the default port may be monitored, in Stampede for example

src/radical/pilot/agent/radical-pilot-agent-multicore.py
iparask: Again, this is a bit difficult. We create our own YARN cluster, which means that we need to set the above paths. They can be parametrized after the /tmp folder.

Is there a specific reason for making them parametric?

src/radical/pilot/agent/radical-pilot-agent-multicore.py
iparask: I do not understand here what you believe needs to be parametrized

src/radical/pilot/agent/radical-pilot-agent-multicore.py
andre-merzky: How does  that work for multiple users, in those cases where the agent runs on a head node?  They can't all end up in the same dir structure, can they?

src/radical/pilot/agent/radical-pilot-agent-multicore.py
andre-merzky: yes: http://stackoverflow.com/questions/89228/calling-an-external-command-in-python

src/radical/pilot/agent/radical-pilot-agent-multicore.py
andre-merzky: I am not   sure I understand that: the part you shield with the timestamp check is one substraction and two assignment operators (and two hash lookups I guess) -- is that worth it?  Should be cheap enough to simply call every time?

src/radical/pilot/agent/radical-pilot-agent-multicore.py
andre-merzky: What   will happen if two users  run an agent at the same time on the headnode?

src/radical/pilot/agent/radical-pilot-agent-multicore.py
iparask: This shutdown hook is called when we have launched the YARN cluster, e.g. on Stampede. I believe it runs on the compute node, does it?

If that is correct, the /tmp folder are the local hard drives of the node, as we know. So if two user run the same configuration they will be on diferent compute nodes thus different /tmps. 

Am I understanding something wrong?

src/radical/pilot/agent/radical-pilot-agent-multicore.py
iparask: On the head node? This is the hook that launches the YARN cluster. Doesn't it run on the compute node?

If not there will be a conflict apparently and I need to think about it.

andre-merzky: The agent is not always  running on a compute note.  In fact, that is an exception at the moment...  In general it might be a head-, mom- or compute node.  Even on compute nodes it is not necessarily safe to assume that they are single user (even though that it is usually the case...).  Mom-nodes are also multi-user...

iparask:
I found it. The construct command for YARN has been updated but not for the rest of the launchers. Shouldn't be be updated or passing the whole cu to the launcher is just for YARN and I should change the POPEN spanwer?

andre-merzky:
The `construct_command` signature should be the same for all...

andre-merzky:
Actually, I don't think I see any different signatures:

```
$ grep def\ construct_command src/radical/pilot/agent/radical-pilot-agent-multicore.py 
    def construct_command(self, cu, launch_script_hop):
    def construct_command(self, cu, launch_script_hop):
    def construct_command(self, cu, launch_script_hop):
    def construct_command(self, cu, launch_script_hop):
    def construct_command(self, cu, launch_script_hop):
    def construct_command(self, cu, launch_script_hop):
    def construct_command(self, cu, launch_script_hop):
    def construct_command(self, cu, launch_script_hop):
    def construct_command(self, cu, launch_script_hop):
    def construct_command(self, cu, launch_script_hop):
    def construct_command(self, cu, launch_script_hop):
    def construct_command(self, cu, launch_script_hop):
    def construct_command(self, cu, launch_script_hop):
    def construct_command(self, cu, launch_script_hop):
    def construct_command(self, cu, launch_script_hop):
    def construct_command(self, cu, launch_script_hop):
```

What am I missing?

iparask:
Take a look here: https://github.com/radical-cybertools/radical.pilot/blob/feature/yarn/src/radical/pilot/agent/radical-pilot-agent-multicore.py#L1630

[task_exec, task_args] creates something like this [ , []] because task_args is a list. I can easily be solved by doing task_args.insert(0,task_exec) before the join but I do not know if is right based on the abstractions. At least it works

marksantcroos:
> because task_args is a list.

It is considered a string. Where is it a list?

marksantcroos:
> > because task_args is a list.
> It is considered a string. Where is it a list?

So its not anymore indeed. The signature branch changed the semantics.

marksantcroos:
> The signature branch changed the semantics.

Btw, thats an euphemism for 'breaks things'.

iparask:
Hahahaha!!!

http://radicalpilot.readthedocs.org/en/stable/apidoc.html#computeunitdescription

It says a list of strings and I think that is where it should change.

iparask:
So what needs to be done?

marksantcroos:
> http://radicalpilot.readthedocs.org/en/stable/apidoc.html#computeunitdescription

This has nothing to do with the API.

iparask:
Okay! Can you give it a try? It is either that or my last merge with devel went extremely wrong

marksantcroos:
> So what needs to be done?

Either revert/modify the signature change, or do the the conversion in every LM.

marksantcroos:
See https://github.com/radical-cybertools/radical.pilot/blob/devel/src/radical/pilot/agent/radical-pilot-agent-multicore.py#L3860 where the expansion is done and https://github.com/radical-cybertools/radical.pilot/blob/devel/src/radical/pilot/agent/radical-pilot-agent-multicore.py#L3878 where it is passed (in devel).

So in the yarn branch, at https://github.com/radical-cybertools/radical.pilot/blob/feature/yarn/src/radical/pilot/agent/radical-pilot-agent-multicore.py#L4629 the task_args_string is still constructed, but not used.

marksantcroos:
So its unlikely that that signature change could have worked.

andre-merzky:
at long last...

PR #845, opened 2015-10-17
Feature/double shutdown 2

andre-merzky:
Mark, I hope you forgive the reduced diff readability, and appreciate the pep-piness :)

### DIFF
--- a/src/radical/pilot/controller/pilot_launcher_worker.py
+++ b/src/radical/pilot/controller/pilot_launcher_worker.py
@@ -306,6 +306,11 @@ class PilotLauncherWorker(threading.Thread):
 
                 if self._disabled.is_set():
                     # don't process any new pilot start requests.  
+                    # NOTE: this is not clean, in principle there could be other
+                    #       launchers alive which want to still start those 
+                    #       pending pilots.  In parctive we only ever use one
+                    #       pmgr though, and its during its shutdown that we get
+                    #       here...
                     ts = timestamp()
                     compute_pilot = pilot_col.find_and_modify(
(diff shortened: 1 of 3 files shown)


src/radical/pilot/controller/pilot_launcher_worker.py
```diff
@@ -306,6 +306,19 @@ def run(self):
 
                 if self._disabled.is_set():
                     # don't process any new pilot start requests.  
+                    # NOTE: this is not clean, in principle there could be other
+                    #       launchers alive which want to still start those 
+                    #       pending pilots.  In practice we only ever use one
+                    #       pmgr though, and its during its shutdown that we get
+                    #       here...
```
marksantcroos: To be clear, we are not ruling out the use of two pmgr's here are we?

andre-merzky: No, I don't think so.  But in principle two launch controllers could be used for one pmgr, and this commit explicitly breaks that.  Well, OTOH, this is already broken in other places, so it doesn't change much :/

marksantcroos: > ut in principle two launch controllers could be used for one pmgr, and this commit explicitly breaks that.

So as long as one doesn't explicitly instantiate the PM with multiple workers all is fine?

andre-merzky: Yes

...

marksantcroos:
Thanks! Looking forward to the refactoring though :-)

PR #847, opened 2015-10-17
Feature=/install examples

andre-merzky:
We do have downloadable links in the docs, I know, and should keep them.  But then again, I think its a good thing to have the examples (and possible later also the docs and tests, FWIW) coming along with the installation.


andre-merzky:
I should mention that this is specifically useful when we do the CECAM tutorial on the workflow machine, as a  downloadable link does not help much -- its wget-cut-and-paste, or download-plus-scp.  That will be annoying for 10+ example scripts + configs etc.

marksantcroos:
Thanks!

PR #848, opened 2015-10-18
Feature=/install examples

andre-merzky:
Here we   go again :/


marksantcroos:
Hmm, if I wouldn't know better I ....

PR #864, opened 2015-10-21
Allow Session to function as context manager for `with`

eirrgang:
defines `Session.__enter__()` and `Session.__exit__()` so that `Session.close()` will be called when exiting a `with` clause. E.g.

    with Session() as session:
        # do stuff with session
    # session.close() has been called


marksantcroos:
Thanks!

PR #868, opened 2015-10-22
change construct_command signature as discussed

andre-merzky:
Mark, Iannis, could you please both look over this?  Thanks.

### DIFF
--- a/src/radical/pilot/agent/radical-pilot-agent-multicore.py
+++ b/src/radical/pilot/agent/radical-pilot-agent-multicore.py
@@ -1545,8 +1545,7 @@ class LaunchMethod(object):
 
     # --------------------------------------------------------------------------
     #
-    def construct_command(self, task_exec, task_args, task_numcores,
-                          launch_script_hop, opaque_slots):
+    def construct_command(self, cu, launch_script_hop, opaque_slots):
         raise NotImplementedError("construct_command() not implemented for LaunchMethod: %s." % self.name)
 
 
@@ -1620,15 +1619,20 @@ class LaunchMethodFORK(LaunchMethod):
 
     # --------------------------------------------------------------------------
     #
-    def construct_command(self, task_exec, task_args, task_numcores,
-                          launch_script_hop, opaque_slots):
+    def construct_command(self, cu, launch_script_hop):
+
+        opaque_slots = cu['opaque_slots']
+        cud          = cu['description']
+        task_exec    = cud['executable']
+        task_args    = cud.get('arguments')
+        task_cores   = cud.get('cores', 1)
 
         if task_args:
             command = " ".join([task_exec, task_args])
         else:
             command = task_exec
 
-        self._log.info('LaunchMethodFORK returns command : %s'%command)
+        self._log.info('LaunchMethodFORK returns command : %s', command)
         return command, None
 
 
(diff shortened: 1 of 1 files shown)


src/radical/pilot/agent/radical-pilot-agent-multicore.py
```diff
@@ -1620,15 +1619,20 @@ def lrms_config_hook(cls, name, cfg, lrms, logger):
...
     #
-    def construct_command(self, task_exec, task_args, task_numcores,
-                          launch_script_hop, opaque_slots):
+    def construct_command(self, cu, launch_script_hop):
+
+        opaque_slots = cu['opaque_slots']
+        cud          = cu['description']
+        task_exec    = cud['executable']
+        task_args    = cud.get('arguments')
+        task_cores   = cud.get('cores', 1)
```
marksantcroos: I rather have it fail here if it is unset.

src/radical/pilot/agent/radical-pilot-agent-multicore.py
```diff
@@ -6544,11 +6592,16 @@ def start_sub_agents(self):
                         'task_slots'   : ['%s:0' % node], 
                         'task_offsets' : [], 
                         'lm_info'      : self._cfg['lrms_info']['lm_info']}
-                cmd, hop = agent_lm.construct_command(task_exec="/bin/sh",
-                        task_args="%s/bootstrap_2.sh %s" % (os.getcwd(), sa),
-                        task_numcores=1, 
-                        launch_script_hop='/usr/bin/env RP_SPAWNER_HOP=TRUE "%s"' % ls_name,
-                        opaque_slots=opaque_slots)
+                agent_cu = {
```
marksantcroos: `agent_cmd` (or some variation?). Its not a CU.

src/radical/pilot/agent/radical-pilot-agent-multicore.py
andre-merzky: the cores, not the args, right?  Yeah, makes sense - let me change that...

andre-merzky:
Iannis, I have a general question about the spawner / launch_method:  basically, the launcher will create a command like

```
yarn -jar ... -ShellScript ExecScript.sh... \n cat Ystdout
```

Will yarn command (the first part) return after yarn received the job, or will it terminate?  I assume the Ystdout remains alive   forever, ie. until you later detect that the work is done and then you cancel it?

Those commands are always running on the agent node, AFAIKC - is that correct?

If the latter is correct, then there is actually no need to invoke the Popen or Shell spawners -- those add unneccessary complexity for something which always goes to the agent node.  What I would propose is to create a Yarn spawner, and to trim that down significantly.  That would achieve two things:

  * make the yarn code path simpler
  * avoid the many `if yarn` switches (I think it does away with almost all of them).

src/radical/pilot/agent/radical-pilot-agent-multicore.py
andre-merzky: I would actually argue that at this point it is a CU, kind of, but I can see that one can argue both ways.  I don't mind changing it, will do.

marksantcroos: More "kind of" than "is" :)
But thanks for the change!
Given that its now also used for the agents, it should ideally become CU/agent neutral.

iparask:
The YARN command returns when the whole execution of the application has finished for any reason, either success or failure. The Ystdout is the log file created inside the YARN application and it is downloaded when the application has finished. Also it is not created in the agent node, but all the yarn commands are executed in the agent node, which is also the entry point for the YARN scheduler.

A YARN spawner is required as it looks like.

andre-merzky:
Well, maybe not required, but it will better encapsulate things indeed...

Issue #640, opened 2015-06-10
RP fails on rice.biou: pip not found.


vivek-bala:
Adding ```module load python/2.7.6``` in https://github.com/radical-cybertools/radical.pilot/blob/master/src/radical/pilot/configs/rice.json#L42 fixes it.

Default is python/2.6.6

andre-merzky:
We'll pick this up once rice is used again.

andre-merzky:
We don't have access to the machine anymore AFAIK.

PR #871, opened 2015-10-23
Load python module.

marksantcroos:
Address #640.


andre-merzky:
I am not sure why you remove the lines ...

PR #873, opened 2015-10-23
Build sdist in bdist_wheel

0:
With this change, the sdist is included in the wheel when I build radical.pilot using `pip wheel`. I suspect that a similar change would work for radical.utils and saga-python.


andre-merzky:
Hi Dmitri, 

thanks for the PR, I'm glad you found this!  

Best, Andre.

PS.: nice user name ;)

PR #874, opened 2015-10-25
Feature/anaconda setup

iparask:
I removed commands module. The change from PR #873 worked like a charm. I believe that the rest are a good starting point. It works in localhost when Anaconda is present.


marksantcroos:
Thanks!
Given that there is a switch now, why can't the switch be used to switch within the bootstrapper, instead of duplicating 95% of the code?

andre-merzky:
I think it was my  suggestion to create a new bootstrapper.  I guess at the end it will depends on how similar they'll be in terms of code: if the current bootstrapper ends up with `if/then/else` all over the place I'd prefer a new one.  If the diff at the end is minimal, then I see no problem in having one universal one.

My impression was that anaconda differs fundamentally in how virtualenvs are created / managed, and how modules are installed.  Given that the bootstrapper code is mostly about exactly that, I would think that a new bootstrapper would make sense.  Maybe my assumptions are flawed?

src/radical/pilot/bootstrapper/default_bootstrapper_anaconda.sh
```diff
@@ -0,0 +1,1459 @@
...
+#!/bin/sh
+
+# some inspection for logging
+hostname
+
+# make sure we use the correct sandbox
+cd $SANDBOX
+
+# activate virtenv
+source activate $VIRTENV
```
andre-merzky: How can that possibly work?  Does an `activate` script exist in `$SANDBOX`?  Where does it come from?

marksantcroos:
Ok, I'll take that back, its 92% instead of 95%.

```
$ wc -l default_bootstrapper.sh 
    1459 default_bootstrapper.sh
$ wc -l others 
    1331 others
```

src/radical/pilot/bootstrapper/default_bootstrapper_anaconda.sh
iparask: That is how it is done! I do not know more yet.

iparask:
That 3% made the difference!! :P 

I agree it can be one bootstrapper. But I am not sure how it is should be done....

And we need something stable for the next tutorial

src/radical/pilot/bootstrapper/default_bootstrapper_anaconda.sh
andre-merzky: I don't get it.  `source` is a shell command which does not respect `$PATH` or anything else for its arguments.  So `activate` needs to be in the dir the `source` command is called in.  That seems to be `$SANDBOX` per 3 lines above.  How did the `activate` command get there?

iparask: So there is an `activate` script in the path of the virtual env. `$VIRTENV` has the whole path pointing to the environment. It really does no matter from where you run the commands. I can paste the activate script if you want.

I assume that source looks at the VE path for activate and runs it.

andre-merzky: I trust you that it works -- it is just that I am puzzled...  I was wrong BTW wrt. the `PATH` interpretation: source *does* search in `$PATH`.  But I don't see where `$PATH` is set to include the virtenv -- that looks like a chicken and egg problem to me... :P  

Anyway, if it works, it works, no worries!

Thanks, Andre.

andre-merzky:
> I am not sure how it is should be done....

You should leave the config entry in place -- but the PilotLauncher component needs to forward that information to the bootstrapper then.  For that you likely want to introduce a new command line flag, based on that you'd set a variable in the bootstrapper, and then `if/then/else/fi` on all relevant places...

iparask:
Ah! Okay I will give it a try!

andre-merzky:
Hey Iannis -- I made a  couple of small  changes, please check if that still works as expected with Anaconda.  There is one semantic change: `bootstrap_2.sh` now   uses `/bin/bash` instead of `/bin/bash -l` -- was there a specific reason to use a login shell at that point?

marksantcroos:
Let me give this a try on SuperMUC.

andre-merzky:
ack, thanks!

iparask:
As far as I am concerned ```-l``` does not matter. It still works

andre-merzky:
> As far as I am concerned -l does not matter. It still works

Ok, good.  In general, we try to stay away from `-l`, as it pulls in user specific settings which make it hard to reproduce the environment on problems...

andre-merzky:
Mark, can you give it a run on supermuc, please?  Thanks!

src/radical/pilot/bootstrapper/default_bootstrapper.sh
```diff
@@ -1210,6 +1229,7 @@ mkdir -p "$VIRTENV"
 echo "VIRTENV : $VIRTENV"
 VIRTENV=`(cd $VIRTENV; pwd -P)`
 echo "VIRTENV : $VIRTENV (normalized)"
+rmdir "$VIRTENV" 2>/dev/null
```
marksantcroos: Whats the rationale for this? To remove it if it is empty?

andre-merzky: It's complicated :P

pip/python do not play together well when one is using a symlinked path, and the other isn't.  So we try to get the normalized path for the virtenv location (`pwd -P`).  If the virtenv exists, that is not a problem - if it does not, it is.  So we create the dir (`mkdir -p` will not complain if it exists), do our check, and remove it again.  The removal will fail on a non-empty dir, ie. on an existing virtenv, so we redirect the error.

An alternative approach would be to do an `if test -d` ...

andre-merzky:
Thanks Iannis, this seems to be working as expected so far! :D

Issue #875, opened 2015-10-25
check if all incarnations of python in bootstrapper should be $PYTHON


marksantcroos:
Post-papers though.

PR #876, opened 2015-10-26
python -> $PYTHON

andre-merzky:
adresses #875


marksantcroos:
Are you intentionally not including the ones in lines 668 and 669?

andre-merzky:
> Are you intentionally not including the ones in lines 668 and 669?

No, that was not intentional, so thanks!  I added those two, and can't find any others now.  I consider this fit for merge.

PR #877, opened 2015-10-26
fix handling of task_args

andre-merzky:
Lets see if I got it right this time :)


andre-merzky:
will be rerouted toward devel

PR #882, opened 2015-10-29
Fix/state ordering

andre-merzky:
## enforce unit state ordering before pushing db updates
This removes the sleep-based 'ordering'.  Also in this commit:

  - subscriber threads are now watched by the component, and failures propagate
  - logging noise decreases for non-blowup scenarios
  - rename: component._threads -> component._subscribers

This needs thorough testing.  The code also needs to kept in sync with state
model evolution, and should be moved to the module side after refactoring.
Until that happens, there can only be one instance of an UpdateWorker per agent
instance.


marksantcroos:
Thanks! Its holding up pretty well!

andre-merzky:
glad (and relieved) to hear that!

Issue #881, opened 2015-10-28
replace sleep at end of bootstrapper with something more adaptive


marksantcroos:
See discussion in https://github.com/radical-cybertools/radical.pilot/commit/f67d0dd9f5cecd6d260b418d69a311335a5ef508#commitcomment-14052647

andre-merzky:
There is an adaptive sleep now.

PR #883, opened 2015-10-29
address #881 (sleep before tar)

andre-merzky:


src/radical/pilot/bootstrapper/default_bootstrapper.sh
```diff
@@ -1426,10 +1426,27 @@ then
...
-    echo "# Sleeping for $FINAL_SLEEP seconds to let dust settle ..."
+    echo "# We wait for at most 30 seconds for the FS to flush profiles."
+    echo "# Success is assumed when all profiles end with a 'QED' event."
     echo "#"
     echo "# -------------------------------------------------------------------"
-    sleep $FINAL_SLEEP
+    nprofs=`echo *.prof | wc -w`
+    nqed=`tail -n 1 *.prof | grep QED | wc -l`
+    nsleep=0
+    while ! test "$nprofs" = "$nqed"
```
marksantcroos: `while test  "$nqed" -lt "$nprofs"`?

(For the case where there are multiple QEDs per prof?)

src/radical/pilot/bootstrapper/default_bootstrapper.sh
```diff
@@ -1426,10 +1426,27 @@ then
...
+    do
+        nsleep=$((nsleep+1))
+        if test "$nsleep" = "30"
+        then
+            echo "abort profile sync @ $nsleep: $nprofs != $nqed"
+            break
+        fi
+        echo "delay profile sync @ $nsleep: $nprofs != $nqed"
+        sleep 1
+        # recheck nprofs too, just in case...
```
marksantcroos: Good idea. I saw them dripping in.

src/radical/pilot/bootstrapper/default_bootstrapper.sh
andre-merzky: Yeah, I guess that makes sense.  But with that assumption we may get false positives here:
```
a.prof:0
b.prof:2
```
would count as success...

marksantcroos: Hmmm, wait. I might have been misinterpreting that. Let me look a bit closer later.

marksantcroos: Right, you mention a "count" output, but we don't use that. 
So my initial argumentation was wrong, and although it might be semantically nicer, the test for equal might catch other goof ups, so lets keep it like this.

PR #884, opened 2015-10-31
get signature variable name in sync with type

andre-merzky:


marksantcroos:
Weren't you intending to deprecate this?

andre-merzky:
Yes - but Matteo still uses it.  That is how it popped up in the first place, on experiments in the aimes/swift context.

Issue #801, opened 2015-10-06
ValueError: doubled info entry asoc_adv_u


marksantcroos:
```
Traceback (most recent call last):
  File "inject.py", line 216, in <module>
    inject_all(session_ids, 'pickle')
  File "inject.py", line 147, in inject_all
    inject(sid)
  File "inject.py", line 105, in inject
    tr_cu_prof_fr = rpu.get_info_df(cu_prof_fr)
  File "/Users/mark/.virtualenv/rp/lib/python2.7/site-packages/radical/pilot/utils/analysis.py", line 522, in get_info_df
    raise ValueError('doubled info entry %s' % i)
ValueError: doubled info entry asoc_adv_u
```

Will try to dig deeper.

marksantcroos:
After adding the uid to the message too:
```
ValueError: doubled info entry asoc_adv_u (uid:unit.000441)
```

```
$ grep advance *|grep AgentStagingOutput | grep unit.000441
agent_0.AgentStagingOutputComponent.0.child.prof:43.9447,agent_0.AgentStagingOutputComponent.0.child:MainThread,unit.000441,AgentStagingOutput,advance,
agent_0.AgentStagingOutputComponent.0.child.prof:43.9892,agent_0.AgentStagingOutputComponent.0.child:MainThread,unit.000441,Failed,advance,
agent_4.AgentExecutingWatcher_POPEN.6.prof:22.1908,agent_4.AgentExecutingWatcher_POPEN.6:Watcher,unit.000441,AgentStagingOutputPending,advance,
```

So when a unit fails, it creates a message that causes a duplicate entry.

andre-merzky:
Its not actually a duplicated profile entry though, as it refers to different states...  I think the reason for this is that we defer the `FAILED` state until after we handled stdout/stderr, which we also want to get for a `FAILED` unit.  So we move the unit to output staging, do that output staging (so it goes into that state in the output stager), and then the output stager advances to `FAILED`, causing that second entry.

I guess the simplest and cleanest solution would be to differentiate the two DF entries based on state?

andre-merzky:
see https://github.com/radical-cybertools/radical.pilot/blob/devel/src/radical/pilot/agent/radical-pilot-agent-multicore.py#L3992

and https://github.com/radical-cybertools/radical.pilot/blob/devel/src/radical/pilot/agent/radical-pilot-agent-multicore.py#L5067

marksantcroos:
[...]

PR #886, opened 2015-11-01
Also create separate info entries for Canceled and Failed units.

marksantcroos:
This addresses #801.

### DIFF
--- a/src/radical/pilot/utils/analysis.py
+++ b/src/radical/pilot/utils/analysis.py
@@ -26,6 +26,11 @@ _info_pending = {
         'Pending'    : '_pend'
         }
 
+_info_premature_final = {
+        'Failed'   : '_fail',
+        'Canceled'   : '_canc'
+}
+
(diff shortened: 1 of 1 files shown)


marksantcroos:
Andre, feedback would be welcome whether this is the right approach or not.

andre-merzky:
Your code works and yields the expected result AFAICS -- but I am not sure you would need it.  Wouldn't selecting all columns with a specific info entry and FAILED or CANCELED state yield the same result?  Because, changing the info entry now makes it more cumbersome for search for that info entry for failed and non-failed units.  But I may be missing something...

marksantcroos:
> But I may be missing something...

If this refers to missing the original problem, then yes.

andre-merzky:
Yes, I did miss that of course :(

andre-merzky:
Yes, in this context this is the right thing to do I think.  Sorry again...

PR #888, opened 2015-11-01
catch up with state model changes

andre-merzky:
This catches up with some changes of unit states on the agent side which did not properly percolate to the RP module.  So here is this patch, instead of waiting on the module refactoring...


marksantcroos:
Thanks!

PR #889, opened 2015-11-01
avoid the repeated profiling of the AGENT_INPUT_STAGING_PENDING state

andre-merzky:


andre-merzky:
This should be complete now...

andre-merzky:
Too many open strands -- but the last commit should get things working.  Really! :P

PR #893, opened 2015-11-05
Only reschedule 1 CU after unschedule.

marksantcroos:
In steady state, there will only be one free slot, so going over all the CUs is
then fruitless.
This is now an enormous throughput improvement.

Scheduler needs attention in general, post-paper though.


andre-merzky:
nice catch!

Issue #65, opened 2014-02-06
Add lonestar.tacc.xsede.org to xsede.json


mturilli:
lonestar.tacc.xsede.org
- URL: LoadLeveler, LSF, and SGE 
- filesystem: sftp should work, otherwise globus-url-copy
- default_queue: normal (but also development, largemem, gnu, vis, serial)
- python_interpreter" : default: /usr/bin/python (2.4.3 :/); module load python: /opt/apps/python/epd/7.2.2/bin/python (2.7.2).
- pre_bootstrap: modules loaded by default:  1) TACC    2) TACC-paths    3) Linux    4) cluster    5) cluster-paths    6) intel/11.1    7) mvapich2/1.6    8) gzip/1.3.12    9) tar/1.22
- task_launch_mode: SSH??
- valid_roots: SCRATCH=/scratch/02855/mturilli; WORK=/work/02855/mturilli; home=/home1/02855/mturilli

PR #900, opened 2015-11-06
use RU id resettting

andre-merzky:


andre-merzky:
This PR addresses the problem that unit ID counters are not reset on new sessions.  Now they are.  This fix depends on PR #65 or radical.utils, which implements the actual reset logic.

PR #903, opened 2015-11-09
Only break out when we can't schedule.

marksantcroos:
This is a data supported incrmental change to the recent scheduler change.
In the previous incarnation we would gradually reach a lower number of
concurrent cores as we would never allocate more than one new, even if multiple
units had completed.

This is probably not a great change though in the light of mixed size units.


marksantcroos:
Dont pull me just yet.

marksantcroos:
Doing this in a new branch.

PR #904, opened 2015-11-11
Break out the debug parameters to ease debugging.

marksantcroos:
This should be a semantically void change.


andre-merzky:
Yep, looks void...

PR #912, opened 2015-12-09
Fix/lm cc signature

andre-merzky:
This is to be merged after feature/yarn.


andre-merzky:
This PR now reintroduces the shell quoting mechanism we had in place in the agent, and reconciles it with the changed  signature for construct_command.  I consider it ready for merge.

PR #916, opened 2015-12-09
Fix/866

andre-merzky:
Is this mergeable?


marksantcroos:
Now it is ready.

andre-merzky:
*nod* looks good enough to me,  thanks!

PR #919, opened 2015-12-09
Fix/resched3

andre-merzky:
I think we both agreed on this one, right?


marksantcroos:
Yes, although we don't complete understand the decline yet.
Opened a new ticket for that: #922.

Also, this may leave units unscheduled, also a ticket for that #921.

andre-merzky:
ack on both accounts.

PR #925, opened 2015-12-10
Fix/bw config

marksantcroos:
New python and openmpi config, tested by @vivek-bala and myself.


andre-merzky:
Mark said its safe to merge, so blame goes to him.  Not me.  Him!

PR #927, opened 2015-12-11
Feature/osgng

marksantcroos:
Putting this up for incremental review.


src/radical/pilot/agent/radical-pilot-agent-multicore.py
```diff
@@ -3242,7 +3247,8 @@ def stop(self):
         # During LRMS termination, we call any existing shutdown hooks on the
         # launch methods.  We only call LM shutdown hooks *once*
         launch_methods = set() # set keeps entries unique
-        launch_methods.add(self._cfg['mpi_launch_method'])
+        if 'mpi_launch_method' in self._cfg:
+            launch_methods.add(self._cfg['mpi_launch_method'])
         launch_methods.add(self._cfg['task_launch_method'])
```
andre-merzky: The task_lm would also need a guard?  Or we use `cfg.get()` and later ignore the `None` in the set?

src/radical/pilot/controller/pilot_launcher_worker.py
```diff
@@ -362,16 +362,17 @@ def run(self):
...
+                        runtime         = compute_pilot['description']['runtime']
+                        queue           = compute_pilot['description']['queue']
+                        project         = compute_pilot['description']['project']
+                        cleanup         = compute_pilot['description']['cleanup']
+                        resource_key    = compute_pilot['description']['resource']
+                        schema          = compute_pilot['description']['access_schema']
+                        memory          = compute_pilot['description']['memory']
+                        candidate_hosts = compute_pilot['description']['candidate_hosts']
+                        pilot_sandbox   = compute_pilot['sandbox']
+                        global_sandbox  = compute_pilot['global_sandbox']
```
andre-merzky: Thanks :)

src/radical/pilot/controller/pilot_launcher_worker.py
```diff
@@ -663,28 +667,32 @@ def run(self):
...
                         msg = "Copying agent configuration file '%s' to sandbox (%s)." % (cf_url, pilot_sandbox)
                         logentries.append(Logentry (msg, logger=logger.debug))
-                        sandbox_tgt.copy(cf_url, 'agent_0.cfg')
+                        if shared_filesystem:
+                            sandbox_tgt.copy(cf_url, agent_cfg_name)
 
-                        # close and remove temp file
+                        # Close agent config file
                         os.close(cfg_tmp_handle)
-                        os.unlink(cf_tmp_file)
```
andre-merzky: Why no unlink?

src/radical/pilot/controller/pilot_launcher_worker.py
```diff
@@ -718,8 +726,45 @@ def run(self):
...
+                                    #'%s > %s' % (rp_sdist_path, os.path.basename(rp_sdist_path)),
+                                    '%s > %s' % (rp_sdist_path, os.path.join(jd.working_directory, 'input', os.path.basename(rp_sdist_path))),
+                                    #'%s > %s' % (saga.sdist_path, os.path.basename(saga.sdist_path)),
+                                    '%s > %s' % (saga.sdist_path, os.path.join(jd.working_directory, 'input', os.path.basename(saga.sdist_path))),
+                                    #'%s > %s' % (ru.sdist_path, os.path.basename(ru.sdist_path)),
+                                    '%s > %s' % (ru.sdist_path, os.path.join(jd.working_directory, 'input', os.path.basename(ru.sdist_path)))
+                                ])
+
+                            if agent_config:
+                                jd.file_transfer.append('%s > %s' % (cfg_tmp_file, os.path.join(jd.working_directory, 'input', agent_cfg_name)))
```
andre-merzky: We always have a cfg file.  I think your assumption is that in the default agent_config we won't need it on the other hosts?  Will that always be true?  Why do we then copy the dists?  I am probably missing the point here wrt to unavailable shared FS...

andre-merzky:
This looks overall pretty good to me actually.  Do you consider this a good merge point, or do you think this will need more work before?

andre-merzky:
Mark? :)

mturilli:
I am testing devel already while collecting data for AIMES-Experience. If you are going to merge this into devel, I will add OSG to those tests. It might be a good opportunity.

src/radical/pilot/agent/radical-pilot-agent-multicore.py
marksantcroos: So far I've only made MPI tasks optional. If you want to do the same for regular tasks, that needs changes in more places.

andre-merzky: There was a change to `get` in line 4618, that is why I am asking.  But if that is not related, then happy to leave as is...

src/radical/pilot/controller/pilot_launcher_worker.py
marksantcroos: > We always have a cfg file.

Good point, changed that.

src/radical/pilot/agent/radical-pilot-agent-multicore.py
marksantcroos: It is related, but changing the case for regular tasks here wouldn't make any difference because of other places in the code (both client and agent). So I suggest to leave it indeed.

andre-merzky:
Thanks! :)

PR #941, opened 2016-01-10
Feature/split agent

andre-merzky:
Merging this will unstall agent development.  This will need some testing though, to avoid loss of functionality.


marksantcroos:
Thanks Andre, looks good!

PR #944, opened 2016-01-18
Fix/ve misc

marksantcroos:
This makes OSG work again.

### DIFF
--- a/src/radical/pilot/bootstrapper/bootstrap_1.sh
+++ b/src/radical/pilot/bootstrapper/bootstrap_1.sh
@@ -653,6 +653,12 @@ virtenv_activate()
         source activate $virtenv/
     else
         . "$virtenv/bin/activate"
+        if test -z "$VIRTUAL_ENV"
+        then
+            echo "Loading of virtual env failed!"
+            exit 1
+        fi
+
     fi
     VIRTENV_IS_ACTIVATED=TRUE
 
--- a/src/radical/pilot/configs/resource_osg.json
+++ b/src/radical/pilot/configs/resource_osg.json
@@ -31,8 +31,9 @@
       "module load python/2.7"
     ],
     "rp_version": "debug",
-    "virtenv": "%(global_sandbox)s/ve_xd-login",
-    "virtenv_mode": "create"
+    "virtenv": "./ve_xd-login",
+    "virtenv_mode": "create",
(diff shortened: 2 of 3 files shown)


andre-merzky:
FWIW, it is annoying that a failing bootstrap is not resulting in a `FAILED` pilot, but one basically only find out about the failure by looking at the remote logfiles...

andre-merzky:
Otherwise the dir permission problem seems  gone - YAY!  I do end up with: bootstrap_1.sh this though:

```
 line 83: cc: command not found
```

any advise?

marksantcroos:
> FWIW, it is annoying that a failing bootstrap is not resulting in a FAILED pilot, but one basically only find out about the failure by looking at the remote logfiles...

You mean because the jobs goes into `HOLD` or in general?

marksantcroos:
> line 83: cc: command not found
> any advise?

Yes, blacklist the site similar as https://github.com/radical-cybertools/radical.pilot/blob/devel/examples/getting_started_osg.py#L123

You need to get the `GLIDEIN_Site` entry from bootstrap_1.out.

andre-merzky:
> You mean because the jobs goes into HOLD or in general?

Yes, it goes into `HOLD` and never leaves (or so it seems).  But the bootstrapper clearly failed at this point.  I think we discussed this, you said Condor is waiting for output files which will never appear.  Should we 'touch' them in the bootstrapper?  Is there any other way to get things unstuck?  Leaving it stuck does not seem like an option, really...

andre-merzky:
> blacklist...

Ah: that line was disabled in the branch.  Lets see how that goes...  Thanks!

marksantcroos:
> Yes, it goes into HOLD and never leaves (or so it seems). But the bootstrapper clearly failed at this point. I think we discussed this, you said Condor is waiting for output files which will never appear. Should we 'touch' them in the bootstrapper? Is there any other way to get things unstuck? Leaving it stuck does not seem like an option, really...

Right, this falls in the 80% part of getting OSG to work :)

marksantcroos:
> Ah: that line was disabled in the branch. Lets see how that goes... Thanks!

I'll start to create separate tickets for these site specific issues.

andre-merzky:
Thanks!

andre-merzky:
This seems to work nicely now - thanks!

PR #959, opened 2016-01-29
scattered fixes toward the release...

andre-merzky:


src/radical/pilot/agent/staging_output/default.py
```diff
@@ -164,13 +164,10 @@ def work(self, cu):
...
                 self._log.info("Going to '%s' %s to %s", directive['action'], abs_source, target)
 
-                if directive['action'] == LINK:
-                    # This is probably not a brilliant idea, so at least give a warning
-                    os.symlink(abs_source, target)
-                elif directive['action'] == COPY:
-                    shutil.copyfile(abs_source, target)
-                elif directive['action'] == MOVE:
-                    shutil.move(abs_source, target)
+                # This is probably not a brilliant idea, so at least give a warning
```
marksantcroos: Can you revert this? The comment looses meaning for the sake of spacing.

src/radical/pilot/agent/staging_output/default.py
```diff
@@ -139,7 +139,7 @@ def work(self, cu):
                         directive, cu['_id'], workdir)
 
                 # Convert the target_url into a SAGA Url object
-                target_url = rs.Url(directive['target'])
+                target_url = ru.Url(directive['target'])
```
marksantcroos: I figured that we use SAGA urls and not RU urls, as per the comment.

src/radical/pilot/agent/staging_output/default.py
andre-merzky: will do

src/radical/pilot/agent/staging_output/default.py
andre-merzky: Let me fix the comment.  In other places that makes sense (for stupid reasons) -- here it does not, as the instance is not forwarded to a saga call.

PR #969, opened 2016-02-20
fix reordering of states for some update sequences

andre-merzky:
This seems to fix the state update issues we saw lately.

Under stress, the first two state updates can already happen out of order.   I missed that case, which led to the state history being updated, but the state update itself was ignored.


marksantcroos:
Thanks Andre, first run looks promising! Will give it some more airtime before I merge though.

andre-merzky:
Yes, sure, that's appreciated :)

marksantcroos:
Remains stable!

PR #971, opened 2016-02-25
Feature/yarn multinode

iparask:
I think it is ready. Waiting for comments

### DIFF
--- a/src/radical/pilot/agent/lm/yarn.py
+++ b/src/radical/pilot/agent/lm/yarn.py
@@ -103,8 +103,8 @@ class Yarn(LaunchMethod):
             lines = yarn_site_file.readlines()
             yarn_site_file.close()
 
-            free_mem_str=subprocess.check_output(['grep','MemTotal','/proc/meminfo'])
-            total_free_mem=int(free_mem_str.split()[1])/1048
+            total_mem_str=subprocess.check_output(['grep','MemTotal','/proc/meminfo'])
+            total_free_mem=int(total_mem_str.split()[1])/1048
 
             if nodelist.__len__() == 1:
                 cores_used = cores/2
@@ -113,6 +113,8 @@ class Yarn(LaunchMethod):
                 cores_used = cores*(nodelist.__len__()-1)
                 total_mem = total_free_mem*(nodelist.__len__()-1)
 
+            max_app_mem = total_mem/cores_used
+
             prop_str  = ' <property>\n'
             prop_str += '  <name>yarn.nodemanager.aux-services</name>\n'
             prop_str += '    <value>mapreduce_shuffle</value>\n'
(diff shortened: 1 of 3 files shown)


src/radical/pilot/agent/lm/yarn.py
```diff
@@ -160,7 +215,12 @@ def config_yarn_site():
             # Solution to find Java's home folder:
             # http://stackoverflow.com/questions/1117398/java-home-directory
 
-            jpos = subprocess.check_output(['readlink','-f', '/usr/bin/java']).split('bin')
+            java = subprocess.check_output(['which','java'])
```
andre-merzky: there is this:  https://github.com/radical-cybertools/radical.utils/blob/devel/src/radical/utils/which.py#L9

src/radical/pilot/agent/lm/yarn.py
```diff
@@ -160,7 +215,12 @@ def config_yarn_site():
...
+            if java != '/usr/bin/java':
+                jpos=java.split('bin')
+            else:
+                jpos = subprocess.check_output(['readlink','-f', '/usr/bin/java']).split('bin')
```
andre-merzky: https://docs.python.org/2/library/os.html#os.readlink

andre-merzky: I should motivate the comments: shell callouts can be expensive, and they are not as portable as one would think.

andre-merzky:
Apart from the comments made above, that PR looks good to me - thanks!

src/radical/pilot/agent/lm/yarn.py
iparask: I have to think a bit how to incorporate flag -f of the readlink. It makes the command to follow the symbolic link until it get to the actual executable. For example, if you do:

  readlink /usr/bin/java

on Ubuntu, it will return 

  /etc/alternatives/java

which is also a symlink.

I could do a while loop, but I am not sure it is going to be very nice.

andre-merzky:
```
merzky@cameo:~ $ cd /tmp/
merzky@cameo:/tmp $ touch a
merzky@cameo:/tmp $ ln -s a b
merzky@cameo:/tmp $ ln -s b c
merzky@cameo:/tmp $ ln -s c d
merzky@cameo:/tmp $ vi test.py
merzky@cameo:/tmp $ cat test.py 

import os

path = '/tmp/d'

while os.path.islink(path):
    path = os.readlink(path)

print path
merzky@cameo:/tmp $ python test.py 
a
```
We can make this a call in radical.utils...

:)

src/radical/pilot/agent/lm/yarn.py
```diff
@@ -160,7 +214,12 @@ def config_yarn_site():
             # Solution to find Java's home folder:
             # http://stackoverflow.com/questions/1117398/java-home-directory
 
-            jpos = subprocess.check_output(['readlink','-f', '/usr/bin/java']).split('bin')
+            java = ru.which('java')
+            if java != '/usr/bin/java':
+                jpos=java.split('bin')
+            else:
+                jpos = os.path.realpath('/usr/bin/java').split('bin')
+
```
andre-merzky: Wouldn't you always do the 'realpath' expansion?  Why only for `/usr/bin/`?

PR #975, opened 2016-03-05
make orte-submit aware of non-mpi CUs

andre-merzky:
it now allocates the same set of cores to the CU, but only creates one
application instance when cud['mpi'] == False


marksantcroos:
Ah good catch, I can see the context of this :) Sorry @antonst!

PR #979, opened 2016-03-08
Enable scheduling time measurements.

marksantcroos:
I dont think that without this it is possible/easy to measure the time it takes to schedule a cu.


marksantcroos:
Thanks!

PR #981, opened 2016-03-09
use high water mark for ZMQ to avoid message drops on high loads

andre-merzky:
the 'linger' timeout make sure the sockets don't stay around after shutdown.  that was the default in earlier versions of pyzmq, but isn't anymore...


marksantcroos:
You mean pyzmq right?

andre-merzky:
I think it is enacted on the ZMQ layer really.

marksantcroos:
But you referred to pymongo

andre-merzky:
Ah, sorry, ztupid me.  ZMQ it is.

PR #993, opened 2016-03-29
Disable vnode splitting.

marksantcroos:
Fix for changed ARCHER vnode format. Neither the original nor the new is perfect, but currently this is the only PBS Pro installation we have and its pretty non-parseable.


andre-merzky:
Thanks!

PR #1011, opened 2016-04-26
build gtod as part of rp install

andre-merzky:
we need ditutils to be able to compile things anyway, and this can potentially remove some magic from the bootstrapper.  But, it only makes gtod available once RP is installed, ie. basically when the bootstrapper is done.


andre-merzky:
obsolete...

PR #1012, opened 2016-04-26
Feature/bw ppn xk

andre-merzky:
This is supposed to provide support toward BW's GPU usage.  It might not be the right approach -- needs discussion.


andre-merzky:
see https://github.com/radical-cybertools/saga-python/pull/540

PR #1015, opened 2016-04-26
Feature/pilotdata

andre-merzky:


marksantcroos:
Why did you move it to docs?

andre-merzky:
Because I would hesitate to expose API which is not implemented.

marksantcroos:
Thats why it is one a branch :-)
Would you mind keeping it as such?

andre-merzky:
Hmm, I don't like branches lying around which do nothing, that is how we ended up with about 30 of them.  But otherwise, sure :)

PR #1021, opened 2016-04-27
Fix/anaconda cu 998

iparask:


iparask:
Hello,

Is there a plan when this fix will go through?

Thank you

andre-merzky:
Thanks for the reminder!

PR #1025, opened 2016-05-02
Remove temp agent config directory.

marksantcroos:
Change is probably void for the client refactoring branch, but still ...


andre-merzky:
thanks! :)

PR #1026, opened 2016-05-02
Tempfile is not a tempfile so requires explicit removal.

marksantcroos:
Incremental fix to the previous one.


marksantcroos:
Wait ... wtf ...

PR #1039, opened 2016-06-14
Feature/spark

iparask:
I think it is time for a code review and to see if this feature can become part of devel & released in the near future.


bin/radicalpilot-create-static-ve
```diff
@@ -1,6 +1,44 @@
 #!/bin/sh
 
+<<<<<<< HEAD
```
iparask: This conflict is being resolved by Andre.

andre-merzky:
Thanks Iannis, George!  I agree that a new review makes sense at this point.

src/radical/pilot/agent/rm/spark.py
```diff
@@ -0,0 +1,57 @@
...
+                        detected_cpus, selected_cpus)
+                selected_cpus = detected_cpus
+
+            elif detected_cpus > selected_cpus:
+                self._log.warn("more cores available: using requested %d instead of available %d.",
+                        selected_cpus, detected_cpus)
+
+
+   
+
```
andre-merzky: Can we remove some of the empty lines, please?

(I would never have thought I'd ever make this comment on code :P )

iparask: Yes, let me take a look

src/radical/pilot/agent/rm/spark.py
```diff
@@ -0,0 +1,57 @@
...
+        hostname = os.environ.get('HOSTNAME')
+
+        if hostname == None:
+            self.node_list = ['localhost']
+        else:
+            self.node_list = [hostname]
+        self.cores_per_node = selected_cpus
```
andre-merzky: This works better as:

```
if not hostname:
    hostname = 'localhost'

self.node_list = [hostname]
```

This catches an empty hostname, too.

src/radical/pilot/agent/scheduler/spark.py
```diff
@@ -0,0 +1,97 @@
...
+
+from .base import AgentSchedulingComponent
+
+
+#===============================================================================
+#
+class Spark(AgentSchedulingComponent):
+
+    # FIXME: clarify what can be overloaded by Scheduler classes
+
```
andre-merzky: I don't understand this comment.  Does that apply to this class implementation?

src/radical/pilot/agent/scheduler/spark.py
```diff
@@ -0,0 +1,97 @@
...
+        # Find out how many applications you can submit to YARN. And also keep
+        # this check happened to update it accordingly
+
+
+        #if 'rm_ip' not in self._cfg['lrms_info']:
+        #    raise RuntimeError('rm_ip not in lm_info for %s' \
+        #            % (self.name))
+
+        self._log.info('Checking rm_ip %s' % self._cfg['lrms_info']['lm_info']['rm_ip'])
+        self._rm_ip = self._cfg['lrms_info']['lm_info']['rm_ip']
```
andre-merzky: A general remark: instead of

```
self._log.info('something with a %s' % string)
```

please use

```
self._log.info('something with a %s' % string)
```

In the first case the string will get expanded when the log call is made, so always when that code section is executed.  In the second case, it only gets expanded by the logging module when the log level requires it.  As string expansion can be rather expensive in python, we want the latter specifically in frequently called sections -- but it is a good habit to always use the comma'd version.

src/radical/pilot/agent/scheduler/spark.py
```diff
@@ -0,0 +1,97 @@
...
+        sample_time = rpu.timestamp()
+
+        #-----------------------------------------------------------------------
+        # Find out the cluster's resources
+
+        self._mnum_of_cores = metrics['clusterMetrics']['totalVirtualCores']
+        self._mmem_size = metrics['clusterMetrics']['totalMB']
+        self._num_of_cores = metrics['clusterMetrics']['allocatedVirtualCores']
+        self._mem_size = metrics['clusterMetrics']['allocatedMB']
+
```
andre-merzky: What does the `m` prefix in `mnum` and `mmem` mean?  They are all metrics?

iparask: Maximum

src/radical/pilot/agent/scheduler/spark.py
```diff
@@ -0,0 +1,97 @@
...
+        #with self._slot_lock:
+        self._log.info('Releasing : {0} Cores, {1} RAM'.format(opaque_slot['task_slots'][0],opaque_slot['task_slots'][1]))
+        self.avail_cores +=opaque_slot['task_slots'][0]
+        self.avail_mem +=opaque_slot['task_slots'][1]
+        self.avail_app['apps']+=1
+        return True
+
+
+
+
```
andre-merzky: Too many empty lines :)

src/radical/pilot/agent/scheduler/spark.py
```diff
@@ -0,0 +1,97 @@
...
+        self.avail_mem = self._mmem_size - self._mem_size
+
+
+    # --------------------------------------------------------------------------
+    #
+    def _release_slot(self, opaque_slot):
+        #-----------------------------------------------------------------------
+        # One application has finished, increase the number of available slots.
+        #with self._slot_lock:
+        self._log.info('Releasing : {0} Cores, {1} RAM'.format(opaque_slot['task_slots'][0],opaque_slot['task_slots'][1]))
```
andre-merzky: the `.format` syntax is also always expanded -- please use the log syntax mentioned above.  Thanks!

src/radical/pilot/configs/resource_local.json
```diff
@@ -82,6 +82,63 @@
...
+        "local"                       :
+        {
+            "job_manager_endpoint"    : "fork://localhost/",
+            "filesystem_endpoint"     : "file://localhost/"
+        },
+        "default_remote_workdir"      : "$HOME",
+        "lrms"                        : "FORK",
+        "agent_type"                  : "multicore",
+        "agent_scheduler"             : "CONTINUOUS",
+        "agent_spawner"               : "POPEN",
```
andre-merzky: Would the scheduler not also be set to `SPARK`?

src/radical/pilot/agent/scheduler/spark.py
iparask: I think it was forgotten there.

See [here](https://github.com/radical-cybertools/radical.pilot/blob/devel/src/radical/pilot/agent/scheduler/base.py#L27)

src/radical/pilot/configs/resource_local.json
```diff
@@ -82,6 +82,63 @@
...
+            "job_manager_endpoint"    : "ssh://localhost/",
+            "filesystem_endpoint"     : "sftp://localhost/"
+        },
+        "local"                       :
+        {
+            "job_manager_endpoint"    : "fork://localhost/",
+            "filesystem_endpoint"     : "file://localhost/"
+        },
+        "default_remote_workdir"      : "$HOME",
+        "lrms"                        : "FORK",
```
andre-merzky: Would the `lrms` not also be set to `SPARK`?

src/radical/pilot/agent/lm/spark.py
```diff
@@ -0,0 +1,285 @@
...
+        import radical.utils as ru
+
+        logger.info('Hook called by SPARK LRMS with the name %s'%lrms.name)
+
+        # If the LRMS used is not SPARK the namenode url is going to be
+        # the first node in the list and the port is the default one, else 
+        # it is the one that the SPARK LRMS returns
+        spark_home = None   
+        if lrms.name == 'SPARKLRMS':
+            logger.info("Found SPARK ")
```
andre-merzky: I could not find where `SPARLRMS` is ever set in the source code.  Could you clarify this, please?

src/radical/pilot/agent/lm/spark.py
```diff
@@ -0,0 +1,285 @@
...
+            launch_command = ru.which('spark')
+
+        else:
+            # Here are the necessary commands to start the cluster.
+            if lrms.node_list[0] == 'localhost':
+                node_name = lrms.node_list[0]
+            else:
+                logger.info("Bootstrap SPARK on " + socket.gethostname())
+                node_name =  socket.gethostname()
+
```
andre-merzky: I am not sure I aunderstant this switch: why not always use `gethostname` then, because that should also work for localhost?  But also, I don't see where `node_name` is used, really?

src/radical/pilot/agent/lm/spark.py
```diff
@@ -0,0 +1,285 @@
...
+                logger.info("Bootstrap SPARK on " + socket.gethostname())
+                node_name =  socket.gethostname()
+            
+            VERSION = "1.5.2"
+            SPARK_DOWNLOAD_URL= "http://d3kbcqa49mib13.cloudfront.net/spark-1.5.2-bin-hadoop2.6.tgz" #prebuilt
+            #Download the tar file
+            opener = urllib.FancyURLopener({})
+            download_destination = os.path.join(os.getcwd(),"spark-" + VERSION + ".tar.gz")
+            logger.info("Download: %s to %s"%(SPARK_DOWNLOAD_URL, download_destination))
+            opener.retrieve(SPARK_DOWNLOAD_URL, download_destination)
```
andre-merzky: Please use `debug` log level for information which are by default not interesting for users, but are only interesting for developers when, well, debugging :)

src/radical/pilot/agent/lm/spark.py
```diff
@@ -0,0 +1,285 @@
...
+            spark_tar = "spark-" + VERSION + ".tar.gz"
+            if not os.path.isfile(spark_tar):
+                raise RuntimeError("Spark wasn't downloaded properly. Please try again")
+
+            os.system("tar -xzf" + spark_tar + "; rm " + spark_tar ) #untar and delete tarball 
+            os.system("mv spark-1.5.2-bin-hadoop2.6 spark-1.5.2")
```
andre-merzky: Is there a reason to not use the original tarball name if you end up moving the dir anyway? ;)

src/radical/pilot/agent/lm/spark.py
```diff
@@ -0,0 +1,285 @@
...
+            spark_home = os.getcwd() + '/spark-' + VERSION
```
andre-merzky: The above lines mix strings which have `1.5.2` hard coded with strings which use `VERSION`.  You should decide for either option, otherwise `VERSION` is useless, really ;)

src/radical/pilot/agent/lm/spark.py
```diff
@@ -0,0 +1,285 @@
...
+
+            #-------------------------------------------------------------------
+            platform_os = sys.platform
```
andre-merzky: Introducing this variable is not useful.

src/radical/pilot/agent/lm/spark.py
```diff
@@ -0,0 +1,285 @@
...
+            java_home = os.environ.get('JAVA_HOME')
+
+            if platform_os == "linux" or platform_os == "linux2":
+                if not java_home:
+                    java = ru.which('java')
+                    if java != '/usr/bin/java':
+                        jpos=java.split('bin')
```
andre-merzky: How do you know there is a `bin` in the resulting path?

src/radical/pilot/agent/lm/spark.py
```diff
@@ -0,0 +1,285 @@
...
+                    else:
+                        jpos = os.path.realpath('/usr/bin/java').split('bin')
```
andre-merzky: same here actually...

src/radical/pilot/configs/resource_local.json
iparask: It assumes that the Spark cluster does not exist. I will these to George.

src/radical/pilot/agent/lm/spark.py
```diff
@@ -0,0 +1,285 @@
...
+
+            # if no installation found install scala 2.10.4
+            scala_home=ru.which('scala')
+            if not scala_home:
+                os.system('cd')
+                os.system('wget http://www.scala-lang.org/files/archive/scala-2.10.4.tgz')
+                if not os.path.isfile('scala-2.10.4.tgz'):
+                    raise RuntimeError("Scala wasn't downloaded properly. Please try again")
+                os.system('tar -xvf scala-2.10.4.tgz ; cd scala-2.10.4 ; export PATH=`pwd`/bin:$PATH; export SCALA_HOME=`pwd`')
+                os.system('rm scala-2.10.4.tgz')
```
andre-merzky: Do the exports have any effect beyond the shell used for the `os.system` call?

andre-merzky:
``` python
>>> import os
>>> os.system('export FOO=bar')
0
>>> os.system('echo $FOO')

0
```

src/radical/pilot/agent/lm/spark.py
```diff
@@ -0,0 +1,285 @@
...
+                scala_home = os.getcwd() + '/scala-2.10.4'
+                os.system('cd')
+
+
+            if lrms.node_list[0]!='localhost':
+                hostname = subprocess.check_output('/bin/hostname').split(lrms.node_list[0])[1].split('\n')[0]
+
```
andre-merzky: I am not sure I am parsing this one correctly.  So, if the first lrms node is not `localhost`, then you run `/bin/hostname`, and you split the result by the 2nd character of the first nodename?  And the result of that you split by `\n` and then take the first line or something?  What the heck do you have in that first node name?? ;)

src/radical/pilot/agent/lm/spark.py
```diff
@@ -0,0 +1,285 @@
...
+            spark_env_file.write('export PYSPARK_PYTHON='+python+'\n')
+
+            spark_env_file.close()
+
+
+            #### Start spark Cluster
+            spark_start = subprocess.check_output(spark_home + '/sbin/start-all.sh')
+            if 'Error' in spark_start:
+                raise RuntimeError("Spark Cluster failed to start: %s" % spark_start)
+            else:
```
andre-merzky: No else is needed after a raise...

src/radical/pilot/agent/lm/spark.py
```diff
@@ -0,0 +1,285 @@
...
+                logger.info('Start Spark Cluster')
+            launch_command = spark_home +'/bin'
```
andre-merzky: Is this really a command, or a path?

src/radical/pilot/agent/lm/spark.py
```diff
@@ -0,0 +1,285 @@
...
+
+        return lm_info
+
+
+    # --------------------------------------------------------------------------
+    #
+    @classmethod
+    def lrms_shutdown_hook(cls, name, cfg, lrms, lm_info, logger):
+        if 'name' not in lm_info:
+            raise RuntimeError('rm_ip not in lm_info for %s' \
```
andre-merzky: The error message complains about `rm_ip`, the check is for `name` -- which is it?

src/radical/pilot/agent/lm/spark.py
```diff
@@ -0,0 +1,285 @@
...
+                    % (self.name))
+
+        if lm_info['name'] != 'SPARKLRMS':
+            logger.info('Stoping SPARK')
```
andre-merzky: Another use of `SPARKLRMS` which is not set anywhere, AFAICS?

src/radical/pilot/agent/lm/spark.py
```diff
@@ -0,0 +1,285 @@
...
+            stop_spark = subprocess.check_output(lm_info['spark_home'] + '/sbin/stop-all.sh') 
+            if 'Error' in stop_spark:
+                logger.info("Spark didn't terminate properly")
+            else:
```
andre-merzky: this should be log level `error`, or at least `warning`

src/radical/pilot/agent/lm/spark.py
```diff
@@ -0,0 +1,285 @@
...
+        if task_args:
+            command = " ".join(task_args)
+        else:
+            command = " "
+
+        spark_configurations = " "
+        # if the user hasn't specified another ui port use this one
+        if not 'spark.ui.port' in command:
+            spark_configurations += ' --conf spark.ui.port=%d '  % (random.randint(4020,4180))  # can i use this range? TODO
+
```
andre-merzky: Why the random?

andre-merzky:
I don't see any conceptual problems with the PR, beyond the comments made on the diffs. Sorry for the multitude of comments -- but I think most are minor, really.  Thanks!

src/radical/pilot/agent/scheduler/spark.py
andre-merzky: Ah!  You might want to move `max` then or something.  But wither way, got it, thanks.

src/radical/pilot/agent/scheduler/spark.py
andre-merzky: Right, should be removed then, thanks

src/radical/pilot/configs/resource_local.json
andre-merzky: Ah, I see.  Are both options (existing and non-existing cluster) functional at this point?  Should there be config entries for both options?

src/radical/pilot/agent/lm/spark.py
iparask: The whole URL of the first master node is needed. e.g <nodename>.stampede.tacc.utexas.edu

src/radical/pilot/agent/lm/spark.py
iparask: It is a path which with the executable of a CU becomes the actual command. Spark offers a series of commands that allow a user to execute over a Spark Cluster.

src/radical/pilot/configs/resource_local.json
iparask: I think existing is not.

@georgeha ?

src/radical/pilot/agent/lm/spark.py
andre-merzky: I see - but then its not a `launch_command`, but a `launch_path`?  Sorry for being picky, but it would be good if we can still understand the code a couple of months from now... :)

src/radical/pilot/agent/lm/spark.py
andre-merzky: Ah, so its not a `hostname` you are getting here, but a domain name?  You may want to rename that then.  Ar you sure that a node name is never fully qualified?  Because what you do there will break then.  Also, are you sure that the output of `/bin/hostname` *is* fully qualified?  Because it often isn't ;)

```
$ ssh archer hostname
$ ssh archer hostname
eslogin003
```

Lastly, the domain returned on the headnode might not be the domain used by the compute nodes, eg. if multiple network cards are involved, which is often the case on clusters.  Guessing host names like this is tricky...

Would spark be able to handle IP addresses?  If so, you may want to look up the addresses for the nodes and use those.  That still can go wrong depending on network configurations, but does not suffer from the domain ambiguity and parsing problems....

src/radical/pilot/compute_pilot.py
```diff
@@ -301,7 +301,7 @@ def resource_detail(self):
         resource_details = {
             'nodes':          pilot_json['nodes'],
             'cores_per_node': pilot_json['cores_per_node'],
-            'lm_detail': pilot_json['lm_detail']
+            'lm_detail': pilot_json['lm_detail'] if pilot_json.has_key('lm_detail') else None
         }
```
andre-merzky:
```
'lm_detail': pilot_json.get('lm_detail')
```
;)

src/radical/pilot/agent/scheduler/spark.py
```diff
@@ -0,0 +1,94 @@
...
+        self.avail_mem = self._mmem_size - self._mem_size
+
+
+    # --------------------------------------------------------------------------
+    #
+    def _release_slot(self, opaque_slot):
+        #-----------------------------------------------------------------------
+        # One application has finished, increase the number of available slots.
+        #with self._slot_lock:
+        self._log.info('Releasing : %s Cores, %s RAM',[opaque_slot['task_slots'][0],opaque_slot['task_slots'][1])
```
marksantcroos: The bracket in front of the first opaque_slot needs to go.

iparask: oops...

src/radical/pilot/agent/lm/spark.py
```diff
@@ -0,0 +1,279 @@
...
+    # --------------------------------------------------------------------------
+    #
+    @classmethod
+    def lrms_config_hook(cls, name, cfg, lrms, logger):
+        """
+        FIXME: this config hook will inspect the LRMS nodelist and, if needed,
+               will start the SPARK cluster on node[0].
+        """
+        import radical.utils as ru
```
andre-merzky: Why is that a `FIXME`?  This seems implemented?

andre-merzky:
What are the plans for this branch?  This is now idle for a while, and has a couple of open comments...

iparask:
I will close it and open it again when George comes over

PR #1063, opened 2016-07-05
yet another attempt on fixing the interface detection fun

andre-merzky:


andre-merzky:
Uh, sorry for the accidental merge.  I'll ask for a  review separately then...

marksantcroos:
Which case does this fix?
(And more importantly, which ones does it break? :-)

andre-merzky:
The previous code did no check if the picked interface is actually connected, and the code bailed out if it preferred a disconnected interface over a connected one.  Thus also the restructuring of the code: that avoids to have the same set of checks in all three loops, each of which picked a candidate from a different list.  Now we  first unify the list, then have one loop for pick-and-check.

The semantically crucial part is if that candidate list is assemblied correctly.

marksantcroos:
Sounds good, thanks.

PR #1090, opened 2016-07-27
attempt to fix #1071

### LINKED ISSUE #1071: MongoDB on RADICAL vm is not usable for experiments on XSEDE and BlueWaters
Less than 1/3 of CUs reach state 'Done'. This is observed consistently and as a result not even a single simulation cycle can finish. Switching to mlab's MongoDB seems to solve the issue, but mlab is not suitable for experiments: http://docs.mlab.com/production-guide/#predictable-performance

andre-merzky:


marksantcroos:
Thanks Andre!

marksantcroos:
Note that this section could use some more documentation, as it would have been difficult for me to spot this.

andre-merzky:
Yes, fair.  It was difficult for me to spot this, too :)  But that code was expected to have a very limited shelf-life, and as such I won't add more docs for now, apologies!

The split-module branch gets rid of the problem, as stat updates are now truly pushed in any order.  The state-model interpretation is now always done on the receiving end, and the code (and documentation) now lives here:

https://github.com/radical-cybertools/radical.pilot/blob/feature/split_module_2/src/radical/pilot/states.py#L139

I would not mind your pair of eyes on that code, obviously :)  It remains a crucial part wrt. to state consistency...

Thanks, Andre.

marksantcroos:
> But that code was expected to have a very limited shelf-life, and as such I won't add more docs for now, apologies!

In that case don't worry.

Issue #1107, opened 2016-08-20
Wrangler example configuration


iparask:
Hello,

Wrangler's entry in the [config.json](https://github.com/radical-cybertools/radical.pilot/blob/master/examples/config.json#L80) is not complete. Should I use ssh or gsissh? I will fill it for my tests.

andre-merzky:
If the setting is `null`, then RP should use the default setting from the resource config, i.e. the first entry in  https://github.com/radical-cybertools/radical.pilot/blob/master/src/radical/pilot/configs/resource_xsede.json#L96, which would be ssh.  Is that failing for you?

iparask:
It failed because it was missing the cores, not the schema. I will update it for the tests. Should I commit the config.json?

andre-merzky:
Yes, please.

PR #1108, opened 2016-08-23
Resolving #1107

iparask:


marksantcroos:
Is there a special reason your PR is against master? (as the default is devel)
If not, please do a PR against devel.
(I tried to change it, but then the base is wrong)

andre-merzky:
Yes, I asked for that, as this is towards the current round of testing, and I am confident that this is without side effects.  But thanks for being careful here!

marksantcroos:
Ok!

Issue #1106, opened 2016-08-19
FR: Create flag for RP directives to skip files not found on source


vivek-bala:
Currently, if a file is not found in the source leads to an error. It would be good to have a flag which allows it to not prompt an error (and terminate) but gracefully skip (with a warning).

marksantcroos:
That should be relatively easy I think. We can add a `NON_FATAL` flag to the staging directive and let the code continue if the transfer fails of a non-fatal staging directive.

andre-merzky:
maybe rather use `skip_missing`?  Then we can (a) optimize, and (b) still raise on other errors (timeout, connection break etc) which the user would have otherwise to distinguish herself.

marksantcroos:
There is something to say for your suggestion, although in reality its not always possible to distinguish a missing file from a connection error.

Can you describe the use case so that we can make an informed decision?

andre-merzky:
The pilot is very well able to check the existence of files, and there is little overhead:

  * the pilot interprets some staging_output directives anyway, so there is little overhead in checking the respective ones and confirm file existence
  * the pilot also pushes a full unit update to UMGR after staging -- file existence information can well be part of that push, with little overhead (some bools).

In fact, that would also be an optimization in the normal case, as we don't even need to start remote checks or transfers for non-existing files, and can fail units earlier, before attempting to stage other files.  Assume 10 output files, the last one is missing: the agent can fail that unit immediately after the existence check (if `skip_missing == False`).

marksantcroos:
So you only want this for missing output? You didn't specify that earlier.

andre-merzky:
From the title: `... to skip files not found on source` ;)

marksantcroos:
I saw that, thats why I was surprised by your agent focus.

vivek-bala:
Example usecase: First 100 simulation tasks start executing. After time T, 1 analysis task is submitted for execution. The analysis task attempts to stage in a checkpoint file from all these 100 simulation tasks (intermediate data, so the simulation tasks are not yet DONE). The analysis task will operate on as many as checkpoint files as it can (so it doesn't HAVE to wait for all 100). So if only 40 simulation tasks have reached the state which produces the checkpoint file, analysis will stage in these 40 checkpoint files and operate on them (although it would "poll" all simulation tasks for the checkpoint file).

[...]

PR #1109, opened 2016-08-23
Feature/non fatal missing

marksantcroos:
Fix for #1106, verified by Vivek.

### DIFF
--- a/src/radical/pilot/agent/staging_input/default.py
+++ b/src/radical/pilot/agent/staging_input/default.py
@@ -135,6 +135,19 @@ class Default(AgentStagingInputComponent):
 
                 self._log.info("Going to '%s' %s to %s", directive['action'], source, abs_target)
 
+                # For local files, check for existence first
+                if directive['action'] in [rpc.LINK, rpc.COPY, rpc.MOVE]:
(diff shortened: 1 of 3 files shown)


andre-merzky:
So that only covers input, and only local ops (not transfer).   Shouldn't we do those all in one go?  Happy to take over if you consider that out of scope wrt. your time allotment!

PS.: and I still think `SkipMissing` fits better, as the above code will still be (and should be) fatal on failing ops... ;)

marksantcroos:
I think we have enough real use cases to not solve problems for non-existing use cases. But don't let me stop you.

andre-merzky:
Yeah, I agree in principle, but don't feel very comfortable to introduce a flag which applies sometimes...  Well, I guess such are our different roles now :)

marksantcroos:
Thats easily solved by renaming the flag to `SKIP_MISSING_FILES_ON_AGENT`.
Note that in that case I'm ok with using the word `missing` in it. Anything that is not local I don't think we should try to distinguish missing and failing.

andre-merzky:
Merged now, but I'll keep the ticket open to address the other cases.  Thanks Mark!

PR #1123, opened 2016-10-03
when calling a unit state callback, have missed states also trigger CBs

andre-merzky:
This is a hack.

### DIFF
--- a/examples/error_handling.py
+++ b/examples/error_handling.py
@@ -115,7 +115,7 @@ if __name__ == "__main__":
 
         # we submit one compute unit which will just fail
         cud = rp.ComputeUnitDescription()
-        cud.executable = '/bin/date'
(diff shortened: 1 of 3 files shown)


examples/error_handling.py
```diff
@@ -115,13 +115,17 @@ def unit_state_cb (unit, state):
 
         # we submit one compute unit which will just fail
         cud = rp.ComputeUnitDescription()
-        cud.executable = '/bin/fail'
+        cud.executable = '/bin/date'
```
marksantcroos: This is probably not intentional?

andre-merzky: Ah, thanks for the catch, will fix.

marksantcroos:
Thanks!

PR #1125, opened 2016-10-05
add cu tmp dir information to resource config kitchen sink

andre-merzky:
Iannis, please have a look at this branch and give it a try.  Yarn should find a `cu_tmp` key in the agent config and can then use the location for tmp storage.


andre-merzky:
ping @iparask for testing

PR #1134, opened 2016-10-10
Fix/issue 1043 am

andre-merzky:


marksantcroos:
Can you add some documentation for this, can't see what it is supposed to do.

andre-merzky:
Let me quickly reply here - but I'll add some comments to the code, too.

In the `bootstrapper`, we now store a copy of the *original* env as we find it when we land on the node:

```
env | sort | grep '=' | tee env.orig
```

When creating a CU script, we now  can transplant parts of that virgin env into the CU env.  What parts we transplant is defined in the resource config, like on stampede:

```
       "export_to_cu"  : ["LMOD_CMD",
￼                          "LMOD_SYSTEM_DEFAULT_MODULES",
￼                          "LD_LIBRARY_PATH"], 
￼       "cu_pre_exec"   : ["module restore"]
```

with the intent to give us a similarly 'usable' env as we expect from any other batch job.

The rest of the diff really is just the mechanics to communicate the config settings, and to inject the env into the CU script.

PR #1136, opened 2016-10-13
update resource file, fix system call

georgeha:
- Changed one system call from check_call to check_output
- updated in resource file python version
- removed ui port random selection

### DIFF
--- a/src/radical/pilot/agent/lm/spark.py
+++ b/src/radical/pilot/agent/lm/spark.py
@@ -74,9 +74,8 @@ class Spark(LaunchMethod):
         scala_home=ru.which('scala')
         if not scala_home:
             try:
-                #subprocess.check_call('cd')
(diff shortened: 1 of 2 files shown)


src/radical/pilot/agent/lm/spark.py
```diff
@@ -109,7 +108,7 @@ def lrms_config_hook(cls, name, cfg, lrms, logger):
         	master_ip = lrms.node_list[0]
         else:
         	try:
-        		master_ip = subprocess.check_call('hostname -f')
+        		master_ip = subprocess.check_output('hostname -f'.split()).strip()
```
iparask: There is no need to split 'hostname -f'. ```check_output``` accepts a list of inputs. I propose it to be ```['hostname','-f']```

PR #1155, opened 2016-11-10
Feature/cherries

marksantcroos:
Sweet cherries longing to get merged into devel.

### DIFF
--- a/src/radical/pilot/agent/radical-pilot-agent-multicore.py
+++ b/src/radical/pilot/agent/radical-pilot-agent-multicore.py
@@ -580,6 +580,12 @@ def bootstrap_3():
                            {"$set": {"lm_info"  : lrms.lm_info.get('version_info'),
                                      "lm_detail": lrms.lm_info.get('lm_detail')}})
 
+            # OSG Specific
+            osg_resource_name = os.environ.get('GLIDEIN_ResourceName')
+            if osg_resource_name:
+                mongo_p.update({"_id": pilot_id},
+                               {"$set": {"osg_resource_name": osg_resource_name}})
+
         # we now have correct bridge addresses added to the agent_0.cfg, and all
         # other agents will have picked that up from their config files -- we
         # can start the agent and all its components!
--- a/src/radical/pilot/utils/analysis.py
+++ b/src/radical/pilot/utils/analysis.py
@@ -59,6 +59,7 @@ _info_states = [
         ]
 
 _info_entries = [
+    # info,              name,                     event,       msg
     ('umgr_get_u',      'MainThread',             'advance',   'New'),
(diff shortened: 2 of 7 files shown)


src/radical/pilot/agent/radical-pilot-agent-multicore.py
```diff
@@ -580,6 +580,12 @@ def sigalarm_handler(signum, frame):
                            {"$set": {"lm_info"  : lrms.lm_info.get('version_info'),
                                      "lm_detail": lrms.lm_info.get('lm_detail')}})
 
+            # OSG Specific
+            osg_resource_name = os.environ.get('GLIDEIN_ResourceName')
+            if osg_resource_name:
+                mongo_p.update({"_id": pilot_id},
+                               {"$set": {"osg_resource_name": osg_resource_name}})
```
andre-merzky: This should probably go into the `resource_detail` attribute of the pilot?

src/radical/pilot/utils/analysis.py
```diff
@@ -569,6 +570,9 @@ def get_info_df(df):
         for i in set(l):
             cols.add(i)
             if l.count(i)>1:
+                # TODO: this is a hack to workaround a problem in the state management
+                if i == 'mod_adv_u':
```
andre-merzky: which one? :)

src/radical/pilot/agent/radical-pilot-agent-multicore.py
marksantcroos: What is `resource_detail`?

src/radical/pilot/utils/analysis.py
marksantcroos: i55 pimgpsh_fullsize_distr

src/radical/pilot/agent/radical-pilot-agent-multicore.py
andre-merzky: See https://github.com/radical-cybertools/radical.pilot/blob/devel/src/radical/pilot/compute_pilot.py#L290 - is basically exposes backend specific pilot information for user consumption.  The Spark LM for example exposes the spark service URL in `pilot.resource_detail['lm_detail'].

But no  worries - since you put the information into the DB, the exposure via API etc. is orthogonal.

src/radical/pilot/utils/analysis.py
andre-merzky: Thanks, I see.  I started to look into a fix to avoid the repetitions - but nevertheless, the fix you have here makes sense then.

marksantcroos: I noticed. Although even if it gets fixed, we probably want to keep it around for a bit for existing profiles.

PR #1156, opened 2016-11-10
Get going on as many OSG sites possible.

marksantcroos:
Preserve for future generations to come ...


marksantcroos:
Might need some cleanup as it has some pre-merge-from-devel semantics.

andre-merzky:
Thanks

PR #1157, opened 2016-11-10
Trigger only once.

marksantcroos:
Not sure if this should be kept, or where it even comes from ...


andre-merzky:
Thanks - I'll test this later, but that can likely be merged as is...

ibethune:
Is this one of the outstanding fix branches that we want to merge for the 0.45 release?

andre-merzky:
Yeah, we should probably merge it, there is no point in keeping it around.  Will do.

PR #1158, opened 2016-11-10
Agent based staging directives.

marksantcroos:
This are the agent staging directives in isolation. Still a bit SRM centric, but probably good to go.

### DIFF
--- a/src/radical/pilot/agent/staging_input/default.py
+++ b/src/radical/pilot/agent/staging_input/default.py
@@ -13,6 +13,10 @@ from ...  import utils     as rpu
 from ...  import states    as rps
 from ...  import constants as rpc
 
+os.environ['RADICAL_SAGA_VERBOSE'] = 'DEBUG'
+os.environ['RADICAL_SAGA_LOG_TGT'] = 'saga.log'
+import saga
+
 from .base import AgentStagingInputComponent
 
 
@@ -105,8 +109,7 @@ class Default(AgentStagingInputComponent):
         try:
             for directive in cu['Agent_Input_Directives']:
 
-                self._prof.prof('Agent input_staging queue', uid=cu['_id'],
-                         msg="%s -> %s" % (str(directive['source']), str(directive['target'])))
+                self._prof.prof('begin', uid=cu['_id'], msg=str(directive['_id']))
 
                 # Perform input staging
                 self._log.info("unit input staging directives %s for cu: %s to %s",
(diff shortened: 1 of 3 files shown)


src/radical/pilot/agent/staging_input/default.py
```diff
@@ -13,6 +13,10 @@
 from ...  import states    as rps
 from ...  import constants as rpc
 
+os.environ['RADICAL_SAGA_VERBOSE'] = 'DEBUG'
+os.environ['RADICAL_SAGA_LOG_TGT'] = 'saga.log'
+import saga
+
```
andre-merzky: This can probably go?

src/radical/pilot/agent/staging_input/default.py
```diff
@@ -151,17 +157,26 @@ def work(self, cu):
                 if   directive['action'] == rpc.LINK: os.symlink     (source, abs_target)
                 elif directive['action'] == rpc.COPY: shutil.copyfile(source, abs_target)
                 elif directive['action'] == rpc.MOVE: shutil.move    (source, abs_target)
-                else:
-                    # FIXME: implement TRANSFER mode
-                    raise NotImplementedError('Action %s not supported' % directive['action'])
+                elif directive['action'] == rpc.TRANSFER:
+                    srm_dir = saga.filesystem.Directory('srm://proxy/?SFN=bogus')
+                    if srm_dir.exists(source):
```
andre-merzky: Hmm, I am not sure I understand that part.  I can see how that works for SRM - but it immediately excludes other URLs which are otherwise valid, and could be transferred via SAGA.  Wouldn't it make sense to here switch for the schema?  (and additionally, to get really fancy, and to include a resource config option for an SRM proxy?)

src/radical/pilot/agent/staging_output/default.py
```diff
@@ -143,37 +144,50 @@ def work(self, cu):
...
+                    rpu.rec_makedir(os.path.dirname(target))
                     shutil.move(abs_source, target)
-                else:
-                    # FIXME: implement TRANSFER mode
-                    raise NotImplementedError('Action %s not supported' % directive['action'])
+                elif directive['action'] == rpc.TRANSFER:
+                    src_url = saga.Url(abs_source)
+                    src_url.schema = 'file'
+                    srm_dir = saga.filesystem.Directory('srm://proxy/?SFN=bogus')
+                    srm_dir.copy(src_url, target)
```
andre-merzky: Same.  Also I would think the copy needs a `CREATE_PARENT` flag?  The unit sandbox might not yet exist...

src/radical/pilot/agent/staging_input/default.py
marksantcroos: Assuming you don't mean the import, yes I think so.

andre-merzky: yeah, the import is kinda fine :)

src/radical/pilot/agent/staging_output/default.py
andre-merzky: Yeah, please do so.  In devel, the workdir does indeed exist, but that assumption is not valid in the split branch anymore...

andre-merzky: Ah, sorry, my bad - that is all right then...

andre-merzky:
Thanks Mark!

PR #1172, opened 2016-12-23
add rsh launch method

andre-merzky:
This is needed for the Allegro cluster at FU Berlin.

### DIFF
--- a/src/radical/pilot/agent/lm/yarn.py
+++ b/src/radical/pilot/agent/lm/yarn.py
@@ -6,6 +6,8 @@ __license__   = "MIT"
 import os
 import subprocess
 
+import radical.utils as ru
+
 from .base import LaunchMethod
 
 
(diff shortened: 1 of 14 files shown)


marksantcroos:
Can you remove _which() then?
And have _find_executable() also use ru.which?

andre-merzky:
That makes sense of course, will do.

andre-merzky:
Mark, if you have the time, please look over this PR again.  Thanks!

src/radical/pilot/agent/lm/yarn.py
```diff
@@ -6,6 +6,8 @@
 import os
 import subprocess
 
+import radical.utils as ru
```
marksantcroos: There is a duplicate import of this now.

andre-merzky: thanks, will fix

andre-merzky:
Thanks Mark!

Issue #1166, opened 2016-11-29
Failure when more than one pilot is used on TITAN


AA919:
Very strange phenomenon. It happens only when I use more than one pilot.
I have already tried to delete ve_titan.
This is the error that I found in the bootstraps:
```
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/lustre/atlas2/csc108/scratch/aleang9/radical.pilot.sandbox/ve_titan/rp_install/lib/python2.7/site-packages/radical/pilot/__init__.py", line 18, in <module>
  File "/lustre/atlas2/csc108/scratch/aleang9/radical.pilot.sandbox/ve_titan/rp_install/lib/python2.7/site-packages/radical/pilot/session.py", line 23, in <module>
    from .unit_manager    import UnitManager
  File "/lustre/atlas2/csc108/scratch/aleang9/radical.pilot.sandbox/ve_titan/rp_install/lib/python2.7/site-packages/radical/pilot/unit_manager.py", line 26, in <module>
    from .controller   import UnitManagerController
  File "/lustre/atlas2/csc108/scratch/aleang9/radical.pilot.sandbox/ve_titan/rp_install/lib/python2.7/site-packages/radical/pilot/controller/__init__.py", line 10, in <module>
    from pilot_manager_controller import PilotManagerController
  File "/lustre/atlas2/csc108/scratch/aleang9/radical.pilot.sandbox/ve_titan/rp_install/lib/python2.7/site-packages/radical/pilot/controller/pilot_manager_controller.py", line 27, in <module>
    from .pilot_launcher_worker import PilotLauncherWorker
  File "/lustre/atlas2/csc108/scratch/aleang9/radical.pilot.sandbox/ve_titan/rp_install/lib/python2.7/site-packages/radical/pilot/controller/pilot_launcher_worker.py", line 27, in <module>
    _, _, _, rp_sdist_name, rp_sdist_path = ru.get_version([root, pwd])
  File "/lustre/atlas2/csc108/scratch/aleang9/radical.pilot.sandbox/ve_titan/rp_install/lib/python2.7/site-packages/radical/utils/get_version.py", line 85, in get_version
    raise RuntimeError ("Cannot determine version from %s" % paths)
[...]

PR #1175, opened 2017-01-03
deprecate "debug" flag for rp_install

andre-merzky:
As the title says, this deprecates the `debug` mode for the pilot side radical stack installation in favor of the ``local` mode, which now is confined to the pilot sandbox.  This is a cleaner fix than #1166 to avoid race conditions on `pip install` during pilot startup.


andre-merzky:
Note that merge is not urgent, as a simple workaround for #1166 exists, by setting `rp_version = "debug"`.

ibethune:
Issue #1166 implies that the workaround above was implemented, so can this PR be closed?

andre-merzky:
It should be merged I think.  Despite the large number of changed lines, its actually a rather simple fix, as it *removes* some code complexity, and from a fairly rarely used code path at that.  I propose that Iself-review that PR once more and merge toward the release, if that's ok?

ibethune:
OK, let's go with that.

src/radical/pilot/bootstrapper/bootstrap_1.sh
```diff
@@ -429,8 +429,8 @@ virtenv_setup()
     virtenv_mode="$3"
     python_dist="$4"
 
-    virtenv_create=UNDEFINED
-    virtenv_update=UNDEFINED
+    ve_create=UNDEFINED
+    ve_update=UNDEFINED
```
andre-merzky: This name change removes an ambiguity between function and variable names.  The shell parsing had no problem with that - human parsers did... ;)

marksantcroos:
> It should be merged I think. Despite the large number of changed lines, its actually a rather simple fix, as it removes some code complexity, and from a fairly rarely used code path at that.

And removed functionality, so I would like to have noted that I objected to this change :-)

andre-merzky:
Interesting, I though we discussed this!  What functionality do you consider removed?  

The main difference is that the `rp_install` tree now always lives in the pilot sandbox, and never next to the global sandbox -- but the stack you can configure to get installed in `rp_install` did not change.  In `debug`, that did not give any performance boost either, as the stack was always reinstalled (which is exactly the problem this tries to solve, as pip-locking for that reinstall caused trouble).  What am I missing?

marksantcroos:
`debug` and `local` had different semantics, but discussing them is impossible now as my head can't wrap around the renaming anymore. Let it be.

andre-merzky:
Ok.  Once you stumble over something which you need in this context, let me know...

PR #1178, opened 2017-01-09
Add allegro.json config file for FUB allegro cluster

jhprinz:
Renamed the file and branched from devel. This should be correct now. @andre-merzky


andre-merzky:
Ah, that is the expected diff :D  Thanks JH!

PR #1193, opened 2017-01-27
Fix/ortelib nonmpi

andre-merzky:
attempt to change non-mpi multicore semantics in ortelib


marksantcroos:
What is the driver for this change? It only mentions "change", which it indeed does ...

andre-merzky:
Right, should have explained.

To Sergei and Alessio, it seems that the cores are getting over-utilized on Titan, the behavior seems consistent with too many application instances being created on the nodes.  The CU workloads are multicore, but not MPI - so RP is expected to allocate the requested number of cores on a single node, but to only create one process instance.  

However, in the logs they see `ortesubmit -np 1 -host node_1,node_1,node_1,node_1`
From the documentation of `orterun` this would mean, if I am not mistaken, that one process instance (`-np 1`) is created **per host** specified with `-host` - so we actually end up with 4 process instances in total for the example above, instead of one.

This patch keeps that for MPI applications - but for non-MPI, it only specifies the node once via `-host`.

marksantcroos:
You probably just want to do the same as you did for orte in 08b58e2.

andre-merzky:
Hmm, forgot about that.  That fix back then seems not to correct the `host_string` though.  Am I reading the orterun docs correctly actually, any comment on that?

marksantcroos:
```
netbook:~ mark$ orterun -host netbook,netbook /bin/date
Sat Jan 28 17:38:05 CET 2017
Sat Jan 28 17:38:05 CET 2017
netbook:~ mark$ orterun -np 1 -host netbook,netbook /bin/date
Sat Jan 28 17:38:08 CET 2017
```

marksantcroos:
Anyway, stripping duplicate hosts in case of `-np 1` is fine of course.

The semantics of checking for multinode vs single node is already done by the scheduler.

marksantcroos:
Note that we can use `export OMPI_MCA_hwloc_base_binding_policy=none` to do this on a per CU basis.

andre-merzky:
This PR's diff needs to be cleaned up after devel merge - its larger than it should be.  This should be re-opened toward v0.46.

ibethune:
Andre, should this one be reopened then?

andre-merzky:
yes, it should...

ibethune:
Not sure what to do with this - I don't think the PR can be re-opened as it was towards split_module_2 (which no longer exists), but it doesn't seem possible to create a new one towards devel while this exists... @andre-merzky do you know what to in this case?

andre-merzky:
Seems we lost track of this one - I reapplied the fixes to devel by now, and this can remain closed.

It seems that github nowadays allows to rebase a PR to a different branch - which what we now can do in the future instead of closing/reopening.  This can be done via the 'edit' button at the top which formerly only allowed titel editing.

PR #1200, opened 2017-02-06
Feature/shark

marksantcroos:
I'm a researcher from the Netherlands and this PR enables RP on our cluster ;-)

### DIFF
--- a/examples/config.json
+++ b/examples/config.json
@@ -208,5 +208,12 @@
             "queue"    : null,
             "schema"   : "gsissh",
             "cores"    : 1
+        },
+
+        "lumc.shark"   : {
+            "project"  : null,
+            "queue"    : null,
+            "schema"   : "ssh",
+            "cores"    : 16
         }
 }
--- a/src/radical/pilot/agent/rm/sge.py
+++ b/src/radical/pilot/agent/rm/sge.py
@@ -35,18 +35,27 @@ class SGE(LRMS):
 
         # Parse SGE hostfile for nodes
         sge_node_list = [line.split()[0] for line in open(sge_hostfile)]
-        # Keep only unique nodes
-        sge_nodes = list(set(sge_node_list))
-        self._log.info("Found PE_HOSTFILE %s. Expanded to: %s", sge_hostfile, sge_nodes)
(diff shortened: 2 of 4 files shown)


andre-merzky:
We are cautious to Dutch pull requests around here nowadays...  but since you are smiling on your profile picture ... ;-)

Thanks for getting the SGE RM up to par!  Are there any changes on the SAGA layer needed?  SGE is untested since quite a while...

General question: you have this code comment:

> # In case of non-homogeneous counts, consider all slots be single core 

Do we actually care if a slot is single- or multicore?  Our `cores` attribute in the CU description really maps to a `slot` in this case, doesn't it?


I am surprised that the scattered scheduler is of use to you.  I assume this means your workload is heterogeneous in core size and runtimes?   Eitherway, glad to see that one getting used, too.

I did not test either change, but the PR looks ok, and since the changes are isolated and neither SGE nor the scattered agent scheduler has any other users (that I know of anyway), happy to accept the PR as is.

marksantcroos:
> Are there any changes on the SAGA layer needed?

Nope.

> nor the scattered agent scheduler has any other users (that I know of anyway)

Unlikely, as this PR added it :)

> Do we actually care if a slot is single- or multicore? Our cores attribute in the CU description really maps to a slot in this case, doesn't it?

A slot is always single core. What I meant to say is that `nodeN:3` is translated to `nodeN:1,nodeN:1,nodeN1'.

> I assume this means your workload is heterogeneous in core size and runtimes?

No, it has to do with the heterogeneous core counts and the fact that the continuous scheduler can't deal with that.

andre-merzky:
>> nor the scattered agent scheduler has any other users (that I know of anyway)
> Unlikely, as this PR added it :)

oops...  :)  
 
>> Do we actually care if a slot is single- or multicore? Our cores attribute in the CU description really maps to a slot in this case, doesn't it?
> A slot is always single core. What I meant to say is that nodeN:3 is translated to `nodeN:1,nodeN:1,nodeN1'.
> 
>> I assume this means your workload is heterogeneous in core size and runtimes?
> No, it has to do with the heterogeneous core counts and the fact that the continuous scheduler can't deal with that.

Ack, thanks.

Issue #1160, opened 2016-11-10
Repeated record for UNSCHEDULED unit state


andre-merzky:
see comments in #1155.  An approach for a fix is in the [fix/unscheduled_transition](/radical-cybertools/radical.pilot/tree/fix/unscheduled_transition) branch, which needs testing though.

andre-merzky:
This is fixed with PR #1204

PR #1204, opened 2017-02-12
fix state transition to UNSCHEDDULED

andre-merzky:
to avoid repetition and invalid state ordering'.

This is hopefully the last fix for the module side unit state mess before the split_module branch is up to par...


andre-merzky:
This fixes #1160.

PR #1215, opened 2017-02-14
Refer to config.json rather than resources.json

ibethune:
Suggest this goes into either a future release or is merged if there is a 0.45.RC2


andre-merzky:
thanks!

Issue #1211, opened 2017-02-13
Localhost Anaconda execution fails


iparask:
As reported by George:
RADICAL Stack:

```
(test_ve)georgeha@js-17-215:~/radical.pilot.sandbox/rp.session.js-17-215.jetstream-cloud.org.georgeha.017210.0007-pilot.0000$ radical-stack
python : 2.7.6
virtualenv : /home/georgeha/test_ve
radical.utils : v0.45.RC1@detached-v0.45.RC1
saga-python : v0.45.RC1@detached-v0.45.RC1
```
[Archive.zip](https://github.com/radical-cybertools/radical.pilot/files/772364/Archive.zip)

George please report RP's version also.

iparask:
George, 

can you test it please?

georgeha:
Hi Giannis,

I run a couple of examples and it works.

iparask:
Based on PR #1219, the fix has been merged in devel. Thus, I consider the ticket resolved.

PR #1219, opened 2017-02-14
Resolving issue #1211.

iparask:
This is the PR based on the above mentioned ticket.


andre-merzky:
Thanks a bunch, Iannis!  I assume that George tested and confirmed that it works?

iparask:
Yes, check the ticket

andre-merzky:
Ah, right, great! :)

iparask:
I am closing this PR as it was merged with the devel branch. 

Just for the record, the bootstrapped checks whether the installation is for Anaconda python or not. In the case that it is not, it updates setuptools and pip to the versions that are specified. If RP is installed over Anaconda python distribution no update occurs

Issue #1224, opened 2017-02-15
Stampede YARN resource config loads python module wrongly


iparask:
It tries to do
```
module load python/2.7.9
```
and the error from stampede is:

```
login2.stampede(10)$ cat bootstrap_1.err
Lmod has detected the following error: These module(s) exist but cannot be
loaded as requested: "python/2.7.9"

   Try: "module spider python/2.7.9" to see how to load the module(s).
```

iparask:
Closing as resolved

PR #1225, opened 2017-02-15
Updated Config for Stampede and YARN

iparask:
This PR resolves issue #1224.

I have to add that at some point configs should be independent from specific version. If it remains, I think it will cause error like that with every change


marksantcroos:
On the other hand there is less chance that the application functionality changes because of silent usage of new versions. Its a difficult trade-off.

andre-merzky:
On some machines there is no way around a full module specification: the default python module may point to a conda version, or to python 3, and neither would work out of the box.

iparask:
I see. Moving forward with merging

PR #1228, opened 2017-02-16
use new ompi installation on comet

andre-merzky:


andre-merzky:
This needs to be tested by someone to make sure permissions are set correctly.

andre-merzky:
Ah, good point - but I don't understand why we would load mpi4py *for the agent* in the first place, that does not make sense...  I'll remove that.

In general we probably could provide an mpi4py installation alongside with ompi, but I'd rather defer that discussion to later, in the context of testing in general...

vivek-bala:
```
In general we probably could provide an mpi4py installation alongside with ompi,
```

I agree. To be more general, we can document the process to create modules (such as mpi4py, mpi gromacs) against the rp openmpi and attach it to the RTD: at least the modules to be loaded, lua script, etc. This way the users can get any mpi based module X working with RP (in most cases!). Also removes some of the burden from the developers.

PR #1231, opened 2017-02-17
change ompi commit, make commit part of module label

andre-merzky:
As discussed, this changes the ompi commit and adds a config flag to the ompi deployment script.  It also makes the commit hash part of the module flag, to reduce the amount of confusion about used versions.


bin/radicalpilot-deploy-ompi.sh
```diff
@@ -53,7 +53,7 @@ wget https://ftp.gnu.org/gnu/m4/m4-1.4.16.tar.gz
 cd $OMPI_SOURCE
 tar -xvzf $OMPI_DOWNLOAD/m4-1.4.16.tar.gz
 cd m4-1.4.16
-./configure --prefix=$OMPI_TOOLS_PREFIX
+./configure --prefix=$OMPI_TOOLS_PREFIX --disable-pmix-dstore
```
marksantcroos: You might want to consider moving it to the openmpi configure.

andre-merzky:
oh, that's a *great* idea... :-D

PR #1232, opened 2017-02-17
Fix/issue 1226

georgeha:


src/radical/pilot/configs/resource_xsede.json
```diff
@@ -357,12 +357,9 @@
         "task_launch_method"          : "SPARK",
         "mpi_launch_method"           : "MPIRUN_RSH",
         "pre_bootstrap_1"             : [ 
-                                        "module purge",
-                                        "module load TACC",
-                                        "module load python/2.7.12",
+                                        "module load python/2.7.9",
                                         "export TACC_DELETE_FILES=TRUE",
```
andre-merzky: This looks wrong - 2.7.9 is gone, and would not load.  2.7.12 is what we use on stampede, AFAIK.  Is this from a botched merge, maybe?

georgeha:
Hi Andre,

it is the same configuration with wrangler.  ( no spark)
```
login1.wrangler(5)$ module load python/2.7.12
Lmod has detected the following error:  These module(s) exist but cannot be loaded as requested:
"python/2.7.12"

   Try: "module spider python/2.7.12" to see how to load the module(s).
```

python/2.7.12 doesn't work

src/radical/pilot/configs/resource_xsede.json
georgeha: Hi Andre,

it is the same configuration with wrangler. ( no spark)

login1.wrangler(5)$ module load python/2.7.12
Lmod has detected the following error:  These module(s) exist but cannot be loaded as requested:
"python/2.7.12"

   Try: "module spider python/2.7.12" to see how to load the module(s).
python/2.7.12 doesn't work

andre-merzky: Ah, I see - I had (for some reason) assumed that python versions @TACC are in loockstep...

Thanks, that looks ok then!

PR #1251, opened 2017-02-28
raise informative exception on unsupported resource entries

andre-merzky:
This patch supported an additional key `enabled` for our resource configs, which will, if set to `false`, raise a `ValueError` exception when an application attempts to submit a pilot to that specific resource.

This was implemented in the context of the v0.45 release, where we want to ship resource config entries without official support (osg, orte-lib on titan).


marksantcroos:
Thanks.
Note that `"false" == true`.

andre-merzky:
> Thanks.
> Note that "false" == true.

That's true (no pun indented) - but I am not sure if you mean there is a quote-bug somewhere I missed?

marksantcroos:
Nah, it was more a reminder to myself that I actually bumped my head against this at some point with another field.

andre-merzky:
hehe - I am not the only one then... :)

andre-merzky:
This PR corrected the quick'n'dirty disabling of unsupported resource entries toward v0.46.

PR #1259, opened 2017-03-02
Feature/partitions

andre-merzky:


ibethune:
No urgency for this from Vivek, I suggest we defer it.

PR #1277, opened 2017-03-12
update titan config for liborte, use fresh openmpi install inc. libffi

andre-merzky:
I am not exactly sure why this deployment would work differently than what we had, but it *does* run stable for me (tm).  The only real difference is that libffi is installed along openmpi now, and that the module includes the respective settings, too.


andre-merzky:
I should mention that this needs testing by @AA919  before merge (gromacs)

AA919:
I tried it with gromacs but it seems that we have the very same problem.

andre-merzky:
> I tried it with gromacs but it seems that we have the very same problem.

Which one - disappearing units or agent barrier failure?  If it is the first one, can you please send the output of `radical-stack` (to be on the safe side), and send me the python script you are running (please do make your gromacs installation accessible, too - thanks!).

If it's the barrier problem, please send me a tarball of the pilot log, out and err files.  Thanks!

AA919:
Disappearing units.

Stack:

python            : 2.7.9
virtualenv        : /ccs/home/aleang9/ve_titan_last
radical.utils     : 0.45
saga-python       : 0.45.1
radical.pilot     : split-28-g05ca6d4@fix-titan_orte

I run example.py from the archive attached to this message. CUs are defined in CUDef.py.

python example.py 1024 1024 should be more than enough to see the error. If you want to be sure do 2048 2048 (#CUs #cores )


[fireworks.tar.gz](https://github.com/radical-cybertools/radical.pilot/files/841120/fireworks.tar.gz)

andre-merzky:
This basically:

```
 -            "module use --append /lustre/atlas/world-shared/csc230/modules/",
￼ +            "module use --append /lustre/atlas/world-shared/csc230/openmpi/modules/",
```

Issue #1170, opened 2016-12-14
cu.working_directory in split_module branch


marksantcroos:
In the sm branch `cu.working_directory` property is an alias for `sandbox`.
Example 01_unit_details.py uses working directory from the exposed dict, and there the alias doesn't exist.

We could of course add it there too, but why was it renamed in the first place? Are the semantics different?

ibethune:
So this is now tripping up the automated tester, so needs to be fixed in this releae cycle.
s

itomaldonado:
oh, I did not know we had this one, I guess I should have searched. Yes it is not Jenkins, this has been happening on all PR tests.

andre-merzky:
This should be addressed in the `fix/example_01` branch, will open a PR.

andre-merzky:
This is fixed now.

PR #1297, opened 2017-03-27
fix example 01

andre-merzky:
as title says, towards #1170


ibethune:
According to the output from Jenkins, issue #1170 is resolved on this branch.  Just one question before merge - any reason why the number of CUs was decreased 128->2 ?

andre-merzky:
Nope, no reason - this is fixed now, thanks.

ibethune:
Code review FTW :) Happy to merge then.

PR #1301, opened 2017-03-31
Fix/analytics utils

ibethune:
Andre, this was the only urgent issue for Ming, so if you are able to get this merged before diving off into threading/termination work that would be great.


andre-merzky:
Ack - Ming confirmed that the fix is effective.  Just waiting for Jenkins to say something, will merge :)

FWIW, I still feel uncomfortable merging my own PRs *sigh* - can't be helped right now I suppose...

andre-merzky:
Jenkins hangs on the first test.  Merging anyway, as the code changes are in a different code path (only relevant to analysis, not to RP runtime.

PR #1312, opened 2017-04-10
acceptance tests

itomaldonado:
Converted examples 00..03 to test cases using 'unittest'. To run you can:

Plain Python:
```
python -m unittest discover --verbose -s ./tests/acceptance/
```

Using Nose + junit xml output (for Jenkins):
```
# install nose tests
pip install nose

# run tests
nosetests --nologcapture --nocapture --verbose --with-xunit --where="./tests/acceptance/"
```


itomaldonado:
Once this is merged, I will create another PR to convert the rest + in parallel I will move jenkins to using these as the base for testing instead of the examples.

andre-merzky:
Thanks!

PR #1313, opened 2017-04-11
Feature/split termination 4

andre-merzky:
This should be ready to go, but needs testing before merge.

Note that a non-fatal problem remains: the agent will log errors during termination.  It terminates anyway, but those errors should be tracked down.


iparask:
Hello,
The test was executed on Supermic, so that it does not disturb my experiments on Comet. 

When I left the pilot finish completely the execution, the termination should OK. Although, when I ```Ctrl+C``` the execution before submitting a pilot it was also OK.

When the scripts reached anywhere after the pilot submission ( stopped the execution with ```Ctrl+C```) the output looked like:
```
gather results

wait for 128 unit(s)
       \^Cexit requested

--------------------------------------------------------------------------------
finalize

closing session rp.session.js-172-81.jetstream-cloud.org.iparask.017275.0017   \
close unit manager                                                            ok
close pilot manager(Termination)iparask@js-172-81:~/Repositories/RadicalPilotDevel/examples$
```
In addition, when a job was in the queue, the termination did not cancel it. As a result, the job went through the queue and got executed or continued execution.

AA919:
Experiments are ok on TITAN.
AA

andre-merzky:
Thanks @AA919, @iparask 

Iannis, thanks for the note on the pilot termination - that is supposed to work obviously.  I'll check this out, and merge should be delayed until that is understood and fixed.

vivek-bala:
Tested on Comet. All terminations have been successful.

PR #1343, opened 2017-05-09
Feature/updated examples

iparask:
Hello,

I do not know what the procedure should be, but I think it is a good idea to have a PR


andre-merzky:
Closed in favor of #1344

PR #1345, opened 2017-05-14
doc update

### LINKED ISSUE #1242: Create databases 'on the fly'
We have a requirement from one of our collaborators to be able to create databases 'on the fly' given the URL to an existing mongodb instance. See https://github.com/radical-collaboration/GB-SC18/issues/2.

I was checking if this was already supported in RP by doing the following:

```
session = rp.Session(database_url='mongodb://rp:rp@ds015335.mlab.com:15335/', database_name = 'test_newdb')
```
but this fails with:
```
radical.pilot.exceptions.PilotException: Couldn't create new session (database URL 'mongodb://rp:rp@ds015335.mlab.com:15335/test_newdb' incorrect?): command SON([('saslStart', 1), ('mechanism', 'SCRAM-SHA-1'), ('autoAuthorize', 1), ('payload', Binary('n,,n=rp,r=ODg3NjU3Nzg4NjI0', 0))]) on namespace test_newdb.$cmd failed: Authentication failed
```

It might be that it is supported but I'm specifying the parameters incorrectly, do let me know in this case.

But if this feature is not supported, I suppose something like the following should suffice(?):

```
from pymongo import Connection
connection = Connection('mongodb://rp:rp@ds015335.mlab.com','15335')
db = connection['test_newdb']
```

Although this can be done in entk and passed to RP, I think the appropriate place to do this might be RP itself. Thoughts?

andre-merzky:
fixes #1242

### DIFF
--- a/docs/source/installation.rst
+++ b/docs/source/installation.rst
@@ -102,7 +102,9 @@ secured resources, we strongly recommend the usage of a secured MongoDB
 instance!
 
 The ``dbname`` component in the database url can be any valid MongoDB database
-name (ie. it musy not contain dots).  RP will create that DB on the fly.
+name (ie. it musy not contain dots).  RP will not create that DB on the fly
+- but it will create collections in that DB on its own, named after RP session
+IDs.
 
 A MongoDB server can support more than one user. In an environment where
 multiple users use RP applications, a single MongoDB server for all users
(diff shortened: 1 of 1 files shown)


vivek-bala:
Hey @andre-merzky , would you be ok with "RP will not create that DB on the fly and requires the DB to be setup prior to creating the session object. But RP will create collections in that DB on its own, named after RP session IDs" ?

andre-merzky:
Absolutely!  Would you be so kind to update this PR?  Thanks!

vivek-bala:
Done!

andre-merzky:
Thanks!

PR #1367, opened 2017-06-05
use a separate VE for each resource config

### LINKED ISSUE #1363: resource ve is not updated on new dependencies
rp agent now depends on msgpack - but an existing `ve_<resource>` is accepted as functional if it loads, and new dependencies are not installed.  The bootstrapper will then fail when testing for the dependency.   This is a regression unfortunately.  Well, not really maybe, it behaved this way for quite some time - but the release will trigger this problem all over the place now.

andre-merzky:
This removes the 'virtenv' config entry for most resource configs.  That
is now only really needed for shared, globe VEs, as on archer and
supermuc.  The new default is to use
%(global_sandbox)/ve.%(resource).%(rp_version)
as name for the VE.  This fixes two issues:

  - different resource configs need different requirements.  For example
    (and prominently), orte-cffi needs only to be installed for orte and
    ortelib configs
  - when new dependencies are introduced in a new release, then they
    will automatically be installed in the new VE

The first item is not yet taken adavntage of - for that, we need to move
the list of optional pre-requisites to the resource configs, or need to
be more clever in the bootstrapper to figure out what is actually
needed.  But the second item will apply immediately, as otherwise all
existing VEs would break on updates to 0.46.  This fixes #1363

### DIFF
--- a/bin/radicalpilot-version
+++ b/bin/radicalpilot-version
@@ -5,9 +5,8 @@ import sys
 import radical.pilot as rp
 
 if len(sys.argv) > 1 and '-v' in sys.argv:
-    print 'version: %s' % rp.version_detail
-    print 'modpath: %s' % os.path.dirname(rp.__file__)
+    print rp.version_detail
 
 else:
-    print rp.version_detail
+    print rp.version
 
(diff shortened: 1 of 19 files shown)


andre-merzky:
This seems to hold up ok so far - going to merge

Issue #1376, opened 2017-06-22
RP fails on Titan 


jdakka:
I ran a test example using 1 pilot, 16 cores on Titan-ortelib (RP version 0.46)
I get the following errors: 

```================================================================================
 Getting Started (RP version 0.46)                                              
================================================================================

new session: [rp.session.titan-ext7.jdakka.017339.0001]                        \
database   : [mongodb://rp:rp@ds015335.mlab.com:15335/rp]Traceback (most recent call last):
  File "pilot.py", line 38, in <module>
    session = rp.Session()
  File "/lustre/atlas2/csc230/proj-shared/mskcc/rp_experiments/venv/lib/python2.7/site-packages/radical/pilot/session.py", line 257, in __init__
    self._components = ruc.start_components(self._cfg, self, self._log)
  File "/lustre/atlas2/csc230/proj-shared/mskcc/rp_experiments/venv/lib/python2.7/site-packages/radical/pilot/utils/component.py", line 252, in start_components
    comp.start()
  File "/lustre/atlas2/csc230/proj-shared/mskcc/rp_experiments/venv/lib/python2.7/site-packages/radical/utils/process.py", line 462, in start
    timeout))
RuntimeError: unexpected child message (alive) [5.0]
```

And from root message: 

```
PBS Job Id: 3458802
Job Name:   pilot.0000
Exec host:  2342/0-15
An error has occurred processing your job, see below.
request to copy stageout files failed on node '2342' for job 3458802

Unable to copy file /autofs/nccs-svm1_home1/jdakka/3458802.OU to /lustre/atlas/scratch/jdakka/csc230/radical.pilot.sandbox/rp.session.titan-ext4.jdakka.017339.0000/pilot.0000/bootstrap_1.out, error 1
*** error from copy
/bin/cp: cannot stat `/autofs/nccs-svm1_home1/jdakka/3458802.OU': No such file or directory
*** end error output

Unlink of stage out file /autofs/nccs-svm1_home1/jdakka/3458802.ER failed
```

AA919:
I have the same problem. I tried to run experiments but the folders of the units are not copied on Lustre. I also tried to delete pilot's virtualenv but RP is not able to create a new one. Yesterday Titan had a downtime, probably they updated the system and the update is bothering RP.

vivek-bala:
This might not be the root cause of the problem. But I think you need a minimum of 2 nodes (=32 cores on Titan) if you are using orterun/ortelib.

[...]

PR #1378, opened 2017-06-24
bootstrapper can optionally use system virtualenv

andre-merzky:
This fixes the first part of #1376.  Alas, titan still barfs because the pyzmq installation is not usable.  Did I complain about python deployment lately?  No?  Well, today is the day... :/

### DIFF
--- a/src/radical/pilot/agent/bootstrap_1.sh
+++ b/src/radical/pilot/agent/bootstrap_1.sh
@@ -54,6 +54,8 @@ CCM=
 PILOT_ID=
 RP_VERSION=
 PYTHON=
+PYTHON_DIST=
+VIRTENV_DIST=
 SESSION_ID=
 SESSION_SANDBOX=
 PILOT_SANDBOX=`pwd`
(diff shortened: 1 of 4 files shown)


andre-merzky:
> I have some general remarks:
> 
> The file bootstrap_1.sh is very large for a script (1850+ lines) and the code is not divided into functions to make it easy to understand its logic.
> 
> In addition to dividing the code into functions, perhaps some functions could be factored into a separate file that is then sourced by this bootstrap script and possibly by other scripts.

Yes, its long indeed - I would *love* to have it simpler *sigh*.  But I don't understand the part about functions: about 1200 lines out of the 1800 *are* in functions.  The `main` (from line 1335 onward) is still long, and some parts could be factored out for sure, but still...

As for splitting into multiple files: the use case for the bootstrapper is to be staged to the target host and to be executed there.  While one can stage multiple files just as well, that was at least originally the motivation to keep it self-contained.

Nowadays we actually *do* stage more than the bootstrapper: several files and dirs are now put in a tarball, staged, and then unpacked, so we could reconsider a split into multiple files w/o additional penalties I think - thanks for pointing that out.


> My favorite approach to Python deployment is to do as much work as possible with a tool like Ansible, and use (bash) scripts where there is no available Ansible module. An example of Python deployment I have done with Ansible is here:
> 
> https://github.com/gmateesc/WebAppDeploymentAutomation

Alas, the target hosts we are working with usually don't suppor ansible nor many of the other software  management solutions (chef, puppet, salt etc), and I am hesitant to commit supporting any those on our own.  Pip and anaconda,  which are the ones we *do* support, have their own problems without a doubt (if you search for `NOTE:` in the bootstrapper, you'll find some of those documented), but at least they are kind-of part of default python installations.

Thanks, Andre.

gmateesc:
> Alas, the target hosts we are working with usually don't support 
> ansible nor many of the other software management solutions 
> (chef, puppet, salt etc), and I am hesitant to commit supporting 
> any those on our own.

One can use Ansible on a supercomputer that has stripped down 
compute nodes. Ansible only needs to be installed on the head node 
of the cluster, and it will ssh to the target node(s) where it will deploy 
the Python environment. 

If the target node is a compute node that should be accessed only 
through a job manager like SLURM, then this can be achieved as follows:

1. Submit a batch job that requests an allocation on the head node 
   and on the target node(s) where installation is needed;

2. The job will start on the head node and it will be a script that runs an 
    Ansible playbook, passing to the playbook the target node(s) as the 
    list of hosts ("inventory", in Ansible speak) to which Ansible must 
    do the deployment; Ansible will ssh to the target node(s) and do the 
    deployment;

Issue #1377, opened 2017-06-24
RP fails on Supermic


AA919:
Hi,
I tried to use Supermic with RP today with a virtualenv that has always worked but RP returns this error:
```bash
        -2017-06-24 20:29:49,374: radical.saga.pty    : umgr.staging.input.0            : MainThread     : ERROR   : read from process failed '[Errno 5] Input/output error' : (Connection closed
) (/home/aangius/src-devel/ve_devel/local/lib/python2.7/site-packages/saga/utils/pty_process.py +679 (read)  :  % (e, self.tail)))
Traceback (most recent call last):
  File "/home/aangius/src-devel/ve_devel/local/lib/python2.7/site-packages/saga/utils/pty_shell_factory.py", line 307, in _initialize_pty
    n, match = pty_shell.find (prompt_patterns, delay)
  File "/home/aangius/src-devel/ve_devel/local/lib/python2.7/site-packages/saga/utils/pty_process.py", line 790, in find
    raise ptye.translate_exception (e, "(%s)" % data)
NoSuccess: read from process failed '[Errno 5] Input/output error' : (Connection closed
) (/home/aangius/src-devel/ve_devel/local/lib/python2.7/site-packages/saga/utils/pty_process.py +679 (read)  :  % (e, self.tail)))
```

the stack is this one : 
 python               : 2.7.6
  pythonpath           : 
  virtualenv           : /home/aangius/src-devel/ve_devel

  radical.pilot        : split-71-gfdd36239@devel
  radical.utils        : v0.45-46-g591b33d@devel
  saga                 : split-18-gc074c761@devel

Nothing is copied in the sandbox except the creation of the folder rp* pilot.00000 and unit.000000 (Note that I had more than one unit).

andre-merzky:
Hey Ale, could you please send me a tarball of the pilot sandbox?  Thanks!

andre-merzky:
Ale, this seems to be an system problem after all.  Can you please try to switch to plain ssh?  Not sure if your account allows that (mine does not).

andre-merzky:
Supermic admins resolved the sftp issue - can you please give it a try?  Thanks!

AA919:
I am having problems but of a different kind. I am still figuring out if its on me or on RP. I will keep you posted.

AA919:
It works although I have seen that some of the CU fails (around a 100 out 2056). Tomorrow I will try to find at least one of those that failed. In the mean time I will close the ticket.

PR #1379, opened 2017-06-26
Feature/download

andre-merzky:
Session can now download json, log- and profiles on close().
Note that the deprecated argument `delete` has been removed, and
`cleanup` now default to `False`.  Most experiments will now likely
want:

```
session.close(download=True)
```
which results in a suitable configuration for RA.


mturilli:
Successfully tested with `example/00_getting_started.py` on supermic with ssh (gsissh is broken there as per ticket #1377).

PR #1381, opened 2017-07-04
Fixed Syntax Error in visit(...) function

chengwill97:
Prevented me from installing this package, which is how I found it.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -193,7 +193,7 @@ def makeDataFiles(prefix, dir):
     os.path.walk(dir, visit, (prefix, strip, found))
     return found
 
-def visit((prefix, strip, found), dirname, names):
+def visit(prefix, strip, found), dirname, names):
     """ Visit directory, create distutil tuple
 
     Add distutil tuple for each directory using this format:


itomaldonado:

chengwill97:
Sorry, this is the error I was getting when installing

Complete output from command python setup.py egg_info:
      File "<string>", line 1, in <module>
      File "/private/var/folders/t2/bxrhj1dd3pdb8r7rv8vjjs3r0000gn/T/pip-build-68_16zml/radical.pilot/setup.py", line 193
        def visit((prefix, strip, found), dirname, names):
                  ^
    SyntaxError: invalid syntax

chengwill97:
Update:
Sorry for wasting your time. I was using Python3, which doesn't allow tuple unpacking, not 2.7

I thought the requirements in link below which was for Python >= 2.7 meant I could use Python3, too
http://radicalpilot.readthedocs.io/en/latest/installation.html

andre-merzky:
Ah, thanks for your update, I did not understand what the diff was trying to fix :)  

We don't really support Python3, yet, unfortunately.  There exists a set of patches to the stack which supposedly fixes that, but due to lack of user demand and time we did not yet confirm and merge those.  Thus the question: how important is Python3 for you?

Thanks, Andre.

PR #1386, opened 2017-07-24
Fix/issue 1361

andre-merzky:
This is a new attempt to improve resource RP's resource handling.  More specifically, it reduces the reliance on Python's garbage collection to destroy object handles, and indirectly ZMQ message channels.  

Note that this is not a complete fix: repeated creations of RP sessions in the same application (more precisely in the same Python interpreter instance) will still leak resources, and `select()` will complain about too many socket handles after about 60 sessions (default configuration, on Linux).

### DIFF
--- a/docs/architecture/zmq/test_zmq.py
+++ b/docs/architecture/zmq/test_zmq.py
@@ -6,18 +6,11 @@ import zmq
 import time
 import multiprocessing as mp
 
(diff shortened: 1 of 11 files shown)


andre-merzky:
PS.: note that this needs the RU branch `fix/issue_rp_1361`

PR #1406, opened 2017-08-02
Feature/stampede2

iparask:
I may have branched off devel. If you can cherry pick the last commit it is okay. Otherwise I can delete the branch and redo the changes.

### DIFF
--- a/src/radical/pilot/agent/bootstrap_1.sh
+++ b/src/radical/pilot/agent/bootstrap_1.sh
@@ -80,7 +80,7 @@ LOCK_TIMEOUT=600 # 10 min
 VIRTENV_TGZ_URL="https://pypi.python.org/packages/source/v/virtualenv/virtualenv-1.9.tar.gz"
 VIRTENV_TGZ="virtualenv-1.9.tar.gz"
 VIRTENV_IS_ACTIVATED=FALSE
-VIRTENV_RADICAL_DEPS="pymongo==2.8 apache-libcloud colorama python-hostlist ntplib pyzmq netifaces==0.10.4 setproctitle orte_cffi msgpack-python future"
+VIRTENV_RADICAL_DEPS="colorama pymongo==2.8 apache-libcloud colorama python-hostlist ntplib pyzmq netifaces==0.10.4 setproctitle orte_cffi msgpack-python future"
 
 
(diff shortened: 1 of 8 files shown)


src/radical/pilot/agent/bootstrap_1.sh
```diff
@@ -80,7 +80,7 @@ LOCK_TIMEOUT=600 # 10 min
 VIRTENV_TGZ_URL="https://pypi.python.org/packages/source/v/virtualenv/virtualenv-1.9.tar.gz"
 VIRTENV_TGZ="virtualenv-1.9.tar.gz"
 VIRTENV_IS_ACTIVATED=FALSE
-VIRTENV_RADICAL_DEPS="pymongo==2.8 apache-libcloud colorama python-hostlist ntplib pyzmq netifaces==0.10.4 setproctitle orte_cffi msgpack-python future"
+VIRTENV_RADICAL_DEPS="colorama pymongo==2.8 apache-libcloud colorama python-hostlist ntplib pyzmq netifaces==0.10.4 setproctitle orte_cffi msgpack-python future"
```
andre-merzky: colorama is already in the list, isn't it?

iparask: Yes indeed. I added because I was getting an error that there was no such module. Let me give it another try, because this is weird.

Is there a chance that the order is important?

andre-merzky: No, order should not matter.  I also have seen colorama dependency errors from time to time - but it seemed like pip incompatibilities or firewall issues at the end, so that *no* dependency got installed, and colorama was simply the first one the stack complained about...  Please check out the bootstrapper out/err files if that pops up again.

iparask: Okay. I will get back with a new commit

andre-merzky:
Iannis, any news on this one?  I would like to prepare another release candidate... :)

iparask:
I assume an error like:
```
Couldn't install orte_cffi! Lets see how far we get ...
```
is okay since there is no ORTE on Stampede2?

It run without the second colorama dependency

andre-merzky:
Indeed, the cffi error can be ignored in this case.  Thanks!

Issue #1426, opened 2017-08-30
07_shared_unit_data.py fails with RP release


vivek-bala:
See error in pmgr.launching:
```
2017-08-30 14:17:59,781: pmgr.launching.0.child: pmgr.launching.0                : MainThread     : ERROR   : bulk launch failed
Traceback (most recent call last):
  File "/home/vivek/ves/giannis-test/local/lib/python2.7/site-packages/radical/pilot/pmgr/launching/default.py", line 329, in work
    self._start_pilot_bulk(resource, schema, pilots)
  File "/home/vivek/ves/giannis-test/local/lib/python2.7/site-packages/radical/pilot/pmgr/launching/default.py", line 416, in _start_pilot_bulk
    info = self._prepare_pilot(resource, rcfg, pilot)
  File "/home/vivek/ves/giannis-test/local/lib/python2.7/site-packages/radical/pilot/pmgr/launching/default.py", line 819, in _prepare_pilot
    bootstrap_args += " -d '%s'" % ':'.join(sdist_names)
TypeError: sequence item 2: expected string, NoneType found
```

Tested with MacOS/Anaconda, Ubuntu/Anaconda, Ubuntu/GCC (os/python)

vivek-bala:
[rp.session.vivek-HP-Pavilion-m6-Notebook-PC.vivek.017408.0000.zip](https://github.com/radical-cybertools/radical.pilot/files/1264655/rp.session.vivek-HP-Pavilion-m6-Notebook-PC.vivek.017408.0000.zip)

andre-merzky:
Uh, this is somewhat surprising, as this was explicitly part of the test protocol, wasn't it?  So did it work back then?  Also, that code path is the same for all examples - I assume that the other examples also don't work?

You have this file in your launcher log:

```
2017-08-30 14:17:58,668: pmgr.launching.0.child: pmgr.launching.0                : MainThread     : INFO    : ('0.46.1', '0.46.1', None, None, None, '/home/vivek/ves/giannis-test/local/lib/python2.7/site-packages/radical/pilot/pmgr/launching/None')
```

This seems do indicate an incomplete / faulty installation of RP, as the sdist is indeed not found.  The output should be something like:

```
[...]

PR #1431, opened 2017-08-31
This most probably solves issue #1426 

iparask:
Andre, if you agree with the change? #1426


andre-merzky:
Thanks for the PR, Iannis!  Before merging, I want to spend some more time trying to understand why and how that actually broke, just to make sure we don't screw this up for other deployment variants...

iparask:
I agree! I suggest to have a look also to RU get_version method. The if statement that creates the sdist_path is outside the loop that checks the paths that are passed. Another thing I noticed, is that when a specific branch is used that method fills all the values it returns. As a consequence PMGR is able to find the tar balls.

PR #1479, opened 2017-10-24
Updating SuperMIC pre_bootstrap

mingtaiha:
Explicitly specifying the module. Tested successfully using the following stack running `00_getting_started.py`:

```
  python               : 2.7.12
  pythonpath           : 
  virtualenv           : /home/mingtha/rp_testing/supermic_testing

  radical.pilot        : 0.47-v0.46.2-143-g4c99ec74@fix-1478
  radical.utils        : 0.47-v0.46-63-gc1ae8ac@rc-v0.46.3
  saga                 : 0.47-v0.46-20-g8ea2302c@rc-v0.46.3
```

It would be great it someone can confirm this

### DIFF
--- a/src/radical/pilot/configs/resource_xsede.json
+++ b/src/radical/pilot/configs/resource_xsede.json
@@ -720,7 +720,7 @@
         "agent_launch_method"         : "SSH",
         "task_launch_method"          : "SSH",
         "mpi_launch_method"           : "MPIEXEC",
-        "pre_bootstrap_1"             : ["module load python"],
+        "pre_bootstrap_1"             : ["module load python/2.7.7/GCC-4.9.0"],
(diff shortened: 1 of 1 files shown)


andre-merzky:
Thanks Ming, this looks good.  We do something similar for other resources, too.  This will be covered in our release testing toward end of the week, so I don't mind merging without more confirmation at this point.

mingtaiha:
No problem Andre. Just to be safe, I need to stress again that a new virtual environment may need to be built again. So it is worthwhile to include it in the testing protocol.

andre-merzky:
That's a fair point, good catch!  But a version bump will trigger a new VE anyway though.  That mechanism was specifically introduced to support such configuration changes without the need for user interaction.

PR #1482, opened 2017-10-26
specifying python used by agent, default changed on supermic

vivek-bala:


vivek-bala:
Tested with a simple example.

andre-merzky:
How does that relate to https://github.com/radical-cybertools/radical.pilot/pull/1479, which also changes Python on Supermic?

vivek-bala:
I tried loading the "python/2.7.7/GCC-4.9.0" for the agent but that still failed for me. I remember the default python loaded was "python/2.7.7-anaconda" before this change on superMIC. Sorry I didn't see that PR, otherwise would have committed in that PR itself.

andre-merzky:
No worries on missing the PR - but can you please sync with @iparask  on what the best course of action is?  Thanks!

vivek-bala:
I am going to test again, just to be sure. Will wait for @iparask to recommend solution.

iparask:
This anaconda distribution can use the virtualenv that rp uses. I am starting to believe that Supermic uses Anaconda python. Let's verify and if that is correct then we need to change  the python distribution from default to anaconda. 

GCC python: python/2.7.10-mkl-mic
Anaconda Python: python/2.7.13-anaconda-tensorflow
GCC pythono: python/2.7.7/GCC-4.9.0
Anaconda python: python/2.7.7-anaconda (supports virtualenv probably)

I think there is a ticket that discusses the python version of SuperMic. I cannot find it now

vivek-bala:
I can confirm that RP fails with python/2.7.7/GCC-4.9.0 and succeeds with python/2.7.7-anaconda. Tested again now.

andre-merzky:
Guys, what is the conclusion on this one?

andre-merzky:
Folx, is there a conclusion on this?

iparask:
It works for Vivek. So, I suggest to go with this change

andre-merzky:
Uhm, which one - that was the question, wasn't it? ;)

iparask:
The change the PR has

andre-merzky:
Iannis, you realize that I am trying to figure out *which* PR you guys want to keep, right?  So, is it #1479 or #1482?

vivek-bala:
Not sure if you guys discussed this, but I think the 1482 is the fix. 1479 does not resolve the issue.

vivek-bala:
I believe, this PR didn't make it into the release this morning (v0.47). It still says gcc python in https://github.com/radical-cybertools/radical.pilot/blob/master/src/radical/pilot/configs/resource_xsede.json#L724. So, the latest release is failing on superMIC.

vivek-bala:
The pre_bootstrap_1 needs to have 'module load python/2.7.7-anaconda' as the value.

PR #1486, opened 2017-10-30
Non-MPI alloc method commented.

iparask:
Not-sure yet how to optimize. Although I think now it is a bit easier to understand without @andre-merzky  explaining.


src/radical/pilot/agent/scheduler/continuous.py
```diff
@@ -255,50 +262,67 @@ def _get_node_maps(self, cores, gpus, threads_per_proc):
...
+        # For threads, if the entry is not specified in the description, then 
+        # assume it is one thread per process.
         requested_procs  = cud['cpu_processes']
-        threads_per_proc = cud['cpu_threads']
+        threads_per_proc = cud.get('cpu_threads',1)
         requested_gpus   = cud['gpu_processes']
 
-        if not threads_per_proc:
-            threads_per_proc = 1
+        # The total number of requested logical cores is equal to the number of
```
andre-merzky: The change above (from `[]` to `get()` and removal of the check) will miss a case.  Which one? :)

src/radical/pilot/agent/scheduler/continuous.py
```diff
@@ -255,50 +262,67 @@ def _get_node_maps(self, cores, gpus, threads_per_proc):
...
+        # processes times the number of threads per process.
 
         requested_cores = requested_procs * threads_per_proc
 
-        # we want the allocation to fit on a single node - make sure this is
-        # actually possible
+        # The allocation has to fit in a single node. If this is true contnue and
+        # try to find a possible allocation, otherwise raise a value error exception.
         if  requested_cores > self._lrms_cores_per_node or \
```
andre-merzky: 'fit in' -> 'fit on'
'contnue' -> 'continue'

Again I am not sure if the change really adds any information, apart from spelling out `if a>b: raise ValueError`, which is very readable in the code already.

src/radical/pilot/agent/scheduler/continuous.py
```diff
@@ -255,50 +262,67 @@ def _get_node_maps(self, cores, gpus, threads_per_proc):
...
             requested_gpus  > self._lrms_gpus_per_node     :
             raise ValueError('Non-mpi unit does not fit onto single node')
 
-        # ok, we can go ahead and try to find a matching node.
+        # Initiallize core and gpu list that the unit may allocate. Also initialize
+        # node name and node id variables.
         cores     = list()
```
andre-merzky: Isn't this obvious?  You say:
```
cores = list()  # initialize core list
```
etc.  Well, duh! :)

src/radical/pilot/agent/scheduler/continuous.py
```diff
@@ -255,50 +262,67 @@ def _get_node_maps(self, cores, gpus, threads_per_proc):
...
 
-            # attempt to allocate the required number of cores and gpus on this
+            # attempt to find the required number of cores and gpus on this
             # node - do not allow partial matches.
             cores, gpus = self._alloc_node(node, requested_cores, requested_gpus,
                                            partial=False)
 
+            # if the nessecary number of cores and GPUs is found, break the loop.
+            # Else check the next node.
             if  len(cores) == requested_cores and \
```
andre-merzky: there is no `else` ;)  Maybe 'otherwise'?

src/radical/pilot/agent/scheduler/continuous.py
```diff
@@ -255,50 +262,67 @@ def _get_node_maps(self, cores, gpus, threads_per_proc):
...
-            return None  # signal failure
+            return None
 
-        # we have to communicate to the launcher where exactly processes are to
-        # be placed, and what cores are reserved for application threads.
+        # A suitable allocation was found and the core and gpu map, core and gpu
+        # ids, should be return to the launcher. This way the processes will be
+        # placed in cores with distance equal to the number of threads, leaving
+        # enough space between them.
         core_map, gpu_map = self._get_node_maps(cores, gpus, threads_per_proc)
```
andre-merzky: `distance` is actually not correct.  Assume we want `n_procs=6, threads_per_procs=2` and find cores `0, 1, 2, 3, 4, 5, 7, 8` - then they are not equal-distant.
Maybe: 
```
 +        # A suitable allocation was found and the core and gpu map, core and gpu
￼ +        # ids, should be return to the launcher, informing about the exact placement
 +        # of processes and threads, respectively. 
```
?

src/radical/pilot/agent/scheduler/continuous.py
```diff
@@ -326,14 +350,16 @@ def _alloc_mpi(self, cud):
         'threads_per_proc', as otherwise the application would not be able to
         spawn the requested number of threads on the respective node.
         '''
-
+
```
andre-merzky: For some reason, most people (and checkers) dislike spaces at the end of lines, like on the empty line above.  Don't ask me why, but there you go...

src/radical/pilot/agent/scheduler/continuous.py
```diff
@@ -326,14 +350,16 @@ def _alloc_mpi(self, cud):
...
+        # For threads, if the entry is not specified in the description, then 
+        # assume it is one thread per process.
         requested_procs  = cud['cpu_processes']
-        threads_per_proc = cud['cpu_threads']
+        threads_per_proc = cud.get('cpu_threads',1)
         requested_gpus   = cud['gpu_processes']
 
-        if not threads_per_proc:
-            threads_per_proc = 1
-
```
andre-merzky: Same comment as above: you now miss a case.

src/radical/pilot/agent/scheduler/continuous.py
```diff
@@ -326,14 +350,16 @@ def _alloc_mpi(self, cud):
...
+        # The total number of requested logical cores is equal to the number of
+        # processes times the number of threads per process.
         requested_cores = requested_procs * threads_per_proc
```
andre-merzky: What does that comment add which wasn't readable before?  Do the variable names need changing?

src/radical/pilot/agent/scheduler/continuous.py
```diff
@@ -123,16 +123,19 @@ def _allocate_slot(self, cud):
         on the same node).
         '''
 
-        # single_node is enforced for non-message passing tasks
+        # Checking whether the unit is MPI or not. Resource allocation is different
+        # between MPI and non-MPI units. The first can span over different nodes
+        # while the second shoud fit in a node.
```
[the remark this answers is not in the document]
iparask: Yes, but again. Comments should be as descriptive as the code. I understand it a bit over the edge, but that's how I try to write comments.

src/radical/pilot/agent/scheduler/continuous.py
```diff
@@ -142,11 +145,15 @@ def _allocate_slot(self, cud):
...
+        This method is envoked when a unit does not need the resources that it
+        had acquired. 
+
+        The input passed are the resource slots a unit had acquired when 
+        `_allocate_slots()` was called.
         '''
 
-        # reflect the request in the nodelist state.
+        # reflect the request in the nodelist state and change the state of the
+        # slots from BUSY to FREE.
```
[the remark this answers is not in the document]
iparask: It adds the state transition.Otherwise the question is what is the state of the slots before? I will have to search all the code to find what is going on. When looking in a method or function the comments should give all the information to understand the method/function without having to understand the class and the base class.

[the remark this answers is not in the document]
andre-merzky: Did you actually check if this is the case? :-P  Because, the method actually just sets the given values.  You may argue that is not what it should do, but right now your comment does not reflect what the code does.  Wrong comments are worse than no comments...

src/radical/pilot/agent/scheduler/continuous.py
andre-merzky: Yup...

src/radical/pilot/agent/scheduler/continuous.py
andre-merzky: Yes, threads can go to any cores, they don't need to be next to each other.  The only limitation really is that they live on the same OS instance - which we assume to be the case for nodes in Beowulf-like clusters.

Locating threads on the same physical CPU (in multi-CPU nodes) has some performance benefits, but we do not explicitly attempt to exploit that.

PR #1488, opened 2017-10-31
Questions and comments on continuous scheduler

mingtaiha:


src/radical/pilot/agent/scheduler/base.py
```diff
@@ -241,6 +241,10 @@ def initialize_child(self):
         # during agent startup.  We dig them out of the config at this point.
         #
         # NOTE: this information is insufficient for the torus scheduler!
+
+        # MING: What is lm info? All the other things make sense, but lm_info
+        #       is a bit unclear.
+
```
andre-merzky: This is documented here: https://github.com/radical-cybertools/radical.pilot/blob/feature/gpu/src/radical/pilot/agent/scheduler/base.py#L113  Would you mind adding a link/reference?

src/radical/pilot/agent/scheduler/base.py
```diff
@@ -310,14 +314,33 @@ def create(cls, cfg, session):
     #
     # NOTE: any scheduler implementation which uses a different nodelist
     #       structure MUST overload this method.
+
+    # MING: This function is used to update the node list by taking in a list
+    #       of slots that have been assigned an update them to the new state.
+    #       The nodes are represented as a dictionary, where as the slots are
+    #       represented as a dictionary of a different structure. This confused
+    #       me at first when I tried to read it.
+
```
andre-merzky: I can see that.  Can you add references to the documentation of both structures?  Those should be documented in the base class again.  Hmm, I wonder if that method should actually also be moved to the base class.  Any thoughts?

src/radical/pilot/agent/scheduler/base.py
```diff
@@ -310,14 +314,33 @@ def create(cls, cfg, session):
...
         for node_name, node_uid, cores, gpus in slots['nodes']:
 
             # Find the entry in the the slots list
+
+            # MING: Assuming 'uid' is the ID of the node, it seems a bit
+            #       wasteful to have to look at all of the nodes available
+            #       for use if at most one node can have that uid. Maybe
+            #       it would be worthwhile to simply keep a list of nodes
+            #       that we would read, and keep a dictionary that maps
+            #       the uid of the node to the location on the list?
```
andre-merzky: So, basically creating an uid-lookup index into the list? Yeah, makes sense as an optimization, since that index would be static over the lifetime of the agent.  Would you be up to implementing this?  Otherwise we could put it up for a task assignment, its useful, small and fairly self-contained...  Nice one!

src/radical/pilot/agent/scheduler/base.py
```diff
@@ -310,14 +314,33 @@ def create(cls, cfg, session):
...
+
             node = (n for n in self.nodes if n['uid'] == node_uid).next()
             assert(node)
 
+            # MING: This is a bit awkward to read since it node is the node
+            #       dictionary created during initialization, but cores 
+            #       refers to the core map associated with the slots. Can
+            #       be tricky to read at first. Same for GPUs
             for cslot in cores:
```
andre-merzky: Ack.  Any suggestion on how to improve?

src/radical/pilot/agent/scheduler/continuous.py
```diff
@@ -98,6 +98,12 @@ def _configure(self):
         # NOTE: for non-oversubscribing mode, we reserve a number of cores
         #       for the GPU processes - even if those GPUs are not used by
         #       a specific workload.
+
+        # MING: What do we do when nodes are oversubscribed? Do we use
+        #       something else?
+
+        # MING: It seems that there are two separate ways for how we track the
+        #       the cores allocated to the
```
andre-merzky: This comment looks incomplete?

src/radical/pilot/agent/scheduler/continuous.py
```diff
@@ -98,6 +98,12 @@ def _configure(self):
         # NOTE: for non-oversubscribing mode, we reserve a number of cores
         #       for the GPU processes - even if those GPUs are not used by
         #       a specific workload.
+
+        # MING: What do we do when nodes are oversubscribed? Do we use
+        #       something else?
```
andre-merzky: Yeah, we simply don't reserve cores.  Given your comment that seems not obvious from the `NOTE` above.  Any suggestion how to improve it?

src/radical/pilot/agent/scheduler/continuous.py
```diff
@@ -114,7 +120,7 @@ def _configure(self):
 
 
     # --------------------------------------------------------------------------
-    #
+    
     def _allocate_slot(self, cud):
```
andre-merzky: Ey, that line was intentional! :-P

src/radical/pilot/agent/scheduler/continuous.py
```diff
@@ -175,6 +181,7 @@ def _alloc_node(self, node, requested_cores, requested_gpus,
         requested (but the call will never return more than requested).
         '''
 
+        # MING: Maybe we can change the function name to _find_alloc_in_node ?
```
andre-merzky: Maybe :)

src/radical/pilot/agent/scheduler/continuous.py
```diff
@@ -202,11 +209,17 @@ def _alloc_node(self, node, requested_cores, requested_gpus,
 
         # we can serve the partial or full request - alloc the chunks we need
         # FIXME: chunk gpus, too?
+
+        # MING: We don't really need to include chunk in the arithmetic if
+        #       they cancel out right?
         alloc_cores = min(requested_cores, free_cores) / chunk * chunk
```
andre-merzky: Hmm, a comment is needed it seems :)  Assume the following: `requested_cores=10, free_cores=8, chunk=5` - what is returned if that term (`/chunk*chunk`) is left off?  Why?  Is there a better way to achieve this?
Please come up with a comment to clarify this.  No need to do that before the call, I'll bring this up today.  Nice catch!

src/radical/pilot/agent/scheduler/continuous.py
```diff
@@ -202,11 +209,17 @@ def _alloc_node(self, node, requested_cores, requested_gpus,
...
         alloc_gpus  = min(requested_gpus , free_gpus )
 
         # now dig out the core and gpu IDs.
         for idx,state in enumerate(node['cores']):
+            
+            # MING: Just to check, is this true only when there are no free
+            #       cores?
```
andre-merzky: Hmm, no, that should break when no more cores are _needed_.  Or am I missing the point of your question?

src/radical/pilot/agent/scheduler/base.py
mingtaiha: I would to distinguish between `lrms_info` and `lm_info`. `lrms_info` provides info on the nodes available and the number of CPU/GPU cores per node. Within `lrms_info` is something called `lm_info`. I would like to know more about `lm_info`. The only documentation on this [here](https://github.com/radical-cybertools/radical.pilot/blob/feature/gpu/src/radical/pilot/agent/scheduler/base.py#L129). `lm_info` is assigned to `self._lrms_lm_info` during initialization. This information is passed around in `continuous.py` when slots are defined, but nothing is done to `lm_info`. What is it for?

src/radical/pilot/agent/scheduler/continuous.py
mingtaiha: My bad :(

src/radical/pilot/agent/scheduler/continuous.py
mingtaiha: the function is now called `find_resources`, and I'm ok with that.

src/radical/pilot/agent/scheduler/continuous.py
mingtaiha: Oh right, min(10, 8)/5 = 1. I forgot that the real numbers is not a field in computer-land ;)

src/radical/pilot/agent/scheduler/continuous.py
mingtaiha: Oh I see. I get it now. I was just confused about when the loop would break

src/radical/pilot/agent/scheduler/base.py
mingtaiha: It seems like the different implementations of schedulers (derived from the base class) change slot states differently. The continuous scheduler may use `_change_slot_states` from the base class, scattered scheduler uses the `_change_slot_states` from the scattered scheduler class, the yarn scheduler doesn't use `_change_slot_states` at all.

It might be better to push this function to the derived class because the logic to decide how to assign resources to nodes is in the derived classes anyways. Any thoughts?

src/radical/pilot/agent/scheduler/base.py
mingtaiha: This depends on how we deal with how slots are changed, which is being discussed as well. We should come back to this once we've decided what should be done.

src/radical/pilot/agent/scheduler/base.py
mingtaiha: I don't mind implementing this!

PR #1489, opened 2017-10-31
Feature/gpu into experiments/cybermanufacturing

iparask:


iparask:
Closing without merging as per Andre's instructions

PR #1490, opened 2017-10-31
adding some more comments + 1 question for Andre to non mpi alloc

vivek-bala:
Iterating over giannis' code review with some more comments.

### DIFF
--- a/src/radical/pilot/agent/scheduler/base.py
+++ b/src/radical/pilot/agent/scheduler/base.py
@@ -121,9 +121,15 @@ SCHEDULER_NAME_SPARK        = "SPARK"
 #       'cpu_threads'      : 2,
 #       'gpu_processes     : 2,
 #       'slots' :
-#       {                 # [[nodename, [node_uid], [core indexes],   [gpu idx]]]
-#         'nodes'         : [[node_1,   node_uid_1, [[0, 2], [4, 6]], [[0]    ]],
-#                            [node_2,   node_uid_2, [[1, 3], [5, 7]], [[0]    ]]],
+#       {                 # [[nodename, [node_uid], 
+                          # [[core indexes for process 1],.., [core indexes for process 4]],   
+                          # [[gpu idx for process 1], .., [gpu idx for process 4]]]
+#         'nodes'         : [[node_1,   node_uid_1, 
+                          # [[0, 2], [4, 6]], 
+                          # [[0]]],
+#                           [node_2,   node_uid_2, 
+                          # [[1, 3], [5, 7]], 
+                          # [[0]    ]]],
 #         'cores_per_node': 8,
 #         'gpus_per_node' : 1,
 #         'lm_info'       : { ... }
--- a/src/radical/pilot/agent/scheduler/base.py
+++ b/src/radical/pilot/agent/scheduler/base.py
@@ -116,37 +116,29 @@ SCHEDULER_NAME_SPARK        = "SPARK"
 # for system with 8 cores & 1 gpu per node):
 #
 #     unit = { ...
-#       'cpu_processes'    : 4,
-#       'cpu_process_type' : 'mpi',
-#       'cpu_threads'      : 2,
-#       'gpu_processes     : 2,
+#       'cpu_processes'   : 4,
+#       'cpu_process_type': 'mpi',
+#       'cpu_threads'     : 2,
+#       'gpu_processes    : 2,
 #       'slots' :
-#       {                 # [
-                          #     [   nodename, [node_uid], 
-                          #         [[core indexes for process 1],.., [core indexes for process 4]],   
-                          #         [[gpu index for process 1], .., [gpu index for process 4]]
-                          #     ]
-                          # ]
-#         'nodes'         : [
-                          #     [   node_1,   node_uid_1, 
-                          #         [[0, 2], [4, 6]], 
-                          #         [[0]]
-                          #     ],
-                          #     [   node_2,   node_uid_2, 
-                          #         [[1, 3], [5, 7]], 
-                          #         [[0]]
-                          #     ]
-                          # ]
+#       {                 # [[node,   node_uid,   [cpu idx],        [gpu idx]]]
+#         'nodes'         : [[node_1, node_uid_1, [[0, 2], [4, 6]], [[0]    ]],
+#                            [node_2, node_uid_2, [[1, 3], [5, 7]], [[0]    ]]],
 #         'cores_per_node': 8,
 #         'gpus_per_node' : 1,
 #         'lm_info'       : { ... }
 #       }
 #     }
 #
-# The respective launch method is expected to create processes on the respective
-# given set of cores (nodename_1, cores 0 and 4; nodename_2, cores 1 and 5), and
-# on the respective GPUs.  The other reserved cores are for the application to
-# spawn threads on (`cpu_threads=2`).
+# The `cpu idx` field is a list of sets, where in each set the first core is
+# where an application process is places, while the other cores are reserved for
+# that process' threads.  For GPUs we use the same structure, but GPU processes
+# are currently all considered to be single-threaded.
+#
+# The respective launch method is expected to create processes on the set of
+# cpus and gpus thus specified, (node_1, cores 0 and 4; node_2, cores 1 and 5).
+# The other reserved cores are for the application to spawn threads on
+# (`cpu_threads=2`).
 #
 # A scheduler MAY attach other information to the `slots` structure, with the
 # intent to support the launch methods to enact the placement decition made by
--- a/src/radical/pilot/agent/scheduler/base.py
+++ b/src/radical/pilot/agent/scheduler/base.py
@@ -121,22 +121,29 @@ SCHEDULER_NAME_SPARK        = "SPARK"
 #       'cpu_threads'      : 2,
 #       'gpu_processes     : 2,
 #       'slots' :
-#       {                 # [[nodename, [node_uid], 
-                          # [[core indexes for process 1],.., [core indexes for process 4]],   
-                          # [[gpu idx for process 1], .., [gpu idx for process 4]]]
-#         'nodes'         : [[node_1,   node_uid_1, 
-                          # [[0, 2], [4, 6]], 
-                          # [[0]]],
-#                           [node_2,   node_uid_2, 
-                          # [[1, 3], [5, 7]], 
-                          # [[0]    ]]],
+#       {                 # [
+                          #     [   nodename, [node_uid], 
+                          #         [[core indexes for process 1],.., [core indexes for process 4]],   
+                          #         [[gpu index for process 1], .., [gpu index for process 4]]
+                          #     ]
+                          # ]
+#         'nodes'         : [
+                          #     [   node_1,   node_uid_1, 
+                          #         [[0, 2], [4, 6]], 
+                          #         [[0]]
+                          #     ],
+                          #     [   node_2,   node_uid_2, 
+                          #         [[1, 3], [5, 7]], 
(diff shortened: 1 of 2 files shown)


src/radical/pilot/agent/scheduler/continuous.py
```diff
@@ -182,32 +182,45 @@ def _alloc_node(self, node, requested_cores, requested_gpus,
...
         if partial:
-            # For partial requests the check simpliefies: we just check if we 
-            # have either, some cores *or* gpus, to serve the request
+            # For partial requests, the check simplifies: we need to check if we 
+            # have either some cores *or* gpus to serve the request, if yes we
+            # continue. 
+            # In the following line, we check for its equivalent negative logic:
+            # If both no. of cpus and no. of gpus free on this node do not satisfy the 
+            # request (i.e. there are no free cores and no free gpus), return.
+            # NOTE: This condition needs to checked only for MPI units.
```
andre-merzky: non-mpi CUs will never ask for partial allocations anyway - maybe we should make this clearer in the method docstring?

src/radical/pilot/agent/scheduler/continuous.py
```diff
@@ -182,32 +182,45 @@ def _alloc_node(self, node, requested_cores, requested_gpus,
...
+            # For non-partial requests (ie. full requests): we need to check if 
+            # we can serve both, requested cores *and* gpus, if yes we continue.
+            # In the following line, we check for its equivalent negative logic:
+            # If both no. of cpus and no. of gpus free on this node do not satisfy the 
+            # request (i.e. there are no. of requested cores is more than the 
+            # no. of free cores and requested gpus if more than the no. of free
+            # gpus), return. 
             if  requested_cores > free_cores or \
                 requested_gpus  > free_gpus     :
                 return [], []
```
andre-merzky: I am not disagreeing with the comment, but is that really a useful level of detail?  We have seven lines of comment for, essentially:
```
  if request_a > resource_a or \
     request_b > resource_b:
        return [],[]
```

src/radical/pilot/agent/scheduler/continuous.py
```diff
@@ -235,9 +248,18 @@ def _get_node_maps(self, cores, gpus, threads_per_proc):
         core_map = list()
         gpu_map  = list()
 
+        # Question: The following two lines are checking if the number of
+        # available cores is a multiple of the threads per CPU process and then 
+        # finds the number of CPU processes. Why not input the number of CPU
+        # processes to this function?
         assert(not len(cores) % threads_per_proc)
         n_procs =  len(cores) / threads_per_proc
```
andre-merzky: Assume we have `partial=True`, and ask for `cpu_processes=4, threads_per_process=2`, and we have 5 cores free.  So, we find 5 cores, round down to 4 (` n / chunk * chunk`), and for the line above, we then have:
```
len(cores) == 4
threads_per_proc == 2
n_procs = 4 / 2 == 2
```

How would `cpu_processes` (which is 4) play into it?

src/radical/pilot/agent/scheduler/continuous.py
```diff
@@ -235,9 +248,18 @@ def _get_node_maps(self, cores, gpus, threads_per_proc):
...
 
+
+        # core_map is a list of list of core ids.
+        # core_map = [cores_for_proc_1,cores_for_proc_2,....., cores_for_proc_N]
+        # cores_for_proc_x = [core_for_thread_1, core_for_thread_2,.., 
+        #                       core_for_thread_M]
         idx = 0
```
andre-merzky: This is documented here: https://github.com/radical-cybertools/radical.pilot/blob/feature/gpu/src/radical/pilot/agent/scheduler/base.py#L113
Instead of having two places with that documentation, would you mind adding a reference / link?

src/radical/pilot/agent/scheduler/continuous.py
```diff
@@ -264,31 +288,30 @@ def _alloc_nompi(self, cud):
         Find a suitable set of cores and gpus *within a single node*.
 
         Input:
-        cud: Compute Unit description. Needs to specify at least one process,thread
-        or GPU process.
+        cud: Compute Unit description. Needs to specify at least one CPU 
+        process and one thread per CPU process, or GPU process.
```
andre-merzky: What happens if the `threads_per_process` are set to `0`?  :)

src/radical/pilot/agent/scheduler/continuous.py
```diff
@@ -264,31 +288,30 @@ def _alloc_nompi(self, cud):
...
 
-        # Get the number of processes, threads of GPU processes this unit requests.
-        # For threads, if the entry is not specified in the description, then 
-        # assume it is one thread per process.
+        # Get the number of CPU processes, threads per CPU process 
+        # and, GPU processes this unit requests.
         requested_procs  = cud['cpu_processes']
-        threads_per_proc = cud.get('cpu_threads',1)
+        threads_per_proc = cud['cpu_threads']
         requested_gpus   = cud['gpu_processes']
```
andre-merzky: Haha, this is funny - you change back an edit from Iannis, but do not add back a check which was removed by Iannis, too, which relates to that change.  The consequence of that I'll leave as an exercise to the reader  :-D

src/radical/pilot/agent/scheduler/continuous.py
vivek-bala: Ah yes! I keep thinking of the scheduling as finding available cores for each process as opposed to finding processes for available cores. That makes sense. Thanks.

src/radical/pilot/agent/scheduler/continuous.py
vivek-bala: I did notice the check Giannis removed. But isn't the default number of cpu_threads=1? https://github.com/radical-cybertools/radical.pilot/blob/feature/gpu/src/radical/pilot/compute_unit_description.py#L364

vivek-bala: The check that Giannis removed checks if cpu_threads is None (or 0), both of which the user has to manually set (since the default is 1). At which point he/she deserves an error! :P No?

andre-merzky: Indeed, the default is such.  But I found myself sometimes specifying `0` in the sense of `no additional threads` - which is not what the user should do, but there is also no harm in catching that obvious bad value and replacing it with a sensible `1`, right?

src/radical/pilot/agent/scheduler/continuous.py
andre-merzky: Interesting perspective ...

src/radical/pilot/agent/scheduler/continuous.py
vivek-bala: Makes sense. Done! Added some more detail in the base class.

src/radical/pilot/agent/scheduler/continuous.py
vivek-bala: Sorry can you elaborate on no ```additional``` threads. Once you set cpu_threads=0, the number of cores acquired for that process is 0 (=no. of process * threads per process). Do you mean, you set it to 0 by mistake?

src/radical/pilot/agent/scheduler/continuous.py
vivek-bala: Ok, happy to take the NOTE off.

src/radical/pilot/agent/scheduler/base.py
```diff
@@ -121,9 +121,15 @@
...
-#                            [node_2,   node_uid_2, [[1, 3], [5, 7]], [[0]    ]]],
+#       {                 # [[nodename, [node_uid], 
+                          # [[core indexes for process 1],.., [core indexes for process 4]],   
+                          # [[gpu idx for process 1], .., [gpu idx for process 4]]]
+#         'nodes'         : [[node_1,   node_uid_1, 
+                          # [[0, 2], [4, 6]], 
+                          # [[0]]],
+#                           [node_2,   node_uid_2, 
+                          # [[1, 3], [5, 7]], 
+                          # [[0]    ]]],
```
andre-merzky: Uh, is this really more readable?

src/radical/pilot/agent/scheduler/continuous.py
```diff
@@ -256,10 +256,8 @@ def _get_node_maps(self, cores, gpus, threads_per_proc):
         n_procs =  len(cores) / threads_per_proc
 
 
-        # core_map is a list of list of core ids.
-        # core_map = [cores_for_proc_1,cores_for_proc_2,....., cores_for_proc_N]
-        # cores_for_proc_x = [core_for_thread_1, core_for_thread_2,.., 
-        #                       core_for_thread_M]
+        # core_map is a list of list of core indexes.
+        # See Line 124-126 of ../base.py
```
andre-merzky: Line numbers are a fickle thing.  Maybe 'see top level comment in `./base.py`', or something like that?

src/radical/pilot/agent/scheduler/continuous.py
vivek-bala: I agree it is a bit redundant. The idea was to (i) keep the comment symmetrical with the partial case and (ii) explain that, in order to check if we have adequate resources, we are checking for the negative logic in code (or) whereas in the comment we mention the positive logic (and), but both are equivalent. !(A and B) = !A or !B

src/radical/pilot/agent/scheduler/continuous.py
andre-merzky: Yes, by mistake.  I guess that points to a different issue in the API really, which may benefit from a different discussion.  Look, in the API, we specify `cpu_processes` and `cpu_threads`, but we interpret the latter as `threads_per_process`, which is not quite the same, right?  So, in the API, `0` actually makes semantic sense.

Happy to open up a discussion about the API naming - but I'd rather do that independent from the scheduler discussion, if that's ok.  But do you see how a `0` could well happen to be specified by a user?

src/radical/pilot/agent/scheduler/continuous.py
vivek-bala: My vote is to respond to the user with red lines! :P

I am happy to catch the case when user specifies 0 threads. My suggestion would be the following:
(i) we should document that these are not additional threads but the total number of threads. Increment the number of threads to 1 if user specifies it as 0.
or (ii) change the code base to reflect that these are additional threads and document the same.

src/radical/pilot/agent/scheduler/continuous.py
andre-merzky: I would have the same general comment on the partial case :)    Agree on the neg logic - but then we could change the comment to 
```
In the non-partial case its a no-match if either the cpu or gpu request cannot be served.
```

src/radical/pilot/agent/scheduler/continuous.py
vivek-bala: I think I understand what you meant. Please see my comment in the previous change request of this PR.

src/radical/pilot/agent/scheduler/continuous.py
vivek-bala: Ok, I'll change the comment then :)

src/radical/pilot/agent/scheduler/continuous.py
vivek-bala: Makes sense. Done.

src/radical/pilot/agent/scheduler/base.py
vivek-bala: I iterated on it once more to match the lines. I think it is more legible, but this is subjective. I'll let you make the final call :)

PR #1491, opened 2017-10-31
changed _alloc_node name to something more fitting

itomaldonado:
This was messing with me the first few times I read through this code.


andre-merzky:
thanks!

PR #1492, opened 2017-10-31
Feature/gpu jmoney

jdakka:
commented _get_node_maps


andre-merzky:
thanks Jumana!  I rebased that PR to Ming's branch, as I think that is where you started off, ok?

src/radical/pilot/agent/scheduler/continuous.py
```diff
@@ -236,6 +236,11 @@ def _alloc_node(self, node, requested_cores, requested_gpus,
 
     # --------------------------------------------------------------------------
     #
+
+    '''
+    for each process (containing X threads) generate a process map that requests 1 core per thread.
+    core maps returns a list of lists of process maps
+    '''
```
andre-merzky: Two tidbits: first, docstring comments need to be *inside* the documented method, so that they get picked up by, for example, online doc generators.  Second, please add a sentence about GPUs.  Otherwise that level of detail is very suitable for a short, internal helper method - thanks!

src/radical/pilot/agent/scheduler/continuous.py
```diff
@@ -226,6 +226,11 @@ def _find_resources(self, node, requested_cores, requested_gpus,
     # --------------------------------------------------------------------------
     #
     def _get_node_maps(self, cores, gpus, threads_per_proc):
+        '''
+        For each process (containing X threads) generate a process map that
+        requests 1 core per thread.  Core maps returns a list of lists of
+        process maps.
+        '''
```
andre-merzky: Jumana, I don't think this is very clear, yet.  I mean, it might be better than no description, but does not yet explain what the input parameters are, nor what kind of 'node_map' is meant in the method name...

PR #1508, opened 2017-12-03
move pilot staging code

### LINKED ISSUE #1433: pilot staging interferes with unit staging
Unit staging (`TRANSFER`) on its own works, as does pilot staging (`TRANSFER`).  But when used in the same app, the unit staging hangs: apparently the SAGA level PTY shell is gone, channel recovery kicks in, but in done incompletely, and we hang in prompt detection, as usual.

I did not yet manage to track down why the channel disappears -- or if it really disappears in the first place.  Another option might be that the master channel sharing across different processes is what screws this use case up...

andre-merzky:
this avoids exhaustion of ssh connection handles, but also increases code complexity.  On the upside, it also prepares the move to a deicated data staging component/service.

It fixes #1433

### DIFF
--- a/src/radical/pilot/compute_pilot.py
+++ b/src/radical/pilot/compute_pilot.py
@@ -9,18 +9,12 @@ import copy
 import time
 import threading
 
-import saga          as rs
 import radical.utils as ru
-
-rs.fs = rs.filesystem
 
 from . import utils     as rpu
 from . import states    as rps
 from . import constants as rpc
 from . import types     as rpt
-
-from .staging_directives import expand_staging_directives, complete_url
-from .staging_directives import TRANSFER, COPY, LINK, MOVE, STAGING_AREA
 
 
 # ------------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


andre-merzky:
I think that the Jenkins error is really just a Jenkins update problem - example 7 itself seems to pass now.  @itomaldonado : any idea what's up with that?

itomaldonado:
@andre-merzky yeah, it did, I'll take a look at this. It seems to pass. so I'll give it the OK!

src/radical/pilot/compute_pilot.py
```diff
@@ -547,68 +541,19 @@ def cancel(self):
...
-
-            # TODO: respect flags in directive
-
-            action = sd['action']
-            flags  = sd['flags']
-            did    = sd['uid']
-            src    = sd['source']
-            tgt    = sd['target']
-
-            assert(action in [COPY, LINK, MOVE, TRANSFER])
```
itomaldonado: This is something I wanted to talk to you about, `asserts` are good but we should not depend on them, you can run a python program with the Optimized flags (i.e. `-o` or `-oo`) all of these statements are ignored so we should not necessarily depend too much on them. (Just FYI~.)

https://docs.python.org/2.7/tutorial/modules.html#compiled-python-files

itomaldonado: (I know this line is being removed, which is fine, it is just that there are a lot of `assert` statements all over Pilot).

andre-merzky: yeah, I absolutely agree on that!  I only use assert on things which actually never should be incorrect - they should only get triggered if a protocol or call convention changes in one place but is not updated in another place, or stuff like that. Runtime checks should not be asserted otherwise...

andre-merzky: Let me correct: I _should_ only use asserts like the above :D  I'll do a round of checks to make sure that's the case...

PR #1517, opened 2017-12-11
Issue/1513

vivek-bala:


src/radical/pilot/agent/staging_input/default.py
```diff
@@ -176,29 +177,41 @@ def _handle_unit(self, unit, actionables):
...
                 if tgtdir != sandbox:
-                    # TODO: optimization point: create each dir only once
-                    self._log.debug("mkdir %s", tgtdir)
-                    rpu.rec_makedir(tgtdir)
-
-            if   action == rpc.COPY: shutil.copyfile(src.path, tgt.path)
+                    # create each dir only once
+                    if not os.path.isdir(tgtdir):
+                        self._log.debug("mkdir %s", tgtdir)
+                        rpu.rec_makedir(tgtdir)
```
andre-merzky: `rpu.rec_makeddirs` (calls)[https://github.com/radical-cybertools/radical.pilot/blob/devel/src/radical/pilot/utils/misc.py#L53) `os.makedirs` , the first action of that is to (chec)[https://github.com/python/cpython/blob/master/Lib/os.py#L209] if the dir exists.  In this case the comment was out of sync with the implementation, which shows, again, that incorrect comments are worse than no comments...  Apologies!

src/radical/pilot/agent/staging_input/default.py
```diff
@@ -176,29 +177,41 @@ def _handle_unit(self, unit, actionables):
...
+
+            if action == rpc.COPY: 
+                try:
+                    shutil.copytree(src.path, tgt.path)
+                except OSError as exc: 
+                    # python >2.5
```
andre-merzky: What does that comment mean?

src/radical/pilot/agent/staging_output/default.py
```diff
@@ -262,29 +264,41 @@ def _handle_unit_staging(self, unit, actionables):
             if rpc.CREATE_PARENTS in flags and action != rpc.TRANSFER:
                 tgtdir = os.path.dirname(tgt.path)
                 if tgtdir != sandbox:
-                    # TODO: optimization point: create each dir only once
```
andre-merzky: same as above

vivek-bala:
Thanks for the review. I'll make the suggested changes. Unit and
integration tests are in the works too. You should hear an update about
them today.

- Vivek

vivek-bala:
Please review the following suggested test cases:

| Source | Target | Action | Component |
|-----------|---------|----------|-----------------|
| file_a   | file_b |   copy   |ASIC, ASOC|
| file_a   | file_b |    link    |ASIC, ASOC|
| file_a   | file_b |  move  |ASIC, ASOC|
| file_a   |     -    |   copy   |ASIC, ASOC|
| file_a   |     -    |    link    |ASIC, ASOC|
| file_a   |     -    |  move  |ASIC, ASOC|
| dir_a   | dir_b |   copy   |ASIC, ASOC|
| dir_a   | dir_b |    link    |ASIC, ASOC|
| dir_a   | dir_b |  move  |ASIC, ASOC|
| dir_a   |     -    |   copy   |ASIC, ASOC|
| dir_a   |     -    |    link    |ASIC, ASOC|
| dir_a   |     -    |  move  |ASIC, ASOC|
| file_a   | dir_a/file_b |   copy   |ASIC, ASOC|
| file_a   | dir_a/file_b |    link    |ASIC, ASOC|
| file_a   | dir_a/file_b |  move  |ASIC, ASOC|
| file_a   |    dir_a/    |   copy   |ASIC, ASOC|
| file_a   |    dir_b/    |    link    |ASIC, ASOC|
| file_a   |    dir_c/    |  move  |ASIC, ASOC|

*(-) means the file or folder name is not specified and should be assumed to be the same as the source/target

src/radical/pilot/agent/staging_output/default.py
vivek-bala: Done

vivek-bala:
Document the assumptions made (example, no wildcards in the path).

examples/11_unit_input_folder.py
```diff
@@ -0,0 +1,168 @@
...
+        if not os.path.exists(directory):
+            os.makedirs(directory)
+
+        # create a folder in the file
+        filename = '/tmp/stage_in_folder/input_file.dat'
+        afile = open(filename,'w')
+        afile.write('hello world!')
+        afile.close()
+
+        # TODO: transfer file to remote machine
```
vivek-bala: You mean transfer the folder? 

Also, how does this work currently? do you already have the folder in /tmp/ ?

georgeha: Yes, I only test in localhost right now, so I did not have to move the folder. 
After our discussion, I will create the CU who creates the directory.

itomaldonado:
From the list of tests, test number 16 fails....
```
======================================================================
ERROR: test_copy_single_folder_noname (__main__.TestStagingInputComponent)
----------------------------------------------------------------------
  File "/Users/itomaldonado/git/radical-lab/radical.pilot/venv/lib/python2.7/site-packages/mock/mock.py", line 1305, in patched
    return func(*args, **keywargs)
  File "./tests/sample_component.py", line 258, in test_copy_single_folder_noname
    component._handle_unit(self.unit, actionables)
  File "/Users/itomaldonado/git/radical-lab/radical.pilot/venv/lib/python2.7/site-packages/radical/pilot/agent/staging_input/default.py", line 186, in _handle_unit
    shutil.copytree(src.path, tgt.path)
  File "/usr/local/Cellar/python/2.7.12_2/Frameworks/Python.framework/Versions/2.7/lib/python2.7/shutil.py", line 177, in copytree
    os.makedirs(dst)
  File "/Users/itomaldonado/git/radical-lab/radical.pilot/venv/bin/../lib/python2.7/os.py", line 157, in makedirs
    mkdir(name, mode)
OSError: [Errno 17] File exists: '/Users/itomaldonado/git/radical-lab/radical.pilot/tests/staging-testing-sandbox/rp.session.testing.local.0000/pilot.0000/unit.00000'

----------------------------------------------------------------------
```

itomaldonado:
We have all of the tests implemented and the fix (single line) in place for most tests. Still have a question I'd like to propose to @andre-merzky and @chengwill97 .

There is one test that fails, a symlink of a file to a subdirectory in the unit sandbox *without* providing the target name. (test: `test_link_single_file_to_directory_noname`).

Here is the example: `pilot://single-file --(symlink) --> unit://subdirectory/`
the expected result is: `symlink: unit://subdirectory/single-file` 
actual result: `Error: subdirectory already exists`.

I am actually 'OK' with this error, the assumption is that when you do symlinks you have to give the full path of both the target *and* the link.

But here is the behavior on both UNIX vs Python:

```
# Python
os.symlink('test.c', 'venv/')
OSError: [Errno 17] File exists

# Unix - works
ln -s ./test.c ./venv/
```

andre-merzky:
In this case I would indeed expect  Unix (or rather POSIX) semantics.  If Python does not like that case, we would need to work around it.  Motivation is two-fold: first, SAGA is supposed to follow the POSIX specs where possible (see [here](https://www.ogf.org/documents/GFD.90.pdf) page 219, 'Notes', first entry), and RP file transfer directives are supposed to follow SAGA.  Second, I would like for users which specify link directives to have this just work.

itomaldonado:
I agree, so the only weird behavior was that so I fixed it by checking if:
during a link: src == file, dest == folder then do a link as: src == file, dest == folder/file

examples/make_folders.py
```diff
@@ -0,0 +1,16 @@
+import os
```
andre-merzky: why does this live in examples? Sorry, I don't see where its used?

Issue #1518, opened 2017-12-12
example/data_staging/io_staging_dict.py - got an unexpected keyword argument 'name'


FranklinBetten:
in radical pilot examples/data_staging/

python io_staging_dict.py gives following error 
```
Traceback (most recent call last):
  File "io_staging_dict.py", line 50, in <module>
    session = rp.Session(name=session_name)
TypeError: __init__() got an unexpected keyword argument 'name'
```

FranklinBetten:
after replacing name with uid there is a new error 

Traceback (most recent call last):
  File "io_staging_dict.py", line 102, in <module>
    umgr = rp.UnitManager(session, rp.SCHED_DIRECT_SUBMISSION)
AttributeError: 'module' object has no attribute 'SCHED_DIRECT_SUBMISSION'


To solve this we removed the rp.SHED_DIRECT_SUBMISSION option from the rp.UnitManager call. 

        # Combine the ComputePilot, the ComputeUnits and a scheduler via
        # a UnitManager object.
        #umgr = rp.UnitManager(session, rp.SCHED_DIRECT_SUBMISSION)
        umgr = rp.UnitManager(session)


After that correction there was a new issue that showed up ... 

wait for 1 unit(s)
        O[Callback]: ComputeUnit 'unit.000000' state: UMGR_SCHEDULING_PENDING.
[Callback]: ComputeUnit 'unit.000000' state: UMGR_SCHEDULING.
[Callback]: ComputeUnit 'unit.000000' state: UMGR_STAGING_INPUT_PENDING.
[Callback]: ComputeUnit 'unit.000000' state: UMGR_STAGING_INPUT.
/[Callback]: ComputeUnit 'unit.000000' state: AGENT_STAGING_INPUT_PENDING.
\[Callback]: ComputePilot 'pilot.0000' state: PMGR_ACTIVE_PENDING.
\[Callback]: ComputeUnit 'unit.000000' state: FAILED.
-                                                                     ok
caught Exception: 'ComputeUnit' object has no attribute 'execution_locations'
closing session
[Callback]: ComputePilot 'pilot.0000' state: PMGR_ACTIVE.
close unit manager                                                            ok
close pilot manager                                                            \
wait for 1 pilot(s)
                                                                         timeout
                                                                              ok
session lifetime: 21.4s                                                       ok
Traceback (most recent call last):
  File "io_staging_dict.py", line 123, in <module>
    (unit.uid, unit.execution_locations, unit.state,
[...]

PR #1519, opened 2017-12-12
name attribute has been chagned to uid. Fixing issue #1518

iparask:


andre-merzky:
thanks!

andre-merzky:
Thanks, this can be merged as is.

PR #1522, opened 2017-12-18
expose profiler to LM hooks

andre-merzky:


src/radical/pilot/agent/lm/orte.py
```diff
@@ -180,9 +184,12 @@ def lrms_shutdown_hook(cls, name, cfg, lrms, lm_info, logger):
...
                 if not orterun:
                     raise Exception("Couldn't find orterun")
-                subprocess.Popen([orterun, "--hnp", lm_info['dvm_uri'], "--terminate"])
+                ru.sh_callout('%s --hnp %s --terminate' 
+                             % (orterun, lm_info['dvm_uri']))
+                profiler.prof(event='orte_dvm_stop', uid=cfg['pilot_id'])
             except Exception as e:
-                logger.exception('dmv termination failed')
+                logger.exception('dvm termination failed')
+                profiler.prof(event='orte_dvm_fail', uid=cfg['pilot_id'])
```
iparask: Do you think it is correct to have the same event name between a `ORTE DVM` failure and a shutdown failure?

src/radical/pilot/agent/lm/orte.py
```diff
@@ -180,9 +184,12 @@ def lrms_shutdown_hook(cls, name, cfg, lrms, lm_info, logger):
                 orterun = ru.which('orterun')
                 if not orterun:
                     raise Exception("Couldn't find orterun")
-                subprocess.Popen([orterun, "--hnp", lm_info['dvm_uri'], "--terminate"])
+                ru.sh_callout('%s --hnp %s --terminate'
```
iparask: Is this relevant to this PR?

src/radical/pilot/agent/lm/fork.py
```diff
@@ -26,7 +26,7 @@ def _configure(self):
     # --------------------------------------------------------------------------
     #
     @classmethod
-    def lrms_config_hook(cls, name, cfg, lrms, logger):
+    def lrms_config_hook(cls, name, cfg, lrms, logger, profile):
```
iparask: Here also it is `profile`. I think it is better if it is profiler

src/radical/pilot/agent/lm/orte_lib.py
```diff
@@ -184,10 +187,13 @@ def lrms_shutdown_hook(cls, name, cfg, lrms, lm_info, logger):
                 orte_submit = ru.which('orterun')
                 if not orte_submit:
                     raise Exception("Couldn't find orterun")
-                subprocess.Popen([orte_submit, "--hnp", lm_info['dvm_uri'], "--terminate"])
+                ru.sh_callout('%s --hnp %s --terminate'
```
iparask: Same as above.

src/radical/pilot/agent/lm/orte.py
andre-merzky: Actually yes - since the DVM is running asynchronously, we can't really distinguish a shutdown failure from a runtime failure.  In fact, the most common case where we see the shutdown failure is when the shutdown complains that the DVM is already gone, which is then indeed a runtime failure.  So at the moment this is intentional for the profiler, but the logs will distinguish it.  I added a comment in 64cf2da5 to clarify this.

andre-merzky:
The `profiler` / `profile` parameter mixup is a good catch - this is fixed in f31b60d7.  Thanks!

src/radical/pilot/agent/lm/fork.py
andre-merzky: fixed in 64ba387d - thanks.

src/radical/pilot/agent/lm/orte.py
andre-merzky: I hear you, good question, and I assume a matter of taste.  I tend to try to improve code pieces where there are obvious legacy things missed by previous refactoring and stuff as long as it is as small as here (single line).  If that affects more code and convolutes the PR, it should got into a separate PR indeed.

Happy to jump through the exercise to move it out.  To answer explicitly: no, that change is cosmetic, and not related to the PR otherwise.

src/radical/pilot/agent/lm/orte_lib.py
andre-merzky: same reply :)

andre-merzky:
test this please

andre-merzky:
jenkins test this please

iparask:
You want jenkins to do or me?

andre-merzky:
That was supposed to trigger Jenkins, not you :-)

PR #1527, opened 2018-01-12
fix event docs, add an event for symmetry

andre-merzky:


andre-merzky:
the `[APP]` comment on the comments basically says that.  More verbose documentation will have to be left to a different place (how and where to record those events, for example).

Issue #1535, opened 2018-01-23
Functionality/to open discussion for office hours (friday): Partial ordering of tasks 


jdakka:
This is in reference to the Ensemble of Pipelines pattern in EnTK. In the event that we submit multiple `pipelines` in EnTK, we notice an inconsistency in the execution order and the submission order. EnTK submits tasks in a particular order that is in agreement with the synchronization barriers handled by the `stage(s)` component. However when these ordered tasks are submitted to RP (which translates the `Tasks` to CUs in a `set()`), we lose the initial ordering of tasks. In turn certain pipelines compete for resources and finish well before other pipelines even begin to execute.

jdakka:
@andre-merzky I've created an HTBAC script to test this feature, using `devel` in `RS/RU` and `devel` in `EnTK`.

andre-merzky:
Sooo, does it work? :-)

jdakka:
It works apart from a hanging `rabbitmq`

[rabbit_mq_hangup.tgz.zip](https://github.com/radical-cybertools/radical.pilot/files/1717076/rabbit_mq_hangup.tgz.zip)

andre-merzky:
So,  if that behaves as expected, and the remaining problems are confined to  RabbitMQ, ie. to the EnTK layer, can the fifo scheduler then be merged, or is there additional work needed (in RP) to serve this use case?

jdakka:
@vivek-bala @andre-merzky I indeed verify that I can run two concurrent EnTK pipelines and the performance is as expected using feature/FIFO

iparask:
This is resolved!

PR #1537, opened 2018-01-23
Feature/fifo

andre-merzky:
attempt to address #1535 .  Note that this implementation is really just a stopgap solution.

This should probably have been branched off devel instead off `feature/hombre`...  Either way, reviewers please focus on `src/radical/pilot/agent/scheduler/continuous_fifo.py`


vivek-bala:
Quick question: Only the changes in continuous_fifo and base are relevant to the feature/fifo, right? I think the rest (hombre related changes) are from a merge.

andre-merzky:
That's correct.

src/radical/pilot/agent/scheduler/continuous_fifo.py
```diff
@@ -0,0 +1,51 @@
...
+            # nope - we let this attempt fail
+            self._log.debug('defer unit %s [%s]', uid, self._index)
+            return False
+
+        # yep: lets try for real
+        ret = Continuous.try_allocatio(self, cu)
+
+        if ret: 
+            # it worked!  Keep the new index
+            self.last = idx
```
vivek-bala: I think line 45 needs to be ```self._index =  idx```.

Although, in terms of naming, I am in favor of 'last' instead of 'index`.

andre-merzky: ack on both.  To be clear: I did not run this code, this is just for discussion right now, to check if this is a viable approach and actually solves the problem (short-term).

src/radical/pilot/agent/scheduler/continuous_fifo.py
```diff
@@ -0,0 +1,51 @@
...
+# IDs to be of the format `unit.%d'`.
+#
+class ContinuousFifo(Continuous):
+
+    # --------------------------------------------------------------------------
+    #
+    def __init__(self, cfg, session):
+
+        self.slots  = None
+        self._index = 0
```
vivek-bala: Functionally, starting with 0 doesn't look wrong. I think the default value for ```self._index``` should be -1 or None (basically not 0). I think it is easier to understand because when we start from 0, self._index remains 0 even after unit.000000 has been allocated. Would you agree?

src/radical/pilot/agent/scheduler/continuous_fifo.py
vivek-bala: :+1:

src/radical/pilot/agent/scheduler/continuous_fifo.py
vivek-bala: I think this became outdated but still applies for ```self._last = 0```.

andre-merzky: Ah, an off-by-one error!  You are correct :-)

src/radical/pilot/agent/scheduler/continuous_fifo.py
```diff
@@ -0,0 +1,51 @@
...
+
+    # --------------------------------------------------------------------------
+    #
+    def _try_allocation(self, cu):
+
+        # check if the unit is up for scheduling
+        uid = cu['uid']
+        idx = int(uid.split('.')[1])
+
+        if idx > self._last + 1:
```
vivek-bala: Question: Given that all units and their units are created in sequence (i.e. unit.000003 is created only after unit.000002), can we change this condition to :

```
if idx == self._last + 1: 
    # yep: lets try for real
    ret = Continuous.try_allocatio(self, cu)

    if ret: 
        # it worked!  Keep the new index
        self._last = idx

    return ret

else:
    # nope - we let this attempt fail
    self._log.debug('defer unit %s [%s]', uid, self._last)
    return False
```

This question/suggestion is only for ease of understanding of code (if my assumption on sequence of uids is always correct).

andre-merzky: The only difference is the inversion of the `if` test - or am I missing something?  I find both equally easy to read - the whole method is just 10 lines after all, so no preference.  In longer methods I always prefer to have the shorter `if`-branch first, to keep the `else` clause visually close.  And I usually also try to bail-out (return) from a method as early as possible, as soon as that information exists - that avoids too many `else`-clauses.  But, as said, no preference in this case - feel free to adjust either way.

vivek-bala: I agree it is small. I'll add it in a comment then. My concern is not really about the if and else order, it was more about testing '==' instead of '>'. I think using '!=' in the if condition both our suggestions. Let me modify that then.

andre-merzky: Ah, I see - but `!=` would in principle also test for smaller indexes, which by definition cannot happen, right?  Wouldn't that increase confusion for readers? :-)

vivek-bala: That's what I added in the comment above the if condition :)

The '>' condition seems redundant if you know that uids are in order of unit creation. The '!=' seems incomplete (since it might test for small indexes) if you don't know that the uids are in order of unit creation. My preference is for the latter with a note :)

andre-merzky:
@vivek-bala @jdakka : ping

vivek-bala:
Hey Andre, made a small fix on the method name and added a test script. Please see the assertion I make in the test script, I make an assumption there which I commented.  

Jumana is testing this scheduler for her use case. @jdakka please pull this fix.

jdakka:
I ran a real simulation using Blue Waters and NAMD with two concurrent pipelines. I have some data from a run before Vivek made the changes. 
[eop_fifo_queue.tgz.zip](https://github.com/radical-cybertools/radical.pilot/files/1712924/eop_fifo_queue.tgz.zip)

`/u/sciteam/dakka/scratch/radical.pilot.sandbox/rp.session.two.jdakka.017571.0001/pilot.0000` 

The units are spawned and executed in the order I expected. The pilot terminated properly. There were some errors in the enqueue threads that can be found in `debug.log`

I am running another test with Vivek's changes.

vivek-bala:
There was a typo in the continuous_fifo script, you should have got an error if you used it. You have to set the "agent_scheduler" in the resource config file to "CONTINUOUS_FIFO" to use it. For example, if you ran it on localhost, you have to change this line: https://github.com/radical-cybertools/radical.pilot/blob/feature/fifo/src/radical/pilot/configs/resource_local.json#L19

PR #1529, opened 2018-01-16
switch to new OMPI deployment

andre-merzky:
the original OMPI deployment got purged - this is the replacement.

PR #1539, opened 2018-01-25
Fix/issue 1538

andre-merzky:
This requires a simultaneous merge in RS.  The patch allows to share the RP session ID with the inherited SAGA session, which in turn avoids that the RP agent  creates its own session ID, which in turn would require `$HOME/.radical` for  state management, which is not available on some clusters (titan).  Sharing the ID for the inherited session is semantically the right thing to do anyway.  I'll open a ticket to clean up session ID management, which should not be shared between the layers anyway.

### DIFF
--- a/src/radical/pilot/agent/agent_0.py
+++ b/src/radical/pilot/agent/agent_0.py
@@ -55,7 +55,7 @@ class Agent_0(rpu.Worker):
 
         self._uid         = agent_name
         self._pid         = cfg['pilot_id']
-        self._session_id  = cfg['session_id']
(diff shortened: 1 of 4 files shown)


andre-merzky:
This branch has already merged #1529 as that was needed for testing.  That PR is still waiting for review and merge - but only affects config files.

itomaldonado:
This PR missed the environment variable `RADICAL_BASE_DIR` which is needed by the utils PR: https://github.com/radical-cybertools/radical.utils/pull/131

PR #1540, opened 2018-01-25
Feature/config docs

andre-merzky:
This patch adds some basic documentation on agent configuration.  Its by no means complete, but better than nothing, hey? :)


andre-merzky:
@shantenujha : ping

Issue #1466, opened 2017-09-27
AMBER on Comet fails inside RP


SrinivasMushnoori:
The pilot is submitted (apparently) correctly, the executable is called (apparently) correctly, but then the CU fails because it is unable to locate some libraries. The simulations runs just fine standalone, i.e. outside of RP.

```
/opt/amber/bin/pmemd.MPI: error while loading shared libraries: libnetcdff.so.6: cannot open shared object file: No such file or directory
/opt/amber/bin/pmemd.MPI: error while loading shared libraries: libnetcdff.so.6: cannot open shared object file: No such file or directory
/opt/amber/bin/pmemd.MPI: error while loading shared libraries: libnetcdff.so.6: cannot open shared object file: No such file or directory
/opt/amber/bin/pmemd.MPI: error while loading shared libraries: libnetcdff.so.6: cannot open shared object file: No such file or directory
/opt/amber/bin/pmemd.MPI: error while loading shared libraries: libnetcdff.so.6: cannot open shared object file: No such file or directory
/opt/amber/bin/pmemd.MPI: error while loading shared libraries: libnetcdff.so.6: cannot open shared object file: No such file or directory
/opt/amber/bin/pmemd.MPI: error while loading shared libraries: libnetcdff.so.6: cannot open shared object file: No such file or directory
/opt/amber/bin/pmemd.MPI: error while loading shared libraries: libnetcdff.so.6: cannot open shared object file: No such file or directory
```

Unsure how to solve this. CU Description is as follows:

            # -------- BEGIN USER DEFINED CU DESCRIPTION --------- #

            cudesc = rp.ComputeUnitDescription()
            cudesc.environment = {'CU_NO': i}
            cudesc.executable  = "/opt/amber/bin/pmemd.MPI"
            cudesc.pre_exec = ['module load amber']
            cudesc.arguments   = ['-O',  '-i', 'mdin', '-o', 'mdout', '-p', 'prmtop', '-c', 'inpcrd','-inf', 'info.mdinfo']
            cudesc.input_staging = ['mdin','prmtop','inpcrd']
            #cudesc.output_staging = ['outfile']
            cudesc.cores = CU_CORES
            cudesc.mpi = True
            # -------- END USER DEFINED CU DESCRIPTION --------- #

andre-merzky:
[...]

PR #1544, opened 2018-01-25
Fix/issue 1466

andre-merzky:
This fixes env settings for the various startup methods.  The work was prompted by #1466 and by notorious env problems in ORTE/ORTE_LIB, but now includes similar fixes for all launch methods to keep their structure and semantics synced,


andre-merzky:
@iparask : ping

PR #1547, opened 2018-01-30
make use of flags consistent between RP and RS

andre-merzky:
This needs testing...

### DIFF
--- a/src/radical/pilot/agent/staging_input/default.py
+++ b/src/radical/pilot/agent/staging_input/default.py
@@ -173,7 +173,7 @@ class Default(AgentStagingInputComponent):
 
             # SAGA will take care of dir creation - but we do it manually
             # for local ops (copy, link, move)
-            if rpc.CREATE_PARENTS in flags and action != rpc.TRANSFER:
+            if flags & rpc.CREATE_PARENTS and action != rpc.TRANSFER:
                 tgtdir = os.path.dirname(tgt.path)
                 if tgtdir != sandbox:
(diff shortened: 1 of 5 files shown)


andre-merzky:
@vivek-bala : ping

vivek-bala:
Thanks for the reminder Andre! Testing it now. 

Questions:
1) Should 'RECURSIVE' flag from rs.filesystem also be included to rp? Also I believe it can be a default flag in addition to CREATE_PARENTS?
2) I am not entirely sure whether SKIP_FAILED should be the default. The user would expect the task to fail if the data staging failed (prior to execution). When the task fails from 'EXECUTING' state, we expect the executing kernel to have failed due to some computation or resource issues, not due to missing data (which did not get staged in). Would you agree?

vivek-bala:
Tested agent and unit staging. Both working.

andre-merzky:
I'd leave the `recursive` inclusion to one of the RCT class PRs - at the moment its not supported.  But yes, that would be done along the same lines.

The `SKIP_FAILED` is not implemented, AFAIK, and I actually don't know in what use cases that originated.  I would be in favor of removing it to be honest.  But this PR just tries to fix the definitions...  I'll open a ticket for `SKIP_FAILED`.

andre-merzky:
Ticket is #1557

vivek-bala:
Okay. I'll respond on #1557 .

andre-merzky:
Thanks Vivek!

PR #1555, opened 2018-02-05
Fix/issue 1554

andre-merzky:
This fix ensures that shell commands (like `untar`) are not submitted to the batch queue, but instead run interactively on the login nodes.

This is urgent to merge.


andre-merzky:
Thanks for the quick reply.  A hotfix is not needed, as the release is not affected.  The issue crept in with a different merge to devel.

Issue #1512, opened 2017-12-05
Agent Input Staging Proposal


iparask:
Right now each actionable, e.i. each staging directive, is treated independently from any other. Although, it is usual practice to move multiple data from one unit to another. Treating each data movement as something separate instruction, thus doing multiple file movements at the same time resulting to multiple hits on the FS resulting sometimes in varying latencies.

I propose that an on-the-fly analysis on actionables. When multiple action require to move data from the same source to the same target, it can happen as a single instruction. Also to avoid the case where multiple small files are moved around, we can see whether or not creating an uncompressed tarball before any move or copy would make sense.

iparask:
In addition, we can introduce magic characters in the source and target of the staging directive, like `*`, allowing the user to define staging of multiple files with some level of similarity to define only one staging directive.

andre-merzky:
See #1513 as related issue.

Its probably not trivial to analyze staging directives and to bulk them on the fly - but I agree with the premise that many hits on a shared FS are costly.  I would propose to first measure the effect (out of band, in a shell script or so) on a known slow shared FS (like BlueWaters).

For the magic expansion, I would strongly suggest to stick with the shell conventions, and if possible to reuse Python's `glob` module.

iparask:
Is `*` in file names a shell convention?

andre-merzky:
> Is * in file names a shell convention?

Better: its defined by [POSIX](http://pubs.opengroup.org/onlinepubs/009695399/utilities/xcu_chap02.html#tag_02_13)

itomaldonado:
Also, in Python we can use:

https://docs.python.org/2/library/glob.html#glob.glob

andre-merzky:
Iannis, the glob part is somewhat orthogonal to the bulk transfer (via tarballs), so I'd say you should leave it out for the moment, and only focus on one of them.  I'll leave it to you both to judge if the on-the-fly analysis is useful in this context, or if you want to simply always tar on a certain number of staging directives per unit.  Or if you want to optimize across units.

Since this is essentially an optimization, a big part of your initial work will be to *measure* (not guess) what you want to optimize, and if the chosen approach is promising to improve performance.  Please do take that serious: I am not very open to see code complexity increase without documented benefits.

The ticket talks about the agent staging component and optimization of shared FS access.  Please consider (ie. measure) if that optimization would not be more useful on the remote transfer.  But I'll leave this to you to pick either the UMGR or the Agent components as implementation target.

[...]

PR #1561, opened 2018-02-20
Fix/issue 1512

iparask:
This pull request address issue #1512 based on the discussion on that ticket.


examples/05_unit_input_data.py
```diff
@@ -35,7 +35,7 @@
 
     # Create a new session. No need to try/except this: if session creation
     # fails, there is not much we can do anyways...
-    session = rp.Session()
+    session = rp.Session(uid='TarTest')
```
andre-merzky: Please don't hardcode session names in the examples - this will fail on the second execution.

examples/05_unit_input_data.py
```diff
@@ -92,12 +94,15 @@
...
           # this is a shortcut for:
-          # cud.input_staging  = {'source': 'client:///input.dat', 
-          #                       'target': 'unit:///input.dat',
-          #                       'action': rp.Transfer}
+            cud.input_staging  = [{'source': 'client:///input.dat', 
+                                   'target': 'unit:///input.dat',
+                                   'action': rp.TARBALL},
+                                  {'source': 'client:///input2.dat', 
+                                   'target': 'unit:///input2.dat',
+                                   'action': rp.TARBALL}]
```
andre-merzky: Wouldn't it make sense to add an additional example?  I would think we should keep the original example on the canonical staging (otherwise we'll need to explain what  TARBALL means...)

src/radical/pilot/agent/staging_input/default.py
```diff
@@ -103,7 +104,7 @@ def _handle_unit(self, unit, actionables):
         # NOTE: see documentation of cu['sandbox'] semantics in the ComputeUnit
         #       class definition.
         sandbox = unit['unit_sandbox']
-
+
```
andre-merzky: why?  :-)

src/radical/pilot/agent/staging_input/default.py
```diff
@@ -199,7 +200,13 @@ def _handle_unit(self, unit, actionables):
                     self._log.error('no transfer for %s -> %s', src, tgt)
                     self._prof.prof('staging_in_fail', uid=uid, msg=did)
                     raise NotImplementedError('unsupported transfer %s' % src)
-
+            elif action == rpc.TARBALL:
+                # If somethig was staged via the tarball method, the tarball is
+                # extracted and then removed from the unit folder.
+                tar = tarfile.open(tgtdir+'/'+uid+'.tar')
```
andre-merzky: please use spaces around `+` (PEP.8).

andre-merzky: Per the `if` above, you can get `client://` type URLs here - do those need expansion?

src/radical/pilot/agent/staging_input/default.py
```diff
@@ -199,7 +200,13 @@ def _handle_unit(self, unit, actionables):
...
+                tar.extractall(path=tgtdir)
+                tar.close()
+                os.remove(tgtdir+'/'+uid+'.tar')
```
andre-merzky: You may want to keep the tarballs if `RADICAL_DEBUG` is set?

src/radical/pilot/umgr/staging_input/default.py
```diff
@@ -325,26 +325,58 @@ def _handle_unit(self, unit, actionables):
...
+                # and append the directive to the removables.
 
-            self._prof.prof('staging_in_start', uid=uid, msg=did)
+                if not os.path.isfile(uid+'.tar'):
+                    tar_file = tarfile.open(uid+'.tar','w')
+                tar_file.add(src.split('client:///')[1],arcname=tgt.split('unit:///')[1])
+                removables.append(sd)
+                sid = sd.get('uid')
+            elif action == rpc.TRANSFER:
+                self._prof.prof('staging_in_start', uid=uid, msg=did)
```
andre-merzky: You moved the profile entry into the `elif` - now the event is  missing for the tarballs.

src/radical/pilot/umgr/staging_input/default.py
```diff
@@ -325,26 +325,58 @@ def _handle_unit(self, unit, actionables):
...
             did    = sd['uid']
             src    = sd['source']
             tgt    = sd['target']
+            if action == rpc.TARBALL:
+                # If the staging method is tarball, check initially if the tar 
+                # file exists. If not create it. Add each file in the tarball
+                # and append the directive to the removables.
 
-            self._prof.prof('staging_in_start', uid=uid, msg=did)
+                if not os.path.isfile(uid+'.tar'):
```
andre-merzky: please use spaces around `+`

src/radical/pilot/umgr/staging_input/default.py
```diff
@@ -325,26 +325,58 @@ def _handle_unit(self, unit, actionables):
...
+                    tar_file = tarfile.open(uid+'.tar','w')
+                tar_file.add(src.split('client:///')[1],arcname=tgt.split('unit:///')[1])
```
andre-merzky: Are you *sure* that the src and tgt specs match that pattern here?  Maybe use URL parsing and then `url.path`?  This differs from how the `elif` branch handles the URLs.

src/radical/pilot/umgr/staging_input/default.py
```diff
@@ -325,26 +325,58 @@ def _handle_unit(self, unit, actionables):
...
+                removables.append(sd)
+                sid = sd.get('uid')
```
andre-merzky: this is the same as `did` above?

src/radical/pilot/umgr/staging_input/default.py
```diff
@@ -325,26 +325,58 @@ def _handle_unit(self, unit, actionables):
...
 
-            if rpc.CREATE_PARENTS in flags:
-                copy_flags = rs.filesystem.CREATE_PARENTS
-            else:
-                copy_flags = 0
+                saga_dir.copy(src, tgt, flags=flags)
+                self._prof.prof('staging_in_stop', uid=uid, msg=did)
+
+
+        if os.path.isfile(uid+'.tar'):
```
andre-merzky: Wouldn't it make sense to check if the `tar_file` object exists instead of hitting the file system?

src/radical/pilot/umgr/staging_input/default.py
```diff
@@ -325,26 +325,58 @@ def _handle_unit(self, unit, actionables):
...
+            # if a tarball exists close it and do the transfer. After it is done
+            # delete the tarball. Remove any directive that was use to create the
+            # tarball and add a single one that communicates to the agent that
+            # this type of transfer was executed.
+            tar_file.close()
+
+            self._prof.prof('staging_in_start', uid=uid, msg=did)
```
andre-merzky: Ah, here it is.  So the transfer ID is then the did of the last staging directive.  Hmm, Lets discuss this, I am not sure what analytics will think if some SDs do not have any profile events at all.  The files get transferred after all...

src/radical/pilot/umgr/staging_input/default.py
```diff
@@ -325,26 +325,58 @@ def _handle_unit(self, unit, actionables):
...
+
+            src = complete_url(uid+'.tar', src_context, self._log)
+            tgt = complete_url(uid+'.tar', tgt_context, self._log)
 
-            saga_dir.copy(src, tgt, flags=copy_flags)
+            saga_dir.copy(src, tgt, flags=rs.filesystem.CREATE_PARENTS)
```
andre-merzky: You hardcode that flag here - that is different from the TRANSFER case.

src/radical/pilot/umgr/staging_input/default.py
```diff
@@ -325,26 +325,58 @@ def _handle_unit(self, unit, actionables):
...
             self._prof.prof('staging_in_stop', uid=uid, msg=did)
+            for sd in removables:
+                unit['description']['input_staging'].remove(sd)
+            unit['description']['input_staging'].append({'uid': sid, 
+                                                         'priority': 0, 
+                                                         'source': 'client:///'+uid+'.tar', 
+                                                         'flags': ['CreateParents', 'SkipFailed'],
```
andre-merzky: please use `DEFAULT_FLAGS`

src/radical/pilot/umgr/staging_input/default.py
```diff
@@ -325,26 +325,58 @@ def _handle_unit(self, unit, actionables):
...
-            saga_dir.copy(src, tgt, flags=copy_flags)
+            saga_dir.copy(src, tgt, flags=rs.filesystem.CREATE_PARENTS)
             self._prof.prof('staging_in_stop', uid=uid, msg=did)
+            for sd in removables:
+                unit['description']['input_staging'].remove(sd)
+            unit['description']['input_staging'].append({'uid': sid, 
+                                                         'priority': 0, 
+                                                         'source': 'client:///'+uid+'.tar', 
+                                                         'flags': ['CreateParents', 'SkipFailed'], 
+                                                         'action': 'Tarball',
```
andre-merzky: please use the constant.

tests/issue_1512/run_test.sh
```diff
@@ -0,0 +1,44 @@
+#!/bin/bash
+
+for jj in {1,2,4,8,16,32,64,128,256,512,1024,2048}; do
```
andre-merzky: Why the array syntax?  This works just as well and is simpler:
```
$ for i in 1 2 3; do echo $i; done
1
2
3
```

Doesn't matter really - but in general you may want to use POSIX where bash specific features are not needed, to keep scripts portable.

tests/issue_1512/TarCpUntar.py
```diff
@@ -0,0 +1,27 @@
+import os
```
andre-merzky: Please apply the coding guidelines also to tests :-)
More importantly,  how is that code related to testing the #1512?   It looks like you test the tarfile module - that is not our business...

tests/issue_1512/randomFiles.py
```diff
@@ -0,0 +1,28 @@
+import string
```
andre-merzky: FYI, this is a much quicker and simpler way to create files of a certain size in a shell script:
```
$ time head -c $((1024*1024)) /dev/urandom > /tmp/t.dat
r:0m0.012s  u:0m0.001s  s:0m0.011s
```

tests/issue_1512/run_test.sh
```diff
@@ -0,0 +1,44 @@
...
+    python randomFiles.py $jj
+    echo '----------Number of files: '$jj'----------'
+    echo '---------- Half Meg ----------'
+    echo 'Time'>run_half_$jj.csv
+    for ii in {1..10}; do
+            python transfer_files.py halfMeg wrangler.tacc.utexas.edu <path> >>run_half_$jj.csv 2>>log.log
```
andre-merzky: I am missing something - what is `<path>` here?  `> >>` looks strange to me, too.  Does this run?

tests/issue_1512/transfer_files.py
```diff
@@ -0,0 +1,73 @@
+import os
```
andre-merzky: please apply coding guidelines.

tests/issue_1512/transfer_files.py
```diff
@@ -0,0 +1,73 @@
...
+if __name__ == '__main__':
+
+    parser = argparse.ArgumentParser()
+    parser.add_argument('srcfolder',help='Folder which contains the files to be transfered')
+    parser.add_argument('resource', help="Target Resource that support GSISFTP")
+    parser.add_argument('tgtfolder', help="Target folder. Should be the whole path.")
+    parser.add_argument('--tar',help='Transfer without tarball or with',default=False,action="store_true")
+    args = parser.parse_args()
+    
+    TansferFiles = glob.glob(args.srcfolder+'/*')
```
andre-merzky: typo (T*r*ansfer)

src/radical/pilot/umgr/staging_input/default.py
iparask: Yes and also I need to create a new id for this staging

src/radical/pilot/agent/staging_input/default.py
iparask: I think I had something there and I deleted it.

examples/05_unit_input_data.py
iparask: Done

examples/05_unit_input_data.py
iparask: Done

src/radical/pilot/agent/staging_input/default.py
iparask: done

src/radical/pilot/agent/staging_input/default.py
iparask: done

iparask: I do not think there is any need to check about `client:///`. The `if` in line 161 makes sure that whatever passes does not have `client://` at the target. Am I right?

I have a feeling that either I am missing something major,

src/radical/pilot/agent/staging_input/default.py
iparask: Are the files the user transferred enough? I agree it needs a debug message to show what is happening.

iparask:
Hello @andre-merzky , I am not sure how to trigger a new review request. I believe I addressed most of your comments. Also Srinivas has pushed the unit tests. Can you please take another look?

Thank you

examples/05_unit_input_data.py
```diff
@@ -92,12 +92,12 @@
...
-            cud.input_staging  = ['input.dat']
+            #cud.input_staging  = ['input.dat']
             
           # this is a shortcut for:
-          # cud.input_staging  = {'source': 'client:///input.dat', 
-          #                       'target': 'unit:///input.dat',
-          #                       'action': rp.Transfer}
+            cud.input_staging  = [{'source': 'client:///input.dat', 
+                                   'target': 'unit:///input.dat',
+                                   'action': rp.TRANSFER}]
```
andre-merzky: Wait, why is this a list now?  I am missing something...

src/radical/pilot/umgr/staging_input/default.py
```diff
@@ -325,26 +324,61 @@ def _handle_unit(self, unit, actionables):
...
+                sid = ru.generate_id('sd')
+                self._prof.prof('tarball_create_stop', uid=uid, msg=did)
+            elif action == rpc.TRANSFER:
+                self._prof.prof('staging_in_start', uid=uid, msg=did)
+
+                src = complete_url(src, src_context, self._log)
+                tgt = complete_url(tgt, tgt_context, self._log)
+
+                saga_dir.copy(src, tgt, flags=flags)
+                self._prof.prof('staging_in_stop', uid=uid, msg=did)
```
andre-merzky: This event is duplicated for this uid/did now.

andre-merzky:
Please run at least one remote test on XSEDE.

andre-merzky:
Guys, help me out - how did `tests/test_tarball_umgr.py` actually work when the `saga.fs.Directory.copy` call was `mock`'ed away?  I don't get it...

andre-merzky:
Ah, I do get it now (1 min after last comment :-) : you mask the ENOTDIR exception which happens when passing a URL to the shutil copy...  Man, this took me ages...

Leaving SAGA in place to do the actual thing works.  I would argue that one should only mock those parts which interfere with running the test, really.  Otherwise we don't really test much, do we?  @itomaldonado , any opinion?

itomaldonado:
@andre-merzky I wanted to comment on your last point. I suggested them not testing the saga layer because this test specifically isn't looking at how saga does a transfer. From this module's perspective one the saga function returns successful, the assumption is that the file made it to its destination. We can remove the dependency of Saga (and having to set-up a transfer, ssh setups etc.) by mocking the function call. You can also mock saga to return errors (i.e. raise exceptions) and have unit tests covering the scenarios when that function call fails, but we don't really care about actually testing the transfer because that is not what this test is about.

the assumption is that the saga function was tested separately.

Hopefully that makes sense?

PR #1562, opened 2018-02-20
Fix/issue 1514

vivek-bala:


src/radical/pilot/umgr/staging_input/default.py
```diff
@@ -338,12 +340,23 @@ def _handle_unit(self, unit, actionables):
...
+            # Check if create parents flag is set,
+            # if not set it.
+            if rs.filesystem.CREATE_PARENTS not in flags:
+                flags += [rs.filesystem.CREATE_PARENTS]
+
+            if flags:
+                staging_flags = reduce(lambda x, y: x | y, flags)
+            else:
+                staging_flags = 0
+
```
georgeha: Instead of reducing why not or the flags directly?

flags = 0
```
if rs.filesystem.RECURSIVE not in flags:
                     flags  = flags | rs.filesystem.RECURSIVE

if rs.filesystem.CREATE_PARENTS not in flags:
                 flags  = flags | rs.filesystem.CREATE_PARENTS
```

vivek-bala: True. But the user specifies the flags in the form of a list and not an integer. It looks like:
```
cud.input_staging = {'source': xyz, 'action': rp.TRANSFER, 'target': xyz, 'flags': [a,b,c]}
```
So we have to iterate through the list, performing what you suggest or use the reduce function at the end. Would you agree?

georgeha: my suggestion is to use instead of : 
```
flags += [rs.filesystem.RECURSIVE]

flags = flags |  rs.filesystem.RECURSIVE
```

So we will not have to reduce it at the end. Am I missing something?

vivek-bala: I don't think I understand what you are suggesting. 

``flags`` is a list. You are performing an OR with an int when you do ```flags = flags |  rs.filesystem.RECURSIVE``` so that will give you an error.

Second you are adding recursive flag to the list and then doing OR with it again.

I see two methods to do it:
```
>>> import saga.filesystem as rsf
>>> flags = [rsf.CREATE_PARENTS]
>>> flags += [rsf.RECURSIVE]
>>> flags
[64, 2]
>>> reduce(lambda x, y: x | y, flags)
66
```
or
```
>>> import saga.filesystem as rsf
>>> flags = [rsf.CREATE_PARENTS]
>>> flags += [rsf.RECURSIVE]
>>> flags
[64, 2]
>>> staging_flag = 0
>>> for f in flags:
...     staging_flag = staging_flag + f
... 
>>> staging_flag
66
```

Does this help? Is what you are suggesting similar to the second method here? If it is different maybe try the code out and paste the snippet here. That might help me understand.

georgeha:
```
In [11]: import saga.filesystem as rsf

In [12]: staging_flag = 0

In [13]: staging_flag = staging_flag | rsf.CREATE_PARENTS

In [14]: staging_flag = staging_flag | rsf.RECURSIVE

In [15]: staging_flag
Out[15]: 66
```
that's it. If not flags are specified staging_flag will be 0. Let me know if it makes sense now

vivek-bala: What about other flags that the user might pass? (which we obtain in the form of a list)
see https://github.com/radical-cybertools/radical.pilot/blob/fb39871dc1da9bdd0c9896381f6c50190fb7b28d/src/radical/pilot/umgr/staging_input/default.py#L333

georgeha: I see, then we can't avoid the reduce.

src/radical/pilot/umgr/staging_input/default.py
```diff
@@ -338,12 +340,23 @@ def _handle_unit(self, unit, actionables):
...
 
-            if rpc.CREATE_PARENTS in flags:
-                copy_flags = rs.filesystem.CREATE_PARENTS
-            else:
-                copy_flags = 0
+            # Check if the src is a folder, if true
+            # add recursive flag if not already specified
+            if os.path.isdir(src.path):
+                if rs.filesystem.RECURSIVE not in flags:
+                    flags += [rs.filesystem.RECURSIVE]
```
andre-merzky: Why do we need the `RECURSIVE` flag if its always added as needed anyway?

src/radical/pilot/umgr/staging_input/default.py
```diff
@@ -338,12 +340,23 @@ def _handle_unit(self, unit, actionables):
...
 
-            saga_dir.copy(src, tgt, flags=copy_flags)
+            # Check if create parents flag is set,
+            # if not set it.
+            if rs.filesystem.CREATE_PARENTS not in flags:
+                flags += [rs.filesystem.CREATE_PARENTS]
```
andre-merzky: Why do we need the `CREATE_PARENTS` flag in RP if its always added anyway?

src/radical/pilot/umgr/staging_input/default.py
andre-merzky: I am with George here though: If we use bitmask flags, then that should also apply to the API.  We can make this change backward compatible with:

```
if isinstance(sd['flags'], list):
    flags = sd['flags']
    sd['flags'] = 0
    for flag in flags: sd['flags'] |= flag
```

This can be applied once when we expand the staging directive, and then all other code paths can rely on flags being a bitmask.

This does not need to be part of the PR though.

src/radical/pilot/constants.py
```diff
@@ -60,7 +61,7 @@
 #
 # Flags
 #
-CREATE_PARENTS = 'CreateParents'  # Create parent directories while writing file
+CREATE_PARENTS = rsf.CREATE_PARENTS  # Create parent directories while writing file
 SKIP_FAILED    = 'SkipFailed'     # Don't stage out files if tasks failed
```
andre-merzky: either all or none of the flags should be strings.

src/radical/pilot/umgr/staging_input/default.py
```diff
@@ -338,12 +340,23 @@ def _handle_unit(self, unit, actionables):
...
+
+            if flags:
```
andre-merzky: You just made sure that flags is never empty.  You can also remove the special case of an empty list with:
```
>>> flags = list()
>>> reduce(lambda x, y: x | y, flags, 0)
0
```

tests/test_issue_1514/config.json
```diff
@@ -0,0 +1,326 @@
+
```
andre-merzky: Hmm, this seems like a lot of duplication - we probably don't want to maintain a new config file for each issue...

tests/test_issue_1514/test_rp_integration.py
```diff
@@ -0,0 +1,201 @@
...
+  
+    # Extract info from RP config file
+    #-------------------------------------------------------------------------------------------------------------------
+    config_loc = '../../src/radical/pilot/configs/resource_%s.json'%resource.split('.')[0]
+    path_to_rp_config_file = os.path.realpath(os.path.join(os.getcwd(),config_loc))
+    cfg_file = ru.read_json(path_to_rp_config_file)[resource.split('.')[1]]
+    access_schema = config[resource]['schema']
+    if not access_schema:
+        access_schema = 'ssh'
+
```
andre-merzky: The above can be written as
```
access_schema = config[resource].get('schema', 'ssh')
```
Also, the resource configs are available on the session, so you don't need to parse them.  This test will break if RP changes config handling...

tests/test_issue_1514/test_rp_integration.py
```diff
@@ -0,0 +1,201 @@
...
+    ## Test on client
+    local_file = sample_data[0].split('.')[0] + '_2.' + sample_data[0].split('.')[1]
+    assert local_file in [os.path.basename(x) for x in glob('./*')]
+    # ------------------------------------------------------------------------------------------------------------------
+
+
+    # Verify the actionables were done for single folder transfer
+    # ------------------------------------------------------------------------------------------------------------------
+    remote_dir = rs.filesystem.Directory(   units[1].sandbox, 
+                                            session=session)
```
andre-merzky: interesting spaces... :-)

tests/test_issue_1514/test_rp_integration.py
```diff
@@ -0,0 +1,201 @@
...
+    # Verify the actionables were done for single folder transfer
+    # ------------------------------------------------------------------------------------------------------------------
+    remote_dir = rs.filesystem.Directory(   units[1].sandbox, 
+                                            session=session) 
+    ## Test on remote
+    assert sample_data[1] in [x.path for x in remote_dir.list()]
+    for x in remote_dir.list():
+        if remote_dir.is_dir(x):
+            child_x_dir = rs.filesystem.Directory(os.path.join(units[1].sandbox, x.path) ,
+                                                    session=session)
```
andre-merzky: you sometimes use `/`, and sometimes `os.path.join()`.  We don't need the method really, since we won't run on Windows anyway...

tests/test_issue_1514/test_usic.py
```diff
@@ -0,0 +1,303 @@
...
+access_schema = 'ssh'
+#-----------------------------------------------------------------------------------------------------------------------
+
+
+# Extract info from RP config file
+#-----------------------------------------------------------------------------------------------------------------------
+config_loc = '../../src/radical/pilot/configs/resource_%s.json'%resource_name.split('.')[0]
+path_to_rp_config_file = os.path.realpath(os.path.join(os.getcwd(),config_loc))
+cfg_file = ru.read_json(path_to_rp_config_file)[resource_name.split('.')[1]]
+
```
andre-merzky: Same comment on configs as above

tests/test_issue_1514/test_usic.py
```diff
@@ -0,0 +1,303 @@
...
+## Resolve environment variables in cfg
+if '$' in cfg_file['default_remote_workdir']:
+    target = cfg_file[access_schema]["filesystem_endpoint"].split('//')[1]
+
+    shell   = rsups.PTYShell('%s://%s'%(access_schema, target))
```
andre-merzky: you can just parse the file system endpoint here - no need to parse the URL and put is back together.  Also, URL parsing is tricky.

```
shell   = rsups.PTYShell(cfg_file[access_schema]["filesystem_endpoint"])
```

Or am I missing something?

src/radical/pilot/umgr/staging_input/default.py
vivek-bala: If you are commenting that we don't need the recursive flag at the user API, I agree.

src/radical/pilot/umgr/staging_input/default.py
vivek-bala: If you are commenting that we don't need the recursive flag at the user API, I agree. This can be made the default.

andre-merzky: Well, I am not sure.  You changed semantics here I believe, on purpose, so its on you to argue for the API change :)

vivek-bala: CREATE_PARENTS sets up all the parent folders of an file/folder, IIUC. Shouldn't that be the default always? The above was a way of implementing it and making sure the user does not overwrite it. There might be other way of enforcing it though.

src/radical/pilot/constants.py
vivek-bala: Agreed. I would suggest none to be string and we move to using integers as we do in saga.

src/radical/pilot/umgr/staging_input/default.py
vivek-bala: The above was to enforce that CREATE_PARENTS is always there by default(and cannot be overwritten by the user). That means the list will never be empty. If you think that is unnecessary, let me know.

tests/test_issue_1514/test_rp_integration.py
vivek-bala: Do you mean I can get the paths from the session? Can you elaborate on what you mean by 'RP changes config handling'? I don't understand how the test will break

tests/test_issue_1514/test_rp_integration.py
vivek-bala: That might have been me trying to keep things aligned with tabs :P Need to autoformat it to pep8

tests/test_issue_1514/test_rp_integration.py
vivek-bala: Makes sense. Either way I need to stick to one mode.

tests/test_issue_1514/test_usic.py
vivek-bala: You're right. Why did I even do that :-|

src/radical/pilot/umgr/staging_input/default.py
vivek-bala: Ok, assuming that user does not have specify CREATE_PARENTS and RECURSIVE flags. Even if the user does specify, it will not break the staging (64 | 64 = 64).

src/radical/pilot/umgr/staging_input/default.py
vivek-bala: I am not sure I understand. Are you referring to the fact that current some flags are integers and some are strings? I agree that all flags should be changed to integers. But I don't understand how the lamda/reduce is different from what you are suggesting. Let me know.

andre-merzky: Sorry for not being clear.

I want to say that the API *should* use OR'ed integers, not a list.  We should change examples and documentation.  We can make lists work transparently by OR'ing the elements on the API layer.  If that is in place, you won't have to handle lists in the components.  But  all this does not need to be part of this RP.

Did that make more sense?

vivek-bala: I think I understand what you are suggesting, but not why. 

Just to be sure, what you are suggesting is:
* Within RP, ```cud['flags']``` will always be an integer as opposed to a list.
* At the API, the user needs to do the OR'ing between different flags or provides a list of flags that are converted into an integer and assigned to cud['flags'] so that the lower layers get an integer.
Correct me if I am wrong.

I am trying to understand the motivation behind changing something that works now.

andre-merzky: Yes: we should have OR'ed values everywhere (but can convert on the fly in the API if needed).  Motivation: that's how we do it in the other layers - and we really should not have two different ways to render flags, depending on what layer we use...

vivek-bala: Makes sense. Thanks!

vivek-bala:
All tests pass with the latest changes. @georgeha @jdakka please check the changes and test them. @andre-merzky feel free to merge this PR.

tests/test_issue_1514/config.json
andre-merzky: See  #1573

vivek-bala:
> andre-merzky merged commit c2327a6 into master 5 hours ago

@andre-merzky Was the PR merged with master? Was that intended?

andre-merzky:
Yes, that was on purpose - we'll do a point release and will merge some PRs into master for that.
At the point of merge, devel and master were the same though, so it was essentially just a shortcut to avoid the detour over devel... ;-)

vivek-bala:
Ah okk.. Thanks :+1:

PR #1581, opened 2018-03-20
Feature/cu sandbox

andre-merzky:


mturilli:
Is this still actual?

andre-merzky:
Its an ok PR, but the use case for which this was implemented went away, and I am not sure we want to merge w/o a clear use case.  Lets discuss this on a devel call.  Thanks.

andre-merzky:
test was added and passes now.  Thanks!

PR #1582, opened 2018-03-20
Issue/1418

andre-merzky:


andre-merzky:
This PR needs testing against different PyMongo versions before merge.  Also, GPU needs to merged first.

PR #1614, opened 2018-05-08
Fix/issue 1612

kristofarkas:
Is this enough to support local execution with `aprun` on `bw`?

(Apologies if you are already working on this @andre-merzky)


itomaldonado:

andre-merzky:
Yes, this is indeed sufficient for the local `aprun` case.  You need to set `schema = 'local'` in the pilot description to trigger the use of that config section.

PR #1620, opened 2018-05-19
Experiment/iannis kill ibrun

andre-merzky:


andre-merzky:
ensure that units are started in their own process group, to ensure clean cancellation semantics.

PR #1622, opened 2018-05-22
Project/da scheduler algo

iparask:
This is a PR that will proceed with the code review of the project. 

 * Andre will review everything.
 * Jumana will review the continuous scheduler
 * George will review the LRMS classes
 * Giannis will review tests from Srinivas and Will. Vivek can chip in
 * Will and Srinivas can review whatever they want.


chengwill97:
I wanted to ask what assumptions I can make about the behavior of the configuration files regarding the new lfs* keys.
1. Will each config always have the "lfs_path_per_node" AND "lfs_size_per_node"?
2. Will it be valid for either key to exist and not the other? 
3. What happens if lfs_path_per_node DNE?
4. Should shared memory be included in the configs too?

chengwill97:
Update from Team 1:
The DA is currently not supported for the resource_ornl configs because we do not have data on these resources yet.

andre-merzky:
This branch is not in sync with devel, so it won't merge as is.  Please pull pending changes from devel before doing the reviews... - thanks!

chengwill97:
Regarding #1623, the resource manager slurm.py currently has the following
```
        self.cores_per_node = self._cfg.get('cores_per_node', 0)
        self.gpus_per_node  = self._cfg.get('gpus_per_node',  0)  # FIXME GPU
        self.lfs_per_node   = {'path':self._cfg.get('lfs_path_per_node',None),
                               'size':self._cfg.get('lfs_size_per_node',  0)
                              }
```
and I would like to change it to 
```
        self.cores_per_node = int(self._cfg.get('cores_per_node', 0))
        self.gpus_per_node  = int(self._cfg.get('gpus_per_node',  0))  # FIXME GPU
        self.lfs_per_node   = {'path':self._cfg.get('lfs_path_per_node',None),
                               'size':int(self._cfg.get('lfs_size_per_node',  0))
                              }
```

iparask:
Merging conflict is resolved and devel is being merge to this branch. I believe reviews can now start

src/radical/pilot/agent/rm/ccm.py
```diff
@@ -54,6 +54,9 @@ def _configure(self):
         # Some simple arithmetic
         self.cores_per_node = ccm_nodes_length / ccm_node_list_length
         self.gpus_per_node  = self._cfg.get('gpus_per_node', 0) # FIXME GPU
+        self.lfs_per_node   = {'path':self._cfg.get('lfs_path_per_node',None),
+                               'size':self._cfg.get('lfs_size_per_node',  0)
+                              }
```
andre-merzky: PEP-8 wants spaces around `:` and after `,`.

src/radical/pilot/agent/rm/fork.py
```diff
@@ -37,6 +37,9 @@ def _configure(self):
         # to preserve the previous behavior (1 node).
         self.cores_per_node = self._cfg.get('cores_per_node', self.requested_cores)
         self.gpus_per_node  = self._cfg.get('gpus_per_node', 0) # FIXME GPU
+        self.lfs_per_node   = {'path':self._cfg.get('lfs_path_per_node',None),
+                               'size':self._cfg.get('lfs_size_per_node',  0)
+                              }
```
andre-merzky: PEP-8

src/radical/pilot/agent/rm/lsf.py
```diff
@@ -57,6 +57,9 @@ def _configure(self):
         lsf_core_counts      = list(set(lsf_cores_count_list))
         lsf_cores_per_node   = min(lsf_core_counts)
         lsf_gpus_per_node    = self._cfg.get('gpus_per_node', 0) # FIXME GPU
+        lfs_lfs_per_node     = {'path':self._cfg.get('lfs_path_per_node',None),
+                                'size':self._cfg.get('lfs_size_per_node',  0)
+                               }
```
andre-merzky: PEP-8

src/radical/pilot/agent/rm/sge.py
```diff
@@ -56,6 +56,9 @@ def _sigusr2_handler():
         sge_cores_count_list = [int(line.split()[1]) for line in open(sge_hostfile)]
         sge_core_counts      = list(set(sge_cores_count_list))
         sge_gpus_per_node    = self._cfg.get('gpus_per_node', 0) # FIXME GPU
+        sge_lfs_per_node     = {'path':self._cfg.get('lfs_path_per_node',None),
+                                'size':self._cfg.get('lfs_size_per_node',  0)
+                               }
```
andre-merzky: PEP-8

src/radical/pilot/agent/rm/slurm.py
```diff
@@ -75,6 +75,9 @@ def _configure(self):
         # in case of a single partial node allocation.
         self.cores_per_node = self._cfg.get('cores_per_node', 0)
         self.gpus_per_node  = self._cfg.get('gpus_per_node',  0)  # FIXME GPU
+        self.lfs_per_node   = {'path':self._cfg.get('lfs_path_per_node',None),
+                               'size':self._cfg.get('lfs_size_per_node',  0)
+                              }
```
andre-merzky: PEP-8

src/radical/pilot/agent/rm/torque.py
```diff
@@ -54,6 +54,9 @@ def _configure(self):
             self._log.warning(msg)
 
         torque_gpus_per_node  = self._cfg.get('gpus_per_node', 0) # FIXME GPU
+        torque_lfs_per_node   = {'path':self._cfg.get('lfs_path_per_node',None),
+                                 'size':self._cfg.get('lfs_size_per_node',  0)
+                                }
```
andre-merzky: PEP-8

src/radical/pilot/agent/scheduler/base.py
```diff
@@ -265,21 +267,25 @@ def initialize_child(self):
 
         # initialize the node list to be used by the scheduler.  A scheduler
         # instance may decide to overwrite or extend this structure.
+
+        #-----------------------------------------------------------------------
+        # Change required if the DS changes.
+
```
andre-merzky: All changes are required by something - but I don't think that this prominent comment will help anybody.  What's `DS`? Also, the next couple of lines are not event related to data staging...

src/radical/pilot/agent/scheduler/base.py
```diff
@@ -344,19 +352,38 @@ def _change_slot_states(self, slots, new_state):
...
             #       of the node to the location on the list?
 
-            node = (n for n in self.nodes if n['uid'] == node_uid).next()
-            assert(node)
+            node = (n for n in self.nodes if n['uid'] == nodes['uid']).next()
+            # for n in self.nodes:
+            #     if n['uid'] == nodes['uid']:
+            #         node = n
+            #         break
+            # assert(node)
```
andre-merzky: If the list comprehension is too complex, then replace with the verbose version.  If it isn't, then the comment is not needed.

src/radical/pilot/agent/scheduler/base.py
```diff
@@ -332,8 +338,10 @@ def _change_slot_states(self, slots, new_state):
         have been allocated or deallocated.  For details on the data structure,
         see top of `base.py`.
         '''
+        # This method needs to change if the DS changes.
```
andre-merzky: This is not a useful comment.  This method needs to change on a gazillion other cases, and does not need to change on all DS changes...

src/radical/pilot/agent/scheduler/continuous.py
```diff
@@ -16,9 +16,37 @@
 
 # ------------------------------------------------------------------------------
 #
-# FIXME: make this a runtime switch depending on cprofile availability
-# FIXME: move this to utils (implies another parameter to `dec_all_methods()`)
+# ------------------------------------------------------------------------------
```
andre-merzky: Why was the `FIXME` removed?  Its not fixed, yet :/

src/radical/pilot/agent/scheduler/continuous.py
```diff
@@ -171,76 +196,80 @@ def _find_resources(self, node, requested_cores, requested_gpus,
...
-            if  (requested_cores and not free_cores) and \
-                (requested_gpus  and not free_gpus )     :
-                return [], []
+            # For partial requests the check simplifies: we just check if we
+            # have either, some cores *or* gpus *or* local_fs, to serve the
+            # request
+            if (requested_cores and not free_cores) and \
+                (requested_gpus and not free_gpus) and \
+                    (requested_lfs and not free_lfs):
+                return [], [], None
```
andre-merzky: inconsistent indentation

src/radical/pilot/agent/scheduler/continuous.py
```diff
@@ -171,76 +196,80 @@ def _find_resources(self, node, requested_cores, requested_gpus,
...
             # For non-partial requests (ie. full requests): its a no-match if
             # either the cpu or gpu request cannot be served.
-            if  requested_cores > free_cores or \
-                requested_gpus  > free_gpus     :
-                return [], []
-
+            if requested_cores > free_cores or \
+                requested_gpus > free_gpus or \
+                    requested_lfs > free_lfs:
+                return [], [], None
```
andre-merzky: inconsistent indentation

src/radical/pilot/agent/scheduler/base.py
```diff
@@ -344,19 +352,38 @@ def _change_slot_states(self, slots, new_state):
...
+            # print 'pre slot nodes: ', nodes
+            # print 'pre self: ',self.nodes
+            # print 'pre selected node: ',node
+
+            if new_state == rpc.BUSY:
+                print node['uid']
+                node['lfs']['size'] -= nodes['lfs']
+            else:
+                print node['uid']
+                node['lfs']['size'] += nodes['lfs']
```
andre-merzky: please remove the debug prints

src/radical/pilot/agent/scheduler/base.py
```diff
@@ -366,6 +393,8 @@ def slot_status(self):
         Returns a multi-line string corresponding to the status of the node list
         '''
 
+        # This method should change if the DS changes.
+
```
andre-merzky: same

src/radical/pilot/agent/scheduler/continuous.py
```diff
@@ -16,9 +16,37 @@
...
+#               {
+#                   'name'    : 'node_2',
+#                   'uid'     : yyyy,
+#                   'cores'   : 16,
+#                   'gpus'    : 2,
+#                   'lfs'     : 256
+#                },
+#               ]
+# Q: How should the nodes be selected for MPI based units?
+# lfs : in mb
```
andre-merzky: Code comment should be answers, not questions ;-)

src/radical/pilot/agent/scheduler/continuous.py
```diff
@@ -70,7 +97,6 @@ def finalize_child(self):
         # make sure that parent finalizers are called
         super(Continuous, self).finalize_child()
 
-
     # --------------------------------------------------------------------------
```
andre-merzky: POP-8 wants two empty lines

src/radical/pilot/agent/scheduler/continuous.py
```diff
@@ -80,8 +106,8 @@ def _configure(self):
         # * oversubscribe:
         #   Cray's aprun for example does not allow us to oversubscribe CPU
         #   cores on a node, so we can't, say, run n CPU processes on an n-core
-        #   node, and than add one additional process for a GPU application.  If
-        #   `oversubscribe` is set to False (which is the default for now),
+        #   node, and than add one additional process for a GPU application.
+        # If oversubscribe` is set to False (which is the default for now),
         #   we'll prevent that behavior by allocating one additional CPU core
```
andre-merzky: inconsistent indentation

src/radical/pilot/agent/scheduler/continuous.py
```diff
@@ -148,15 +173,15 @@ def _release_slot(self, slots):
         # reflect the request in the nodelist state (set to `FREE`)
         self._change_slot_states(slots, rpc.FREE)
 
-
     # --------------------------------------------------------------------------
```
andre-merzky: pep-8 wants 2 empty lines

src/radical/pilot/agent/scheduler/continuous.py
```diff
@@ -171,76 +196,80 @@ def _find_resources(self, node, requested_cores, requested_gpus,
...
         cores = list()
-        gpus  = list()
+        gpus = list()
+        lfs = None
 
-        # first count the number of free cores and gpus. This is
-        # way quicker than actually finding the core IDs.
+        # first count the number of free cores, gpus, and local file storage.
+        # This is way quicker than actually finding the core IDs.
         free_cores = node['cores'].count(rpc.FREE)
```
andre-merzky: You can't really `count the number of ... file storage` ...

src/radical/pilot/agent/scheduler/continuous.py
```diff
@@ -261,22 +290,22 @@ def _get_node_maps(self, cores, gpus, threads_per_proc):
 
         return core_map, gpu_map
 
-
     # --------------------------------------------------------------------------
```
andre-merzky: pep-8.  The amount of white soace changes make the diff longer than needed - this method for example was not really changed at all.

src/radical/pilot/agent/scheduler/continuous.py
```diff
@@ -286,32 +315,39 @@ def _alloc_nompi(self, cud):
...
 
         # make sure that the requested allocation fits on a single node
-        if  requested_cores > self._lrms_cores_per_node or \
-            requested_gpus  > self._lrms_gpus_per_node     :
+        if requested_cores > self._lrms_cores_per_node or \
+                requested_gpus > self._lrms_gpus_per_node or \
+                requested_lfs > self._lrms_lfs_per_node['size']:
+
+            print 'req:', requested_lfs, 'avail:', self._lrms_lfs_per_node
             raise ValueError('Non-mpi unit does not fit onto single node')
```
andre-merzky: please remove debug prints

src/radical/pilot/agent/scheduler/continuous.py
```diff
@@ -383,96 +426,121 @@ def _alloc_mpi(self, cud):
         # requested_cores > cores_per_node on impossible full-node-chunking
 
         cores_per_node = self._lrms_cores_per_node
-        gpus_per_node  = self._lrms_gpus_per_node
+        gpus_per_node = self._lrms_gpus_per_node
+        lfs_per_node = self._lrms_lfs_per_node   # this is a dictionary
+        # which have two key,s size (MB) and path
```
andre-merzky: Why is this explained here, but not on other uses?  Should probably be explained where its being defined, not when being used?

src/radical/pilot/agent/scheduler/continuous.py
```diff
@@ -383,96 +426,121 @@ def _alloc_mpi(self, cud):
...
+                                                    requested_cores=find_cores,
+                                                    requested_gpus=find_gpus,
+                                                    requested_lfs=find_lfs,
+                                                    core_chunk=threads_per_proc,
+                                                    partial=partial,
+                                                    lfs_chunk=requested_procs)
 
             # and check the result.
-            if not cores and not gpus:
+            if not cores and not gpus and not lfs:
```
andre-merzky: Does getting *only* lsf back make any sense?

andre-merzky:
This branch cannot run `examples/00_getting_started.py` on `localhost`...  Apparently the `slots` structure was changed w/o changing the executors which use that struct.  You may want to reconsider if that change is really necessary.  I don't think the executor is expected to evaluate the LFS information anyway...

src/radical/pilot/agent/scheduler/base.py
iparask: Removed.

src/radical/pilot/agent/scheduler/base.py
iparask: This assert is commented out for some reason. I brought it back and removed the comments.

src/radical/pilot/agent/scheduler/base.py
iparask: Removed

src/radical/pilot/agent/scheduler/base.py
iparask: removed

src/radical/pilot/agent/scheduler/base.py
iparask: Removed

src/radical/pilot/agent/scheduler/tagging.py
```diff
@@ -0,0 +1,312 @@
...
+
+from .continuous import Continuous
+
+
+# ------------------------------------------------------------------------------
+#
+# This is a simle extension of the Continuous scheduler which makes RP behave
+# like a FiFo: it will only really attempt to schedule units if all older units
+# (units with lower unit ID) have been scheduled.  As such, it relies on unit
+# IDs to be of the format `unit.%d'`.
```
andre-merzky: This comment does not match the class.

src/radical/pilot/agent/scheduler/tagging.py
```diff
@@ -0,0 +1,312 @@
...
+    # --------------------------------------------------------------------------
+    #
+    def __init__(self, cfg, session):
+
+        self.slots = None
+        self._tag_history = dict()
+
+        Continuous.__init__(self, cfg, session)
+
+
```
andre-merzky: This inherits from `Continuous`, but then basically re-implements the core scheduling parts of the parent class.  So any changes in the scheduling algorithm and any bug fixes etc. need to be done in both classes.

Can you think of a way to avoid this?

src/radical/pilot/agent/scheduler/tagging.py
vivek-bala: Yup, agreed. I created a separate scheduler/file to test just the tagging part. Both the DA and the tagging work now. So combining them into the continuous scheduler where the tagging will be optional.

The tagging.py file won't exist, this was just temporary.

src/radical/pilot/agent/scheduler/tagging.py
vivek-bala: The tagging.py file won't exist, this was just temporary to test the tagging capability. Do you agree with the tagging method I am using right now? Please see the tests under tests/test_da_scheduler/test_tagging_scheduler.py as well

andre-merzky: Hmm, I see.  The original idea for the RP schedulers though was to keep them lean and mean, and to quickly exchange them as needed / wanted for experiments.  Now the Cont. scheduler seems to grow into a do-it-all thing.  Can't blame you guys, since I started that trend I guess (the continuous scheduler also does scattered allocations...) - but still, I see this as problematic in the long run.  Please do reconsider if you see a way to use inheritance (or some other form of separation) for the tagging functionality.

vivek-bala: I completely agree. It is becoming bigger and precisely the reason where I kept the tagging in another file. Right now I am committing to the same continuous scheduler file. But let me propose separating two files: Continuous (as it was before), ContinuousDataAware (with tagging).

What do you think @iparask and team, @andre-merzky ?

src/radical/pilot/agent/scheduler/continuous.py
vivek-bala: pep8ified all the DA related files

src/radical/pilot/agent/scheduler/continuous.py
vivek-bala: Not sure if this still applies?

src/radical/pilot/agent/scheduler/continuous.py
```diff
@@ -211,7 +211,7 @@ def _find_resources(self, node, requested_cores, requested_gpus,
         free_gpus = node['gpus'].count(rpc.FREE)
         free_lfs = node['lfs']['size']
 
-        # print 'free: ',node['uid'], free_cores, free_gpus, free_lfs
+        alloc_lfs = alloc_cores = alloc_gpus = 0
```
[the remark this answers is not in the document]
iparask: From @vivek-bala : I agree. That was all debug msgs, I am going to cleanup the file now

src/radical/pilot/agent/scheduler/tagging.py
vivek-bala: +1

andre-merzky: Well, `ContinuousBase` could be the old, plain continuous scheduler, but refactored so that it exposes methods which can be reused?  I don't want to pre-scribe that structure though...   

Either way, I'm glad that you are open to the argument.   I share your concern about the deadlines - in the light of that, let me know if I can help with separating those classes...

vivek-bala:
Local integration test is pending https://github.com/radical-cybertools/radical.pilot/issues/1637

src/radical/pilot/agent/lm/aprun.py
```diff
@@ -54,9 +54,22 @@ def construct_command(self, cu, launch_script_hop):
...
+        #         'nodes'         : [{  'name': node_1,
+        #                               'uid': node_uid_1,
+        #                               'core_map': [[0, 2], [4, 6]],
+        #                               'gpu_map': [[0]],
+        #                               'lfs': 1024},
+        #                            {  'name': node_2,
+        #                               'uid': node_uid_2,
+        #                               'core_map': [[1, 3], [5, 7]],
+        #                               'gpu_map': [[0]],
+        #                               'lfs': 1024}
```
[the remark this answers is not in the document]
andre-merzky: replying here on the `task_offsets`: those were not changed by this PR, and only the comments were added, so you should not need to understand what those task offsets are.  For completeness though: instead of an absolute core position, some launch methods require us to specify the position as a relative offset from the start of the allocation, or the start of a node etc - we call those relative location `task_offsets`.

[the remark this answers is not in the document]
georgeha: Thanks Andre. 
Somebody wrote that due to the changes in this PR,  this code needs to be revisited.
That is why I was not sure about it.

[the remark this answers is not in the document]
andre-merzky: Ah, I see, makes sense. But that revision does not need to happen in this PR, fortunately :-)

iparask:
@georgeha, Can you please link to part of the code? Also, if it part of code that we did not change, that is the commit is from devel, I suggest not to include it here.

src/radical/pilot/agent/lm/ibrun.py
```diff
@@ -46,9 +46,9 @@ def construct_command(self, cu, launch_script_hop):
             raise RuntimeError('insufficient information to launch via %s: %s' \
                     % (self.name, slots))
 
-        task_offsets = slots['task_offsets']
-        assert(len(task_offsets) == 1)
-        ibrun_offset = task_offsets[0]
+        task_offsets = slots['task_offsets']        # This needs to revisited 
+        assert(len(task_offsets) == 1)              # since slots structure has 
+        ibrun_offset = task_offsets[0]              # changed
```
georgeha: @iparask  here is this part

src/radical/pilot/agent/lm/dplace.py
```diff
@@ -42,9 +42,9 @@ def construct_command(self, cu, launch_script_hop):
             raise RuntimeError('insufficient information to launch via %s: %s' \
                     % (self.name, slots))
 
-        task_offsets = slots['task_offsets']
-        assert(len(task_offsets) == 1)
-        dplace_offset = task_offsets[0]
+        task_offsets = slots['task_offsets']        # This needs to revisited 
+        assert(len(task_offsets) == 1)              # since slots structure has 
+        dplace_offset = task_offsets[0]             # changed
```
georgeha: @iparask  also here

iparask: I believe this is not part of this PR and is part of devel. @andre-merzky can you verify?

andre-merzky: agree - that was already discussed with @georgeha .

andre-merzky:
One more comment on the changed scheduler / slot structure: can we introduce defines for the tuple indices, so that we have something like: `node[UID]` instead of `node[1]`?  I find myself digging through the code again and again to check which place is UID vs. name or core map vs. gpu map etc...

iparask:
Hello Andre, can you point exactly which part creates the confusion?

andre-merzky:
> Hello Andre, can you point exactly which part creates the confusion?

For example, in the ORTE (and all other) launch method(s), we have to walk through the node list of the allocated slots.  But then I have to dig out the node uid as `node[1]`, etc - and the definition of that structure is far away from the LM (which belongs to the executor), it seeps over from the scheduler.

andre-merzky:
@iparask : sorry for the noise, please ignore the comments about the slot structure indexes for now...

andre-merzky:
Action is on @iparask and @andre-merzky to run tests before merge - otherwise review is considered completed and passed at that point.

Ioannis: Bridges, Wrangler
Andre: Titan, Blue Waters, Stampede2

iparask:
I works on Bridges and Wrangler for me

andre-merzky:
Thanks to everybody involved - well done! :-)

PR #1651, opened 2018-07-03
rewrite and simplify the fifo scheduler

andre-merzky:


src/radical/pilot/agent/scheduler/continuous_fifo.py
```diff
@@ -19,40 +22,103 @@ class ContinuousFifo(Continuous):
...
+            # cache ID int to avoid repeated parsing
+            for unit in units:
+                if 'serial_id' not in unit:
+                    unit['serial_id'] = int(unit['uid'].split('.')[-1])
+
+                assert(unit['serial_id'] > self._last)
+
+            # add units to pool and resort it
+            self._ordered_wait_pool += units
+            self._ordered_wait_pool.sort(key=lambda x: x['serial_id'])
```
vivek-bala: Isn't 'units' already a list (hence ordered)? Is this required in that case? Since the 1) new units (which will be ordered) will have a serial id greater than those in the wait pool and 2) you append the new units to the end of the wait pool.

vivek-bala: Nvm, I think I understand why this is required for the wait pool.

vivek-bala:
I think the scheduler operates as intended. I am wondering whether 'fifo scheduler' is the correct name for it. Since we actually wait for 'unit.X' to arrive even if we already have 'unit.X+1' in our wait pool, maybe we should call it a 'sequence/sequential scheduler' or something like that?

I think FIFO describes that the order of scheduling of units is the order in which they arrive. The current scheduler does not account for the arrival order but the id of the arriving unit. What do you think?

mturilli:
If memory serves well (and we should all refresh our memory about scheduling algorithms) this should implement a type of priority scheduling algorithm. Scheduling depends on the id and some ordering is imposed over these IDs (e.g, lowest integer first).

src/radical/pilot/agent/scheduler/continuous_fifo.py
andre-merzky: Hey Vivek - this PR is outdated - I should remove it.  Please have a look at #1704 which implements a more generic ordering schema, which does not depend on the unit IDs.

andre-merzky:
> I think FIFO describes that the order of scheduling of units is the order in which they arrive. The current scheduler does not account for the arrival order but the id of the arriving unit. What do you think?

Well, now this is getting confusing: `scheduling in the order they arrive` is what the normal cont. scheduler did - but the problem with your workload was, I quote: "we expected it to behave like a fifo" - and with that you meant not the order in which the units arrive, but the order in which units are submitted (which happens to be reflected by the unit ID).  :-)

I think though the FiFo name should die, and we should switch to the the explicitly specified ordering of  #1704 ...

vivek-bala:
Yea I agree. My comment was not regarding the implementation of the scheduler but only its name. I agree that calling it FIFO, initially, might not have been correct. I'll check out #1704 and give the example you showed last week a try.

mturilli:
Can this help? https://ieeexplore.ieee.org/document/4634

andre-merzky:
Closed as outdated - but I'll keep the branch alive for now.

PR #1704, opened 2018-08-30
Feature/schedule ordered

andre-merzky:
This PR adds an `ContinuousOrdered` scheduler, which specifically aims to support the use case of ensembles of heterogeneous pipelines. The algorithm is documented [here](https://github.com/radical-cybertools/radical.pilot/blob/feature/schedule_ordered/src/radical/pilot/agent/scheduler/continuous_ordered.py).


andre-merzky:
@vivek-bala , @jdakka : Vivek, Jumana, can you have a look at this PR, or should I reassign it?  Thanks :-)

tests/test_fifo_scheduler.py
```diff
@@ -0,0 +1,146 @@
...
+import pprint
+import pytest
+
+import threading as mt
+
+import radical.utils           as ru
+import radical.pilot           as rp
+import radical.pilot.constants as rpc
+
+from radical.pilot.agent.scheduler.continuous_fifo import ContinuousFifo
```
jdakka: This module no longer exists, it should be `from radical.pilot.agent.scheduler.continuous_ordered import ContinuousOrdered`

tests/test_fifo_scheduler.py
```diff
@@ -0,0 +1,146 @@
...
+# Cleanup any folders and files to leave the system state
+# as prior to the test
+def tearDown():
+
+    pass
+
+
+# ------------------------------------------------------------------------------
+# Test non mpi units
+@mock.patch.object(ContinuousFifo, '__init__', return_value=None)
```
jdakka: same here: `ContinuousFifo` should be `ContinuousOrdered`

tests/test_fifo_scheduler.py
```diff
@@ -0,0 +1,146 @@
...
+@mock.patch.object(ContinuousFifo, 'advance')
```
jdakka: same here: `ContinuousFifo` should be `ContinuousOrdered`

tests/test_fifo_scheduler.py
```diff
@@ -0,0 +1,146 @@
...
+@mock.patch.object(ru.Profiler, 'prof')
+@mock.patch('radical.utils.raise_on')
+def test_fifo_scheduler(mocked_init,
```
jdakka: Update `test_fifo_scheduler` to `test_ordered_scheduler`

tests/test_fifo_scheduler.py
```diff
@@ -0,0 +1,146 @@
...
+                        mocked_method,
+                        mocked_profiler,
+                        mocked_raise_on):
+
+    cfg, session = setUp()
+
+    component = ContinuousFifo(cfg=dict(), session=session)
```
jdakka: `ContinuousFifo` -> `ContinuousOrdered`

mturilli:
Is this still actual?

andre-merzky:
Yes - I'll resolve the conflicts and merge.

andre-merzky:
This is ready for another review - @vivek-bala , would you have a look, please?

andre-merzky:
The unit tests for this PR are now added and pass.

andre-merzky:
PS.: @iparask : how can I re-trigger a travis test run?  For example, once a MongoDB error is fixed on the DB host, but no commit is needed in code?

andre-merzky:
Well, the test fails when running in Travis.  I am annoyed.   Either travis has some resource limitations (which I would not be surprised about: I see many concurrent connections in MongoDB), an the fails are failing for this, or something else is up.  Either way, since I can't get logfiles back and the test worked on other machines, the I am not sure what to do with that information.

mturilli:
Annoying but I guess this is the use of having a third party system running out tests. I would take this as a indication of fragility either of the code or of our tests. For the logs, I would look into: `after_failure` at https://docs.travis-ci.com/user/job-lifecycle/#the-job-lifecycle

vivek-bala:
@andre-merzky you will have to manually restart the test suite. When you go to the travis page for the repo, you should see a "Restart build" button slightly towards the right (see https://travis-ci.org/radical-cybertools/radical.entk).

There are limits on the processes, its resources and lifetime. Probably there are limits for the number of connections as well. Unfortunately, I couldn't find any numbers in their documentation. This is also the reason why we have [this](https://github.com/radical-cybertools/radical.entk/blob/devel/.travis.yml#L21) and not ```coverage run -m pytest -vvv tests/test_*```. The latter hangs and then times out.

PR #1706, opened 2018-09-05
Summit/scheduler

vivek-bala:
Starting a PR to share ideas before implementation


vivek-bala:
I think we need to merge project/DA-scheduler with devel (or reimplement some things from project/da-scheduler into summit/scheduler). Following up on my idea from last week, my suggestion is to
change
```
for node, node_uid in self._lrms_node_list:
    self.nodes.append({
        'name': node,
        'uid': node_uid,
        'cores': [rpc.FREE] * self._lrms_cores_per_node,
        'gpus': [rpc.FREE] * self._lrms_gpus_per_node,
        'lfs': self._lrms_lfs_per_node
    })
```

to 

```
for node, node_uid in self._lrms_node_list:
    self.nodes.append({
        'name': node,
        'uid': node_uid,
        'sockets': [
            {
                'cores': [rpc.FREE] * self._lrms_cores_per_node,
                'gpus': [rpc.FREE] * self._lrms_gpus_per_node,
                'lfs': self._lrms_lfs_per_node
            },
            {
                'cores': [rpc.FREE] * self._lrms_cores_per_node,
                'gpus': [rpc.FREE] * self._lrms_gpus_per_node,
                'lfs': self._lrms_lfs_per_node
            }
        ]})
```

Modify find_resources method accordingly: similar to what exists right now
but with the added guard that threads from the same process don't land on
different sockets.

iparask:
This modification is based on the fact that the threads of a process should all be in the same socket. I see two possibilities here:
1. We either separate completely the node to two sub-nodes.
2. Separate only the sockets and leave everything as is.

The first also separates the GPUs per socket. The GPUs based on Sumit's documentation is closer to a specific socket, but not exclusive. That means that a process in any socket can use any GPU. 

The second, although it seems correct is already done based on the `COREID`, so I still do not understand why it is needed. Separating the sockets can easily be done by dividing the `COREID` with `CORES_PER_NODE/2`. As far as I know all the resources we are using provide their cores into two socket, except from then KNL nodes on Stampede2.

There is a valid argument that the scheduler should place the process closer to the GPU that it will use, but unless we have an application that has very strict communication requirements between CPU and GPU, I am not sure if the effort is fully justified

Also, shouldn't the scheduler already not separate threads that belong to the same process?

vivek-bala:
> The first also separates the GPUs per socket. The GPUs based on Sumit's documentation is closer to a specific socket, but not exclusive. That means that a process in any socket can use any GPU.

I agree that processes can access GPUs across sockets. My suggestion is based on the need to capture the node architecture for scheduling. For a node, we need to get the relationship between sockets, cores, and gpus. I agree that in the case of socket-core, the relationship is "belongs to" and in socket-gpu, the relationship is "close to". If required, we can separate it into two data structures (or even a different data structure than dict if meaningful).

> The second, although it seems correct is already done based on the COREID, so I still do not understand why it is needed. Separating the sockets can easily be done by dividing the COREID with CORES_PER_NODE/2. As far as I know all the resources we are using provide their cores into two socket, except from then KNL nodes on Stampede2.

I think that's correct. My suggestion is to find the core-socket mapping prior to scheduling (in the resource manager to be precise) and not to do it while scheduling/in the scheduler. I am not aware if it is global (certainly does apply to Summit), but we do seem to be assuming that core ids are sequential on each socket. For example, can it be the case that all core ids are separated as even and odd on two sockets.

> There is a valid argument that the scheduler should place the process closer to the GPU that it will use, but unless we have an application that has very strict communication requirements between CPU and GPU, I am not sure if the effort is fully justified

I don't think this came up in the Princeton use case, but I am sure it will be required since they are quite concerned with the performance.


> Also, shouldn't the scheduler already not separate threads that belong to the same process?

I am not sure I understand the question here. We already have guards to make sure that threads of the same process do not land on different nodes.

vivek-bala:
Proposed DS accepted by all

vivek-bala:
Scattering threads across different sockets to be done via a flag

src/radical/pilot/agent/rm/base.py
```diff
@@ -74,14 +74,15 @@ def __init__(self, cfg, session):
...
-        self.gpus_per_node   = None
-        self.lfs_per_node    = None
+        self.lm_info            = dict()
+        self.lrms_info          = dict()
+        self.slot_list          = list()
+        self.node_list          = list()
+        self.agent_nodes        = dict()
+        self.sockets_per_node   = None
+        self.cores_per_socket   = None
+        self.gpus_per_socket    = None
```
iparask: The notion of `gpus_per_socket` is not correct. In addition, you removed `cores_per_node` and added `socket_per_node` and `cores_per_socket`.  I suggest that instead of all these changes to go to the base class, they can easily go to the actual LSF resource manager and avoid any huge changes in all resource managers.

As we have discussed, there are GPUs which happen to be closer to a socket compared to the other. Still for the base class it should be as it was and any changes like that should be part of the LSF resource manager class

src/radical/pilot/agent/rm/base.py
```diff
@@ -182,13 +183,14 @@ def __init__(self, cfg, session):
...
-        self.lrms_info['agent_nodes']    = self.agent_nodes
-        self.lrms_info['lfs_per_node']   = self.lfs_per_node
+        self.lrms_info['name']              = self.name
+        self.lrms_info['lm_info']           = self.lm_info
+        self.lrms_info['node_list']         = self.node_list
+        self.lrms_info['sockets_per_node']  = self.sockets_per_node
+        self.lrms_info['cores_per_socket']  = self.cores_per_socket
+        self.lrms_info['gpus_per_socket']   = self.gpus_per_socket
+        self.lrms_info['agent_nodes']       = self.agent_nodes
+        self.lrms_info['lfs_per_node']      = self.lfs_per_node
```
iparask: Same as the above comment.

src/radical/pilot/agent/rm/base.py
```diff
@@ -160,8 +161,8 @@ def __init__(self, cfg, session):
                 self._log.info("lrms config hook succeeded (%s)" % lm)
 
         # For now assume that all nodes have equal amount of cores and gpus
-        cores_avail = (len(self.node_list) + len(self.agent_nodes)) * self.cores_per_node
-        gpus_avail  = (len(self.node_list) + len(self.agent_nodes)) * self.gpus_per_node
+        cores_avail = (len(self.node_list) + len(self.agent_nodes)) * self.cores_per_socket * self.sockets_per_node
```
iparask: Same as the above comment

src/radical/pilot/agent/rm/lsf.py
```diff
@@ -48,26 +48,29 @@ def _configure(self):
...
+                       lsf_core_counts, lsf_cores_per_socket)
 
         # node names are unique, so can serve as node uids
-        self.node_list      = [[node, node] for node in lsf_node_list]
-        self.cores_per_node = lsf_cores_per_node
-        self.gpus_per_node  = lsf_gpus_per_node
-        self.lfs_per_node   = lfs_lfs_per_node
-
-
+        self.node_list = [[node, node] for node in lsf_node_list]
```
iparask: Why is the `node` inserted twice in the `node_list`? I really do not understand the reason. If it is for Summit specifically, then why didn't you fix all other attributes as well?

src/radical/pilot/agent/rm/base.py
vivek-bala: I think we did agree on 'gpus_per_socket', 'cores_per_socket' and 'sockets_per_node' to be used. I agree that these shouldn't be in the base class though. Similar to how we have ContinuousSummit scheduler, I can create rm/lsf_summit which will overwrite all/most methods of base (including the constructor). I have to overwrite the constructor as well as since the operations inside depend on whether we have cores_per_node or cores_per_socket+sockets_per_node. 

I wanted to discuss this in today's call. But I'll leave a comment here for now. I think eventually base class will change as we try to port the socket knowledge to all resource managers (similar to the case of the scheduler). Thanks for looking into it.

src/radical/pilot/agent/rm/lsf.py
vivek-bala: I think there should be a comment above that line: "# node names are unique, so can serve as node uids".

vivek-bala:
Hi all, the PR is ready for review. I have reassigned the PR. I have addressed the comments from the previous discussion on the commit. I have added inline comments as well.

iparask:
Shouldn't the PR be towards ` feature/summit` branch? That is the one that has all the work for Summit?

src/radical/pilot/agent/rm/lsf_summit.py
```diff
@@ -0,0 +1,254 @@
...
+        #    structure compatible with that expected in the base class.
+        # 3) Change the base class --- Out of scope of this project
+
+        # 3 is probably the correct approach, but long term. 2 is not a
+        # good approach as we are striving to keep a child class compatible
+        # with a base class (this should never be the case).
+        # We go ahead with 1, process of elimination really, but with the
+        # advantage that we have the code content that will be required when
+        # we implement 3, the long term approach.
+
```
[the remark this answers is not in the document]
vivek-bala: Generally, the base consists attributes and methods common across (and required) by all the
child classes. The child will never needs to be "compatible" with a base class. But in our case,
the base assumes a node architecture which is, currently, common across all child classes but
one (LSF_SUMMIT). In ideal case, we will/should never encounter such a scenario.

So in effect, the long term solution is to modify the base class and investigate whether other
child classes should adhere to the same node architecture (i.e., acknowledge the presence
of sockets if it is just 1) as in LSF_SUMMIT. Hence, temporarily, we pick 1 till we can implement
3. 

Hope that explains the rationale behind the choice.

src/radical/pilot/agent/rm/lsf_summit.py
```diff
@@ -0,0 +1,254 @@
...
+        self._log.info("Found unique core counts: %s Using: %d",
+                       lsf_core_counts, lsf_cores_per_socket)
+
+        # node names are unique, so can serve as node uids
+        # The structure of the node list is 
+        # [[node1 name, node1 uid],[node2 name, node2 uid]]
+        # The node name and uid can be the same
+
+        self.node_list        = [[node, node] for node in lsf_node_list]
```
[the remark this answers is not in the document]
vivek-bala: I agree. Thanks Giannis! I had missed that one. This has been fixed and tested with specification of the uid in lieu of the node name in the second index of each node list entry.

src/radical/pilot/configs/resource_ornl.json
```diff
@@ -203,6 +203,92 @@
...
+     #  "saga_jd_supplement"          : {"candidate_hosts" : ["titan"]}
+    },
+
+    "summitdev": {
+        "description"                 : "The Cray XK7 supercomputer located at the Oak Ridge Leadership Computing Facility (OLCF), (https://www.olcf.ornl.gov/titan/)",
+        "notes"                       : "Requires the use of an RSA SecurID on every connection.",
+        "schemas"                     : ["local"],
+        "local"                       : {
+            "job_manager_hop"         : "fork://localhost/",
+            "job_manager_endpoint"    : "lsf://localhost/",
```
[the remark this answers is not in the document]
iparask: It disappeared as outdated when you change the target branch. It should not be `lsf`, bu `lsfsummit`

tests/test_summit/test_rm_create_on_summit.py
```diff
@@ -0,0 +1,57 @@
...
+
+    lrms = rpa_rm.RM.create(name=cfg['lrms'], cfg=cfg, session=session)
+    assert lrms.lrms_info['agent_nodes'] == {}
+    assert lrms.lrms_info['cores_per_socket'] == 10
+    assert lrms.lrms_info['gpus_per_socket'] == 6
+    assert lrms.lrms_info['lfs_per_node'] == {'path': None, 'size': 0}
+    assert lrms.lrms_info['lm_info'] == {}
+    assert lrms.lrms_info['name'] == 'LSF_SUMMIT'
+    assert lrms.lrms_info['sockets_per_node'] == 2
+
```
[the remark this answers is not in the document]
iparask: It's coming from the LRMS configuration hook that exists in some launch methods. Irrelevant here.

tests/test_summit/test_rm_create_on_localhost.py
```diff
@@ -0,0 +1,76 @@
+import radical.utils as ru
+import radical.pilot as rp
+from radical.pilot.agent import rm as rpa_rm
+import os
+import glob
+import shutil
+from pprint import pprint
```
iparask: I do not think this is needed

PR #1714, opened 2018-09-20
add shellfs spawner

andre-merzky:


andre-merzky:
This PR contains two major changes:

ShellFS spawner: the POPEN spawner lost units at large scale.  This PR introduces a shell based spawner which communicates with the Python counterpart via two named pipes.

Unit sandbox tarring: for large number of units, we create a tarball of their sandboxes, stage it to the target resource, and untar it for fast dir tree creation.  The existing schemes is (a) slowed down by many (local) FS ops, and (b) breaks for large scales when the local file system has limits on number of dir entries (for example, `ext3` is limited to 32k).  The new scheme creates a POSIX tar file programatically, which is faster and also smaller (so optimizes transfer) - but not compressed (so pessimizes transfer ;-) ).

mturilli:
Time to start thinking about a file-based db? Please note the Jenkins failures. We really need to fix that!

andre-merzky:
Time to get rid of the DB you mean ;-)

andre-merzky:
@iparask I Ioannis, can you have a look at this PR, or should I reassign?  Thanks :-)

examples/00_getting_started.py
```diff
@@ -12,6 +12,13 @@
 
 dh = ru.DebugHelper()
 
+N = 2
+N = 1024 * 64
+N = 1024 * 2
+
+P = 16
+U = 33
+
```
iparask: What are these and where are they being used?

src/radical/pilot/agent/rm/fork.py
```diff
@@ -36,10 +36,13 @@ def _configure(self):
         # cores into that many virtual nodes.  cpn defaults to requested_cores,
         # to preserve the previous behavior (1 node).
         self.cores_per_node = self._cfg.get('cores_per_node', self.requested_cores)
-        self.gpus_per_node  = self._cfg.get('gpus_per_node', 0) # FIXME GPU
+        self.gpus_per_node  = self._cfg.get('gpus_per_node', 0)  # FIXME GPU
+
+        if not self.cores_per_node:
```
iparask: Isn't this a bit redundant, since the get above does set the cores per node to `self.requested_cores`?

src/radical/pilot/configs/resource_ornl.json
```diff
@@ -97,8 +97,8 @@
         "lrms"                        : "TORQUE",
         "agent_type"                  : "multicore",
         "agent_config"                : "cray",
-        "agent_scheduler"             : "CONTINUOUS",
-        "agent_spawner"               : "POPEN",
+        "agent_scheduler"             : "HOMBRE",
```
iparask: Isn't `HOMBRE` use case specific? Or it is a requirement on Titan to use homogeneous tasks in the sense of resource requests?

I do not believe that using a scheduler that is so specific as the default is the best way to go. I would suggest that either offer some other configuration, or change the use cases that really needed it change it programmatically.

src/radical/pilot/utils/component.py
```diff
@@ -1429,8 +1429,12 @@ def publish(self, pubsub, msg):
         self.is_valid()
 
         if pubsub not in self._publishers:
-            raise RuntimeError("can't route '%s' notification: %s" % (pubsub,
+            self._log.warn("can't route '%s' notification: %s" % (pubsub,
```
iparask: What is the reason that this is reduced from an error to a warning? I am not say it should stay as an error, I just to be sure.

src/radical/pilot/agent/rm/fork.py
andre-merzky: The setting can result in `None` or `0`, and either would be invalid - thus the additional check...

src/radical/pilot/configs/resource_ornl.json
andre-merzky: Thanks, fixed.

examples/00_getting_started.py
andre-merzky: For completeness: the debug helper attaches some hooks which dumps stack traces for *all* threads on termination.

iparask: okay!

src/radical/pilot/agent/executing/shell_fs.py
```diff
@@ -0,0 +1,527 @@
...
+        self._terminate.set()
+
+        try:
+            self._sh.kill()
+        except:
+            pass
+
+        try:
+            os.unlink(self._fifo_cmd_name)
+            os.unlink(self._fifo_inf_name)
```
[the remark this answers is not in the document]
andre-merzky: yes, I'll add that.  Thanks.

src/radical/pilot/agent/executing/shell_fs.py
```diff
@@ -0,0 +1,527 @@
...
+    #
+    def spawn(self, launcher, cu):
+
+        uid     = cu['uid']
+        sandbox = '%s/%s' % (self._pwd, uid)
+
+        try:
+            os.makedirs(sandbox)
+        except:
+            pass
```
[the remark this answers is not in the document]
andre-merzky: This is expected to happen very frequently, whenever the sandbox was created by data staging.  An actual error will pop up some lines later when trying to write the CU startup script.

src/radical/pilot/agent/executing/shell_spawner_fs.sh
```diff
@@ -0,0 +1,235 @@
...
+        LOG="$WORK/sh.$SID.log"
+        
+        test -z "$BASE" && usage 1 'missing base dir' 
+        test -z "$WORK" && usage 1 'missing work dir'
+        test -z "$SID"  && usage 1 'missing sid'
+    
+        export BASE
+        export WORK
+        export SID
+        export LOG
```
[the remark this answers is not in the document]
andre-merzky: to make then available to sub shells and child processes.

PR #1734, opened 2018-10-09
Feature/jsrun

iparask:
This Pull Request will keep track of the merging between the specific scheduler for Summit and its Launch method.

Based on Issue #1733


iparask:
They are mergable now. I will commit and continue with testing

PR #1745, opened 2018-10-16
Fix/issue 1744

### LINKED ISSUE #1744: SAGA job name should not be fatal on PMGR
Essentially line [694](https://github.com/radical-cybertools/radical.pilot/blob/devel/src/radical/pilot/pmgr/launching/default.py#L694) of PMGR should be a warning not a runtime error

andre-merzky:
This fixes #1744

### DIFF
--- a/src/radical/pilot/pmgr/launching/default.py
+++ b/src/radical/pilot/pmgr/launching/default.py
@@ -661,7 +661,7 @@ class Default(PMGRLaunchingComponent):
         self._log.debug('tar done: %s, %s, %s', j.state, j.stdout, j.stderr)
 
         for pilot in pilots:
-            self._prof.prof('staging_in_stop', uid=pilot['uid'])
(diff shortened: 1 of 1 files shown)


iparask:
I would prefer it as a warning. Do you agree with that?

andre-merzky:
The need for it is basically gone this way.  Or am I missing what you'd prefer to warn about?  The FIXME?

iparask:
I was more in favor of having the message for the name as a warning instead of a runtime error. That way we know we are missing something, it is not fatal and if we really need it at some point, we still have the warning to remind us :)

All that being said, if you are comfortable with this implementation, move forward and merge!

andre-merzky:
Yeah, I see what you mean.  But a warning won't help, as in we need to work around the missing name then anyway - if we don't it is not recoverable.  Lets catch it in tests on the SAGA layer then...

PR #1773, opened 2018-11-30
edits for proposal to separate input from result

vivek-bala:
This PR consists the proposal from Giannis and myself. The input descriptions are separated from the test results. The corresponding results for each description is identified by using the unit id as the tag.

Two key points:
1) Not all the input descriptions apply to each lm test. Currently, the corresponding input descriptions for each lm test can be identified by looking at the results_(lm method).json file. 
2) This approach allows the input descriptions to be used by other tests (possibly in other components as well) which I believe was the intention (the test_cases are located at /tests/, i.e., outside of ./tests/lm/).

@iparask Please add any other points in case I missed any.


iparask:
I do not think you missed any point

andre-merzky:
Both points apply to both approaches.  So now it boils down to having all test results based on the same original unit in the same file as the unit, or to distribute them over files.  

PS.: But you seem to have left the `setup` part with the unit - didn't you plan to split that out, too?  I am not complaining though, just observing ;-)

iparask:
@andre-merzky can you merge the two `lm_test` branches from you and from Vivek and I to the original one or delete what we do not need? I want to continue with test writting.

PR #1776, opened 2018-12-06
Expand env variables in LFS path

### LINKED ISSUE #1775: LFS path that contain evnironment variables do not expand
An example is the following. On Bridges the local filesystem of the nodes is accessed by `$LOCAL`. The scheduler put that value to the CUs environment with the key `NODE_LFS_PATH`. Using the `popen` executor and `ssh` as the launch method creates the following confilct:

The Launch Method passes through ssh the `NODE_LFS_PATH=$LOCAL` and the executor adds in the unit script the following line:
`export NODE_LFS_PATH=$LOCAL` 

When `$LOCAL` is not present in the new `ssh` connection then the line in the script will empty the environment variable `NODE_LFS_PATH`, assuming that it was passed correctly by the `ssh` command. 

I propose the following solution:
In LRMS check if the lfs path has environment variables in it, by simple checking if there is a `$`, if yes then expand it with something like [`os.path.expandvars`](https://docs.python.org/2/library/os.path.html#os.path.expandvars)

That way the `NODE_LFS_PATH` will always be an absolute value and exporting inside the script should do the trick. I also checked the resources that have environment variables as LFS. They all depend on values that have to do with the jodid and/or userid, so the assumption that it is the same for all nodes still holds

iparask:
Fixes #1775 

In the case where the LFS path has environment variables in it, it gets expanded in the resource manager. As a result the scheduler and the unit description just get an absolute path.

### DIFF
--- a/src/radical/pilot/agent/rm/ccm.py
+++ b/src/radical/pilot/agent/rm/ccm.py
@@ -55,9 +55,8 @@ class CCM(LRMS):
         self.cores_per_node = ccm_nodes_length / ccm_node_list_length
         self.gpus_per_node  = self._cfg.get('gpus_per_node', 0) # FIXME GPU
 
-        lfs_path = self._cfg.get('lfs_path_per_node', None)
-        if '$' in path:
-            lfs_path = os.path.expandvars(path)
+        lfs_path = self._cfg.get('lfs_path_per_node', '')
+        lfs_path = os.path.expandvars(lfs_path)
 
         self.lfs_per_node   = {'path' : lfs_path,
                                'size' : self._cfg.get('lfs_size_per_node', 0)
--- a/src/radical/pilot/agent/rm/ccm.py
+++ b/src/radical/pilot/agent/rm/ccm.py
@@ -54,7 +54,12 @@ class CCM(LRMS):
         # Some simple arithmetic
         self.cores_per_node = ccm_nodes_length / ccm_node_list_length
         self.gpus_per_node  = self._cfg.get('gpus_per_node', 0) # FIXME GPU
-        self.lfs_per_node   = {'path' : self._cfg.get('lfs_path_per_node', None),
+
+        lfs_path = self._cfg.get('lfs_path_per_node', None)
+        if '$' in path:
+            lfs_path = os.path.expandvars(path)
+
+        self.lfs_per_node   = {'path' : lfs_path,
                                'size' : self._cfg.get('lfs_size_per_node', 0)
                               }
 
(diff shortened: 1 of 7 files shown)


src/radical/pilot/agent/rm/fork.py
```diff
@@ -38,7 +38,12 @@ def _configure(self):
...
         self.cores_per_node = self._cfg.get('cores_per_node', self.requested_cores)
         self.gpus_per_node  = self._cfg.get('gpus_per_node', 0)  # FIXME GPU
-        self.lfs_per_node   = {'path' : self._cfg.get('lfs_path_per_node', None),
+
+        lfs_path = self._cfg.get('lfs_path_per_node', None)
+        if '$' in path:
+            lfs_path = os.path.expandvars(path)
+
+        self.lfs_per_node   = {'path' : lfs_path,
                                'size' : self._cfg.get('lfs_size_per_node', 0)
```
andre-merzky: Fork'ed CUs will always have the same env as the agent, so this is not needed here.  The problem should only occur on  launch methods which define a hop command, and where that hop command does not forward the agent env.

src/radical/pilot/agent/rm/ccm.py
```diff
@@ -54,7 +54,12 @@ def _configure(self):
         # Some simple arithmetic
         self.cores_per_node = ccm_nodes_length / ccm_node_list_length
         self.gpus_per_node  = self._cfg.get('gpus_per_node', 0) # FIXME GPU
-        self.lfs_per_node   = {'path' : self._cfg.get('lfs_path_per_node', None),
+
+        lfs_path = self._cfg.get('lfs_path_per_node', None)
+        if '$' in path:
```
andre-merzky: This check is not needed - the `expandvars` implementation does the same check.

andre-merzky: You fetch `lfs_path` but check and expand `path`?

src/radical/pilot/agent/rm/ccm.py
```diff
@@ -54,7 +54,12 @@ def _configure(self):
...
+            lfs_path = os.path.expandvars(path)
```
andre-merzky: The dict getter defaults to `None`, but:
```
>>> os.path.expandvars(None)
  File "/usr/lib/python2.7/posixpath.py", line 287, in expandvars
    if '$' not in path:
TypeError: argument of type 'NoneType' is not iterable
```

I guess this should  be either one  of these:

```
lfs_path = self._cfg.get('lfs_path_per_node', None)
if lsf_path: 
    lfs_path = os.path.expandvars(lsf_path)
```

or:

```
lfs_path = self._cfg.get('lfs_path_per_node', '')
lfs_path = os.path.expandvars(lsf_path)
```

I like the second a little more, but either one  works.

iparask: Okay! Fixing!

src/radical/pilot/agent/rm/fork.py
iparask: Okay! I will remove it!

andre-merzky:
Thanks!

PR #1780, opened 2018-12-21
Fix/lfs scheduling

andre-merzky:
fix lfs scheduling for multi-process, non-mpi units

@vivek-bala : Sorry that the commit is so large for the fix - I also did some code cleanup on the scheduler which was on my TODO list for a long while already...

@iparask : Ioannis, I added you also as reviewer for the code restructuring and pep-8-ification - thanks!


andre-merzky:
Yeah, that's the problem with linters; syntactically pylint is right, if `self.nodes` is empty, then `node` is undefined.  Semantically though that is impossible, on many levels.  I added a line to make the linter happy.

Toward master because its a fix and will result in a quick bugfix release, and will not wait to trickle into master through the default release cycle via devel.

Thanks, Andre.

andre-merzky:
>  I would try to correct all of them if possible.

I corrected two more, but I think following those *reduces* readability instead of increasing it, so I'd rather leave them...

iparask:
We still need changes. I made the mistake not to include all the files you changed! I was confused because based on the development guidelines - specifically [Branching-Model](https://github.com/radical-cybertools/radical.pilot/wiki/Branching-Model) - anything branched out from ` master` is a hotfix.

I do not agree that `_` reduces readability, but neither it increases it. So let's have the linter happy. If there is a rule that makes you uncomfortable or you think is redundant, we can easily remove it. Either way these are the new complaints from `flake8` and `pylint` for the changed files **only** (Reminder: I am using the config files we have for the testing branch):
```
(linting) iparask@DESKTOP-R64I4QR:/mnt/d/Git/RADICAL/radical.pilot$ flake8 src/radical/pilot/agent/scheduler/*.py
src/radical/pilot/agent/scheduler/continuous.py:305:18: W504 line break after binary operator
src/radical/pilot/agent/scheduler/continuous.py:406:9: F841 local variable 'total_procs' is assigned to but never used

(linting) iparask@DESKTOP-R64I4QR:/mnt/d/Git/RADICAL/radical.pilot$ pylint src/radical/pilot/agent/scheduler/*.py
Using config file /mnt/d/Git/RADICAL/radical.pilot/.pylintrc
************* Module radical.pilot.agent.scheduler.base
W:258, 8: Attribute '_lrms_node_list' defined outside __init__ (attribute-defined-outside-init)
W:278, 8: Attribute '_slot_lock' defined outside __init__ (attribute-defined-outside-init)
W:259, 8: Attribute '_lrms_cores_per_node' defined outside __init__ (attribute-defined-outside-init)
W:277, 8: Attribute '_wait_lock' defined outside __init__ (attribute-defined-outside-init)
W:276, 8: Attribute '_wait_pool' defined outside __init__ (attribute-defined-outside-init)
W:255, 8: Attribute '_pilot_id' defined outside __init__ (attribute-defined-outside-init)
W:257, 8: Attribute '_lrms_lm_info' defined outside __init__ (attribute-defined-outside-init)
W:261, 8: Attribute '_lrms_lfs_per_node' defined outside __init__ (attribute-defined-outside-init)
W:260, 8: Attribute '_lrms_gpus_per_node' defined outside __init__ (attribute-defined-outside-init)
W:256, 8: Attribute '_lrms_info' defined outside __init__ (attribute-defined-outside-init)
************* Module radical.pilot.agent.scheduler.continuous
W:373,12: Unused variable 'p' (unused-variable)
W:375,16: Unused variable 't' (unused-variable)
W:406, 8: Unused variable 'total_procs' (unused-variable)
W:149, 8: Attribute '_scattered' defined outside __init__ (attribute-defined-outside-init)
W:140, 8: Attribute '_oversubscribe' defined outside __init__ (attribute-defined-outside-init)
```

I am not sure what to do with the `attribute-defined-outside-init` warning. Should we keep it or not? Are these global variables? If yes, why are they defined as such?

andre-merzky:
The base class, `ru.process`, has a two-step initialization: after `fork()`, some parent and child initializers are called which define private variables.  It dies not make sense to declare those in `__init__`, and thus the warning is not appropriate in this context.  In other context, the warning may or may not be useful, I am not sure.

About `_`: I don't see much difference in
```
for i in range(n):
   for j in range(m):
    run()
```
and
```
for _ in range(n):
   for _ in range(m):
    run()
```
but habitually use the explicit variable version, as that is not problematic in all other languages I know.  I do not know why PEP.8 does not like it, I actually think it is not considering the loop context at all in these cases.  It's not much of an effort to change to `_`, but I don't consider this important in those cases.

Does that help?  I am not sure I understand what specific changes you propose to apply, please do let me know.

andre-merzky:
This depends on [this RU PR](https://github.com/radical-cybertools/radical.utils/pull/152)

mturilli:
The RU PR has been merged. Is this ready to be merged now?

andre-merzky:
This is hanging on the linter complains - it (mostly) complains about code which is actually correct but not liked *sigh*  

I'll try to take another look at this, not sure when though.

mturilli:
We need to convert towards a shared set of rules for the linter so that we can all use the same code style guidelines. @iparask the ball is on your court as @andre-merzky asked: 

> Does that help? I am not sure I understand what specific changes you propose to apply, please do let me know.

iparask:
Since it is for valid code, then let's change the rules we check. I agree we need to convert to a single file and nothing else, but also make sure we do not remove rules, because we do not want to make changes. 

In the discussion about for loops, I think it checks whether a loop index is used in a loop and if not it complains. Still both cases are valid. I will not hold it, but we should sit down and discuss the rules we have remove from `flake8` and `pylint` source files. They are already too much and I am afraid some of them are because we did not want to deal with them.

mturilli:
I agree we should not remove too much from flake8 because, otherwise, what do we use it for anyway?! Also, I would suggest we should not get too hung over syntax: you use i, you use _, who cares?! Let's try to have a bit of flexibility and see the bigger picture: having a well-defined set of coding rules helps to keep the code base consistent, lower the barrier of collaboration and, very importantly, the entry barrier for new coders. My suggestion is to abide by default to what flake8 wants but in cases in which it is not possible of that would require writing more code (I did not say different) than without it. Would everyone agree?

andre-merzky:
For the for loop, the following seems somewhat pythonic:
```python
for i in range(10):
    print i

for _ in range(10):
    print 'foo'
```
We allow '_' as var already in the linter.  Would that be agreeable?

For the exposure of private variables to other classes and methods: I think the linter complains  rightly, but I also don't want to change that right now, as it would be somewhat invasive.  I'd be in favor of improving the state in that respect over time.  Would that be acceptable?

Are there any other issues in the context of this PR?

iparask:
Yes, and no issues in the context of the PR.

andre-merzky:
@mturilli : by now we have a set of rules which is, IMHO, close enough to pep8.  The exceptions are mostly around my coding habits (alignment) which, I am terribly happy to say, are at least somewhat accepted as being my spleen.  The other rule changes are owed to our code constraints, like that we use two-step class initialization in several places, or that we have creator class methods, etc - and I agree with you, that those rule changes are better than more (and more convoluted) code.

At this point it is more about cleanup and consistency, which will happen over time....

andre-merzky:
@iparask : cool, then let me fix the loop warnings, and we merge.  Thanks!

mturilli:
Thanks. In order for this to 'happen over time' we need some rules in place. The 'it will happen before or later' is equivalent to say 'it will not happen' :)  I propose:
- PR are not merge if Travis tests do not pass.
- using Travis so that the NEW code is consistent with the more stringent guidelines of our linter. 
- Clean up all the OLD code when transitioning to Python 3.

Would this work for everyone? If so, @iparask could you put the test in place in Travis?

andre-merzky:
I can't fully agree, Matteo:

> PR are not merge if Travis tests do not pass.

They are not passing, and making it pass means fixing **all** complains.  

My policy at the moment (and for quite a while really) is to leave code better than I found it whenever I touch a file.  That tends to convolute PRs somewhat (as it tends to improve code in lines which are not directly related to a PR), but chunk by chunk code improves, and we avoid huge PRs with no semantic changes which are basically impossible to review.  Such efforts are best left to major transitions - the Py3 change is a good opportunity to do major cleanups.

All in all we should not get hung up too much on travis and flake complains IMHO - the code is in decent shape and fairly consistent, and we should invest energy on documentation test coverage etc. instead of moving spaces and variable names around (those are the majority of complains, really).    The main stumbling blocks for coders are not those, but complexity and documentation which are not captured by PEP8...

mturilli:
I agree we should avoid fixing all complains, this is why I write 'NEW code'. The goal here is not to check whether the whole code is pep/flake consistent, it is instead to check the the new code in a PR is consistent. In this way, there is a path to the adoption of a shared and consistent coding style. Travis has to pass in order to merge the PR, assuming the scope of its tests are LIMITED to the PR.

andre-merzky:
Yeah, that's the problem: travis checks all or nothing...

mturilli:
https://medium.com/@joey_9999/how-to-only-lint-files-a-git-pull-request-modifies-3f02254ec5e0

andre-merzky:
Nice - if that can be applied in travis, why not...  It still lints whole files, but lets see how that goes...

mturilli:
Great, thanks @andre-merzky. @iparask would you be happy to give this a go? If so, I am going to open a ticket for it.

iparask:
Yes, please do!

PR #1781, opened 2018-12-21
Feature/mem aware scheduler

vivek-bala:
This branch is branched off fix/lfs_scheduling, so it keeps some of the changes @andre-merzky  did in the fix/lfs_scheduling branch.

We extend the continuous scheduler to account for memory available on each node. We extend the lfs implementation to memory as well and carry the same assumptions (i.e. available memory is statically defined in the config file; memory available is the same on all nodes; once a task on a node ends, we assume the memory is completely freed on that node).


itomaldonado:

vivek-bala:
FYI This branch is for the qcarchive project which requires a memory aware scheduler.

vivek-bala:
I will be extending the 'memory' key to other resource configs very soon.

vivek-bala:
@andre-merzky let me know if you want to redirect the PR towards fix/lfs_scheduling since I branched off that.

iparask:
This is very nice indeed! I think it will be very useful. I have two questions though: 1) If the user does not specify the memory a unit requires, does it assume some kind of distribution? 2) Almost all systems have large memory nodes. Is there a plan to make sure that the scheduler takes the correct value?

vivek-bala:
1) Currently, if no memory is requested for a unit. We assume memory required to be 0. The scheduler then only schedules based on cores, gpus, and lfs requirements.
2) I am not entirely sure about this since I haven't used them. But aren't the large mem nodes available through exclusive queues? We currently assume that the amount of memory available on a node is static all nodes that we acquire. But if there is a possibility of getting, say, small and large memory nodes, then we will have to modify some elements in RP.

iparask:
I think this is a similar issue as the number of cores per node, which can change per queue, or the number and type of GPUs per node.

I am not objecting to what you are proposing, I am saying that we need to find a way to monitor all these details.

vivek-bala:
Yea, I completely agree with that. 

On the same lines, I think there is quite a bit of scope to simplify the scheduler code and make it even more extensible to any other scheduling metrics. Instead of individual lines of code for cpu, gpu, lfs, mem, we can put all the resource requirements into a dict, and iterate through it. Just a suggestion.

andre-merzky:
> @andre-merzky let me know if you want me to redirect the PR towards fix/lfs_scheduling since I branched off that.

If you don't mind, please do that for now, to keep the diff smaller - thanks.

vivek-bala:
Done

andre-merzky:
Looks good to me.  I  think moving to a dict *may* shorten the code, and make it much less readable: LFS is a dict, mem is a scalar, gpus and cores are sets of `BUSY` and `FREE`.

vivek-bala:
Can you elaborate on what you mean by resource stypes? I think at some point we had a discussion about a scheduler class (derived from just cpu based continuous scheduler (CS)?) might be the way to go. But I think that might involve a lot of duplicated code (for example, between cpu+gpu CS and cpu+gpu+lfs CS). Wasn't sure if you were referring to the extensibility aspect, but just wanted to mention that.

I am happy to move to a different CS file but I see no way to do it without duplicating quite a bit of code.

iparask:
I am not sure we can avoid the resource types the scheduler takes into account. With so many diverse use case requirements in all aspects ( CPUs, GPUs, Filesystem, Memory, other accelerators) it becomes difficult not to take them into account.

Would it make sense to decide on a set of standard resource types and have the rest as plug and play based on the pilot description?

andre-merzky:
@iparask : Sure we need to use those types - but I ponder if we should have more schedulers instead of putting them all into one scheduler, which becomes complex and slow over time...

iparask:
No, I do not. I will continue the discussion in the ticket.

iparask:
I would like to add that the value which is added in the config file cannot be the actual memory of the node. That is not the total available since there is the operating system that uses a part of it, along with anything else that might be needed by the node. For example, if a  node has 128GBs of RAM and since it runs some flavor of OS, we should probe and see how much usually the operating system consumes, subtract it from the provided value based on documentation and provide that in the configuration file. I would suggest a bit less actually as a safety feature to reduce the probability of actually filling up the memory of the node.

vivek-bala:
Yea, I agree with that. I think your statement holds true for both available lfs and memory on a node. Currently, they are statically specified in the config file. I think we are leaving the responsibility of the specification (and its correctness) to the user, which is okay from my perspective for now. In the meantime, we can think of the best/correct ways to probe for the correct values (which may change over time within the walltime). Hopefully, we can avoid having a different probing mechanism for each parameter (lfs, memory, etc.). 

I would like to keep that out of the scope of this PR though as it requires more discussion. Is that okay?

mturilli:
Following a discussion with @iparask about this PR and the summary I posted on #1782, we agreed that this PR should not be blocked and that, at the moment, we also agree with @vivek-bala about giving to the user the responsibility of indicating a viable amount of memory in the resource configuration file. In this way, we split the development of the scheduling capability from that of profiling memory availability at runtime. Note that, if we do not know how much memory we really have available, including when running the agent on the same node in which we execute CU, the actual usefulness of this 'scheduler' is limited. @andre-merzky can we create a 'roadmap' document about memory-related capabilities in RP, add it to the wiki, and discuss it in our next RTC call?

src/radical/pilot/agent/rm/base.py
```diff
@@ -83,6 +83,7 @@ def __init__(self, cfg, session):
         self.cores_per_node  = None
         self.gpus_per_node   = None
         self.lfs_per_node    = None
+        self.mem_per_node    = None
```
mturilli: We assume each node has the same amount of memory. Note that this assumption breaks when we run the agent on one of the working nodes. Depending on the total amount of memory installed, the amount of memory actually available, and the agent's memory footprint, this break of assumption might be relevant.

src/radical/pilot/agent/scheduler/continuous.py
```diff
@@ -420,7 +429,8 @@ def _alloc_nompi(self, cud):
                    total_gpus, self._lrms_gpus_per_node)
```
mturilli: Do we have a problem with the indentation of the previous line (428)? (it may be an issue with github diff mode)

mturilli:
Re the discussion about the design of our scheduler(s) (code duplication, performance, modularity, composability, maintainability, etc.), @andre-merzky would you be happy to put together a design proposal, put it in the wiki and schedule it for discussion in our RCT (dev) meeting? IMHO, this would be a starting point to which both @iparask and @vivek-bala would contribute until we can define a roadmap for the implementation.

src/radical/pilot/agent/rm/base.py
vivek-bala: I agree. But as a starting point, I think we can start with this assumption. We can relax this assumption when we have some probing mechanism in place. Would that be okay?

src/radical/pilot/agent/scheduler/continuous.py
vivek-bala: I don't think there is any indentation issue here. Checked in code.

andre-merzky:
DIscussed on RCT call, agree to merge.  Careful to document this is a scheduler hint, not a promise.  System might not provide what is documented.

andre-merzky:
Thanks all for the discussion!

@vivek-bala : please open  a new branch (or reopen the same branch again) if you want to add documentation.

PR #1792, opened 2019-01-24
Feature/summit am

andre-merzky:
KISS ;-)

I know its in the eye of the beholder though.

### DIFF
--- a/src/radical/pilot/agent/lm/jsrun.py
+++ b/src/radical/pilot/agent/lm/jsrun.py
@@ -69,33 +69,22 @@ class JSRUN(LaunchMethod):
 
         """
 
-        node_ids  = list()
-        core_maps = list()
-        gpu_maps  = list()
-
-        for node in slots['nodes']:
-            node_ids += [node['uid']] * len(node['core_map'])
-
-        core_maps = [core_map for core_map in node['core_map'] for node in slots['nodes']]
-        gpu_maps  = [gpu_map  for gpu_map  in node['gpu_map']  for node in slots['nodes']]
-
         rs_id  = 0
         rs_str = ''
 
-        if len(gpu_maps):
-            for node_id,core_ids,gpu_ids in zip(node_ids, core_maps, gpu_maps):
-                rs_str += 'RS %d : { ' % rs_id
-                rs_str += 'host: %d, ' % node_id
-                rs_str += 'cpu: %s, '  % ' '.join(map(str, core_ids))
-                rs_str += 'gpu: %s}\n' % ' '.join(map(str, gpu_ids))
-                rs_id  += 1
-        else:
-            for node_id,core_ids in zip(node_ids, core_maps):
-                rs_str += 'RS %d : { ' % rs_id
-                rs_str += 'host: %d, ' % node_id
-                rs_str += 'cpu: %s '   % ' '.join(map(str, core_ids))
-                rs_str += '}\n'
-                rs_id  += 1
+        for node in slots['nodes']:
+
+            cores = ' '.join([str(core_set[0]) for core_set
+                                               in  node['core_map']])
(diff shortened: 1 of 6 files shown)


src/radical/pilot/agent/lm/jsrun.py
```diff
@@ -27,85 +27,82 @@ def _configure(self):
...
+        for node in slots['nodes']:
+
+            cores = ' '.join([str(core_set[0]) for core_set
+                                               in  node['core_map']])
+            gpus  = ' '.join([str(gpu_set[0])  for gpu_set
+                                               in  node['gpu_map']])
+
+            rs_str           += 'RS %d: {'  % rs_id
+            rs_str           += ' host: %d' % node['uid']
+            if cores: rs_str += ' cpu: %s'  % cores
```
vivek-bala: Not sure if we should have an 'if' condition for cores. Two reasons:

* I am not sure if jsrun will work with a layout file which does not specify a cpu to use.
* Under our constraint that using gpus requires at least one cpu, we will always request at least one cpu.

src/radical/pilot/agent/rm/lsf_summit.py
```diff
@@ -172,83 +174,77 @@ def __init__(self, cfg, session):
...
-            raise RuntimeError(msg)
-
-        # parse LSF hostfile
-        # format:
-        # <hostnameX>
-        # <hostnameX>
-        # <hostnameY>
-        # <hostnameY>
+        if not lsf_hostfile:
+            raise RuntimeError("$LSB_DJOB_HOSTFILE not set!")
```
vivek-bala: I think this writes the error/exception to stderr whereas previously we wrote to the
log file. Is the change intended?  Should we stick to one over the other in all
components/subcomponents?

src/radical/pilot/agent/lm/jsrun.py
andre-merzky: True for summit - I am not sure for other systems.

The decision if GPUs processes require a CPU core is made in the scsheduler,  and it is configurable for both options.   So the executor should not reason about it, just enact it.

src/radical/pilot/agent/rm/lsf_summit.py
andre-merzky: It is caught in the agent when the RP is constructed,  and will be logged there, including the exception trace.  Exception should be used as always, and be logged where they are caught (and they should *always* we caught - if not, we need to fix that),  Error logs near the raise make sense if it gives additional information - but not so much if we end up with the same message twice.

src/radical/pilot/agent/lm/jsrun.py
```diff
@@ -116,21 +113,16 @@ def construct_command(self, cu, launch_script_hop):
...
-        #flags = '-n%d -a1 ' % (task_procs)
-        command = '%s %s %s %s' % (self.launch_command, flags,
-                                            env_string, 
-                                            task_command)
+        env_string = ' '.join(['-E "%s"' % var for var in env_list])
+
+        rs_fname = self._create_resource_set_file(slots=slots, uid=cu['uid'],
+                                                  sandbox=task_sandbox)
+
+      # flags = '-n%d -a1 ' % (task_procs)
```
iparask: I think this can be removed completely

src/radical/pilot/agent/rm/lsf_summit.py
iparask: Why is the check for `LSB_MCPU_HOSTS` removed?

andre-merzky:
> There has been major refactoring especially in LRMS. Are the tests still passing?

buzzkill! :-P

src/radical/pilot/agent/lm/jsrun.py
vivek-bala: Ah yes I forgot about the over subscribe config. Thanks.

PR #1793, opened 2019-01-25
This optimization breaks for tagged units - remove.

andre-merzky:
@iparask : please make sure this does the trick.  Thanks!

### DIFF
--- a/src/radical/pilot/agent/scheduler/base.py
+++ b/src/radical/pilot/agent/scheduler/base.py
@@ -606,15 +606,6 @@ class AgentSchedulingComponent(rpu.Component):
                 with self._wait_lock:
                     self._wait_pool.remove(unit)
 
-            else:
-                # Break out of this loop if we didn't manage to schedule a task
-                # FIXME: this assumes that no smaller or otherwise more suitable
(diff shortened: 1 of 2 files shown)


iparask:
As soon as, I run a test I will let you know! Thank you @andre-merzky

andre-merzky:
Can you please check why the update worker did not come up?

iparask:
I think it was a system issue. The second test started but the first unit did not get scheduled. The last state was `AGENT_SCHEDULING_PENDING`.

andre-merzky:
Sorry Ioannis, the information from the last message is not enough to understand whats up.  Do you have a  set if CUDs and PDs to reproduce this?

iparask:
The Pilot description had  64 cores and 4 GPUs it was submitted at the GPU queue of Bridges. This stopped at the very first unit which is a single core unit that was submitted alone.

andre-merzky:
Can you attach the logs?  This sounds unrelated to this PR, really, doesn't it?  Any errors in the agent logs?

andre-merzky:
Thanks

PR #1797, opened 2019-01-30
Feature/summit am

andre-merzky:
This replaces the blanket `num_procs` mechanism of the cont scheduler with a slightly better version.  It's likely still broken for some gpu placement requests.


andre-merzky:
This PR needs more work before review...

mturilli:
Is this still valid?

andre-merzky:
Yes, this is actively worked on.

andre-merzky:
This PR is now good for review.

mturilli:
@andre-merzky to add some comments to facilitate understanding of the PR

docs/architecture/concepts/re_scheduler.py
```diff
@@ -0,0 +1,111 @@
+#!/usr/bin/env python
```
andre-merzky: This is a proof of concept code which I thought would be a candidate for a simpler cont scheduler.  It is not fit for using, yet, but I left it as a place holder for future work.

report/index.html
```diff
@@ -1,4 +0,0 @@
-<body>
```
andre-merzky: This should have been cleaned out ages ago...

src/radical/pilot/agent/agent_0.py
```diff
@@ -325,25 +325,21 @@ def _start_sub_agents(self):
                 #        work with a number of launch methods.  Can the
                 #        offset computation be moved to the LRMS?
                 ls_name = "%s/%s.sh" % (os.getcwd(), sa)
-                slots = {
```
andre-merzky: The sub-agent startup went out of sync with the scheduler changes.  This fix is needed for running jsrun from the compute nodes.

src/radical/pilot/agent/bootstrap_0.sh
```diff
@@ -176,7 +176,7 @@ profile_event()
 {
     PROFILE="bootstrap_0.prof"
 
-    if test -z "$RADICAL_PILOT_PROFILE"
+    if test -z "$RADICAL_PILOT_PROFILE$RADICAL_PROFILE"
```
andre-merzky: Forward compatibility to the backward compatibility of the upcoming release ;-)

src/radical/pilot/agent/bootstrap_0.sh
```diff
@@ -1058,19 +1058,8 @@ virtenv_create()
...
-            wheeled="--no-binary :all:"
-        else
-            wheeled=""
-        fi
-
         run_cmd "install $dep" \
-                "$PIP install $wheeled $dep" \
+                "$PIP install $dep" \
              || echo "Couldn't install $dep! Lets see how far we get ..."
     done
```
andre-merzky: This is not needed on titan anymore and in fact breaks there.  Both titan and summit can use the default code path here.

src/radical/pilot/agent/executing/popen.py
```diff
@@ -213,8 +213,7 @@ def _handle_unit(self, cu):
                             % (str(e), traceback.format_exc())
 
             # Free the Slots, Flee the Flots, Ree the Frots!
-            if cu.get('slots'):
-                self.publish(rpc.AGENT_UNSCHEDULE_PUBSUB, cu)
+            self.publish(rpc.AGENT_UNSCHEDULE_PUBSUB, cu)
```
andre-merzky: We always have slots here - otherwise the unit would not make it into executing state.

src/radical/pilot/agent/lm/jsrun.py
```diff
@@ -77,9 +77,10 @@ def _create_resource_set_file(self, slots, uid, sandbox):
         rs_str = 'cpu_index_using: physical\n'
         rank = 0
         for node in slots['nodes']:
+
             gpu_maps = list(node['gpu_map'])
             for map_set in node['core_map']:
-                cores = ','.join(str(core * 4) for core in map_set)
+                cores = ','.join(str(core) for core in map_set)
```
andre-merzky: This is now captured via SAGA's SMT settings.

src/radical/pilot/agent/rm/base.py
```diff
@@ -18,7 +18,7 @@
 RM_NAME_TORQUE      = 'TORQUE'
 RM_NAME_YARN        = 'YARN'
 RM_NAME_SPARK       = 'SPARK'
-
+RM_NAME_DEBUG       = 'DEBUG'
```
andre-merzky: I used the DEBUG rm for, well, debugging - it can fake a summit like node layout on localhost.  Its not complete, but already useful...

src/radical/pilot/agent/rm/lsf_summit.py
```diff
@@ -58,6 +59,7 @@ def __init__(self, cfg, session):
         self.cores_per_socket   = None
         self.gpus_per_socket    = None
         self.lfs_per_node       = None
+        self.smt                = int(os.environ.get('RADICAL_SAGA_SMT', 1))
```
andre-merzky: Pick up SMT settings from SAGA, and use for core indexing.

src/radical/pilot/agent/rm/lsf_summit.py
```diff
@@ -131,46 +134,52 @@ def __init__(self, cfg, session):
...
+                'node_list'        : self.node_list,
+                'sockets_per_node' : self.sockets_per_node,
+                'cores_per_socket' : self.cores_per_socket * self.smt,
+                'gpus_per_socket'  : self.gpus_per_socket,
+                'cores_per_node'   : self.sockets_per_node * self.cores_per_socket * self.smt,
+                'gpus_per_node'    : self.sockets_per_node * self.gpus_per_socket,
+                'agent_nodes'      : self.agent_nodes,
+                'lfs_per_node'     : self.lfs_per_node,
+                'smt'              : self.smt
+                }
```
andre-merzky: add smt info here for the scheduler to use.

src/radical/pilot/agent/scheduler/base.py
```diff
@@ -18,16 +18,18 @@
...
 SCHEDULER_NAME_CONTINUOUS_ORDERED = "CONTINUOUS_ORDERED"
 SCHEDULER_NAME_CONTINUOUS_SUMMIT  = "CONTINUOUS_SUMMIT"
+SCHEDULER_NAME_CONTINUOUS         = "CONTINUOUS"
 SCHEDULER_NAME_HOMBRE             = "HOMBRE"
-SCHEDULER_NAME_SCATTERED          = "SCATTERED"
 SCHEDULER_NAME_SPARK              = "SPARK"
 SCHEDULER_NAME_TORUS              = "TORUS"
 SCHEDULER_NAME_YARN               = "YARN"
 
+# SCHEDULER_NAME_CONTINUOUS_FIFO    = "CONTINUOUS_FIFO"
```
andre-merzky: These two are obsolete by now.

src/radical/pilot/agent/scheduler/base.py
```diff
@@ -393,42 +395,21 @@ def slot_status(self):
         Returns a multi-line string corresponding to the status of the node list
         '''
 
+        glyphs = {rpc.FREE : '-',
+                  rpc.BUSY : '#',
+                  rpc.DOWN : '!'}
```
andre-merzky: Additional to `FREE` and `BUSY`, we add `DOWN` to mark cores we cannot use for some reason.  On summit, some cores cannot be addressed via ERF right now.

src/radical/pilot/agent/scheduler/continuous.py
```diff
@@ -267,67 +292,68 @@ def _find_resources(self, node, requested_cores, requested_gpus,
...
-            alloc_gpus = num_procs * gpu_chunk
+            alloc_lfs = min(requested_lfs, free_lfs)
+            num_procs['lfs'] = alloc_lfs / lfs_chunk
+
+
+        # Find normalized cores, gpus and lfs
+        if requested_cores: alloc_cores = num_procs['cores'] * core_chunk
+        if requested_gpus : alloc_gpus  = num_procs['gpus']  * gpu_chunk
+        if requested_lfs  : alloc_lfs   = num_procs['lfs']   * lfs_chunk
+        self._log.debug('alc : %s %s %s', alloc_cores, alloc_gpus, alloc_lfs)
```
andre-merzky: The above was wrong before, for non-uniform processes.  Well, now we agreed that processes *are* uniform, but our CUD does not yet enforce this, so this fix is still needed.

what happened was this:

  - assume we have 2 nodes with 42 cores and  6 gpu each.
  - assume we want to place a unit which asks for 6 cpu process and 12 GPU processes (18 procs in total)
  - assume the nodes have each 9 cores available (18 cores)

We would expect a result like this (a GPU process always also needs a CPU core to run on):

  - node 0:  3 CPU cores + (6 CPU cores + 6 GPUs)
  - node 1:  3 CPU cores + (6 CPU cores + 6 GPUs)

What happened is that the scheduler checked node 0, found space for 3 CPU procs and 6 GPU procs, computed the smalles common denominator (3), and placed 3 CPU procs and 3 (!) GPU procs (1 core + 1 GPU).  3 cores and 3 GPUs were left idle.  The second node was utilized the same way - but the remaining 6 GPU processes could not be placed anymore (out of nodes).

This is fixed now: the least common denominator is not computed anymore, chunking is done per resource type.

src/radical/pilot/configs/agent_summit.json
```diff
@@ -0,0 +1,48 @@
+{
```
andre-merzky: This agent config places the executor on a compute node.

src/radical/pilot/utils/misc.py
```diff
@@ -88,12 +82,13 @@ def hostip(req=None, black_list=None, pref_list=None, logger=None):
     # Known intefaces in preferred order
     if not pref_list:
         pref_list = [
-            'ipogif0', # Cray's
-            'br0'      # SuperMIC
+            'ipogif0',  # Cray's
+            'ib0',      # IBM (Summit)
+            'br0'       # SuperMIC
```
andre-merzky: This fix is needed to make the ZMQ bridges accessible from the compute nodes.  This needs to go into some configuration file eventually.

andre-merzky:
@mturilli , @iparask : comments have been added.  Everything else is basically code cleanup towards pylint checks.

docs/architecture/concepts/re_scheduler.py
```diff
@@ -0,0 +1,111 @@
...
+cross_sockets = False  # can threads   cross sockets?
+cross_nodes   = False  # can processes cross nodes?
+
+
+# In the general case, we search for a mix of GPUs and CPUs, with the
+# following constraints: 
+#
+#   - CPU cores for threads which belong to the same process need to be near
+#     each other  (on the same node or even the same core)
+#   - GPUs need to be near to the processes usinng them  (on the same node or
```
iparask:
```suggestion
#   - GPUs need to be near to the processes using them  (on the same node or
```

docs/architecture/concepts/re_scheduler.py
```diff
@@ -0,0 +1,111 @@
...
+          + '| _###_###### +-+ : #__###__##_ -++ ' \
+          + '| #######_#__ +++ : _##__##_### ++- ' \
+          + '| #####_#_##_ --+ : ##_##__#### +-+ ' \
+          + '|'
+
+tested = '((?=((?:(?:_[^:\|]*?){3})[^\|]*?){2})(?:(?:.*?(?:-)){2}))'
+print tested
+print pattern
+print nodes
+print
```
iparask: I propose to remove this.

src/radical/pilot/agent/bootstrap_0.sh
iparask: This test if any of the two variables exist, right?

src/radical/pilot/agent/lm/jsrun.py
```diff
@@ -110,18 +110,19 @@ def construct_command(self, cu, launch_script_hop):
         task_argstr  = self._create_arg_string(task_args)
         task_sandbox = ru.Url(cu['unit_sandbox']).path
 
-        self._log.debug('prep %s', cu['uid'])
+        assert(slots), 'missing slots for %s' % uid
```
iparask: May I suggest something like:
```suggestion
        if not slots:
            raise raise AssertionError( 'missing slots for %s' % uid)
```
As far as I know, asserts can be closed completely, so it might make sense to test data existence and raise exceptions.

src/radical/pilot/agent/scheduler/continuous.py
```diff
@@ -161,18 +161,41 @@ def _configure(self):
...
+                #
+                # This is related to the known issue listed on
+                # https://www.olcf.ornl.gov/for-users/system-user-guides \
+                #                          /summit/summit-user-guide/
+                #
+                # "jsrun explicit resource file (ERF) allocates incorrect
+                # resources"
+                #
+                smt = self._lrms_info.get('smt', 1)
+                for s in [1]:
```
iparask: This just does one iteration, right? Why is the for loop needed?

src/radical/pilot/agent/scheduler/continuous.py
```diff
@@ -161,18 +161,41 @@ def _configure(self):
...
+                    for i in range(smt):
+                        idx = s * 21 * smt + i
+                        node_entry['cores'][idx] = rpc.DOWN
```
iparask: This makes sure that the cores used by the system are not considered for placing, right?

src/radical/pilot/agent/scheduler/continuous.py
```diff
@@ -182,34 +205,37 @@ def _try_allocation(self, unit):
...
-            unit['slots'] = self._allocate_slot(unit['description'])
+            self._prof.prof('schedule_try', uid=uid)
+            unit['slots'] = self._allocate_slot(unit)
 
         # the lock is freed here
         if not unit['slots']:
 
-            # signal the unit remains unhandled (False signals that failure)
-            self._prof.prof('schedule_fail', uid=unit['uid'])
+            # signal the unit remains unhandled (Fales signals that failure)
```
iparask: Is this a typo?

src/radical/pilot/agent/scheduler/continuous.py
```diff
@@ -318,21 +352,32 @@ def _find_resources(self, node, requested_cores, requested_gpus,
...
+        if requested_gpus : alloc_gpus  = num_procs['gpus']  * gpu_chunk
+        if requested_lfs  : lfs         = num_procs['lfs']   * lfs_chunk
+        if requested_mem  : mem         = num_procs['mem']   * mem_chunk
 
-        # Find normalized lfs, cores and gpus
-        if requested_cores: alloc_cores = nprocs * core_chunk
-        if requested_gpus : alloc_gpus  = nprocs * gpu_chunk
-        if requested_lfs  : lfs         = nprocs * lfs_chunk
-        if requested_mem  : mem         = nprocs * mem_chunk
+        self._log.debug('alc : %s %s %s %s', alloc_cores, alloc_gpus, lfs, mem)
```
iparask: This is a useful message, but there is no way to understand for which unit it is. Does it make sense to add the unit id as well?

src/radical/pilot/agent/scheduler/continuous_summit.py
```diff
@@ -292,11 +290,11 @@ def _change_slot_states(self, slots, new_state):
...
 
-            node = (n for n in self.nodes if n['uid'] == nodes['uid']).next()
-            assert(node)
+            for entry in self.nodes:
+                if entry['uid'] == node['uid']:
+                    break
 
-            # iterate over cores/gpus in the slot, and update state
-            cores = nodes['core_map']
+            assert(entry), 'missing node %s' % node['uid']
```
iparask: The same as the previous assert question.

src/radical/pilot/agent/scheduler/continuous_summit.py
```diff
@@ -292,11 +290,11 @@ def _change_slot_states(self, slots, new_state):
             #       that we would read, and keep a dictionary that maps the uid
             #       of the node to the location on the list?
 
-            node = (n for n in self.nodes if n['uid'] == nodes['uid']).next()
-            assert(node)
+            for entry in self.nodes:
```
iparask: I am not sure what this for loop does. @andre-merzky can you explain it a bit?

src/radical/pilot/agent/scheduler/continuous_summit.py
```diff
@@ -542,12 +556,15 @@ def _find_resources(self, node, requested_cores, requested_gpus,
 
                 # Determine maximum procs on each socket - required during assignment
                 max_procs_on_socket = [0 for _ in range(self._lrms_sockets_per_node)]
+                self._log.debug('mpos %s', max_procs_on_socket)
```
iparask:
```suggestion
                self._log.debug('maximum procs on socket: %s', max_procs_on_socket)
```

src/radical/pilot/agent/scheduler/continuous_summit.py
```diff
@@ -557,29 +574,23 @@ def _find_resources(self, node, requested_cores, requested_gpus,
...
-        if requested_lfs:
-            alloc_lfs = num_procs * lfs_chunk
-        if requested_cores:
-            alloc_cores = num_procs * core_chunk
-        if requested_gpus:
-            alloc_gpus = num_procs * gpu_chunk
+        if requested_cores: alloc_cores = num_procs['cores'] * core_chunk
+        if requested_gpus : alloc_gpus  = num_procs['gpus']  * gpu_chunk
+        if requested_lfs  : alloc_lfs   = num_procs['lfs']   * lfs_chunk
+        self._log.debug('alc : %s %s %s', alloc_cores, alloc_gpus, alloc_lfs)
```
iparask: It makes sense to have an allocation debug message. I think also that the ID should be added to make sure for which unit this allocation is.

src/radical/pilot/agent/scheduler/continuous_summit.py
```diff
@@ -856,48 +864,44 @@ def _alloc_mpi(self, cud):
...
+            node_uid  = node['uid']
             node_name = node['name']
 
             # If unit has a tag, check if the tag is in the tag_history dict,
             # else it is a invalid tag, continue as if the unit does not have
             # a tag
             # If the unit has a valid tag, find the node that matches the
             # tag from tag_history dict
-            if tag and (tag in self._tag_history.keys()):
+            if tag and tag in self._tag_history:
```
iparask: I'm not sure if checking if tag is not `None` and if it is in `tag_history` creates redundancy.
If tag is `None`, it would definitely not be in `tag_history`. Also, if tag has a value it can or cannot be in `tag_history`.

Isn't then:
```python
if tag in self._tag_history:
```
enough?

docs/architecture/concepts/re_scheduler.py
andre-merzky: its not in active code, but lives under `docs/` - I'll certainly remove the prints once it moves to `src/`.  Is that acceptable for the time being?

src/radical/pilot/agent/bootstrap_0.sh
andre-merzky: yes.

src/radical/pilot/agent/lm/jsrun.py
andre-merzky: Nah, then I rather take out the assert.  I don't want to start testing regularly for conditions which are only interesting when debugging.  That assertions can be disabled globally ( via `python -O` or `export PYTHONOPTIMIZE=1`) is a good thing, and exactly why I like to use them - they should not be part of the normal flow of operation.  Thoughts?

src/radical/pilot/agent/scheduler/continuous.py
andre-merzky: When testing core viability, I initially thought that the last physical core on socket 0 was unavailable also, the the loop thus initially was `for s in [0, 1]`.  I'll make this clearer.

src/radical/pilot/agent/scheduler/continuous.py
andre-merzky: correct - but it is not system use we are hiding, but a bug in slurm which invalidates addresses pointing to that core (see comment above the loop).

src/radical/pilot/agent/scheduler/continuous.py
andre-merzky: thanks, fixed.

src/radical/pilot/agent/scheduler/continuous.py
andre-merzky: The `find_resources` method does not know the uid - we'll have to derive that from the surrounding debug messages from the higherlevel scheduler calls.  Since that part of the scheduler code is sequential, this should always be possible.

src/radical/pilot/agent/lm/jsrun.py
andre-merzky: ping

docs/architecture/concepts/re_scheduler.py
andre-merzky: ping

src/radical/pilot/agent/scheduler/continuous.py
andre-merzky: I hope this is clearer now.

andre-merzky:
@iparask : I replied on all comments, this should be ready for another iteration.  Thanks.

src/radical/pilot/agent/lm/prte.py
```diff
@@ -0,0 +1,284 @@
...
+                os.kill(os.getpid())
+
+            logger.info('prte stopped (%d)' % dvm_process.returncode)
+        # ----------------------------------------------------------------------
+
+        dvm_watcher = ru.Thread(target=_watch_dvm, name="DVMWatcher")
+        dvm_watcher.start()
+
+        lm_info = {'dvm_uri'     : dvm_uri,
+                   'version_info': {name: prte_info}}
```
iparask: Isn't a `self.` missing from here?

src/radical/pilot/agent/lm/prte.py
```diff
@@ -0,0 +1,284 @@
...
+          # depth = depths.pop()
+          # if depth > 1: map_flag = '--bind-to none --map-by ppr:%d:core' % depth
+          # else        : map_flag = '--bind-to none'
+            map_flag  = ' --bind-to none -use-hwthread-cpus --oversubscribe'
+            map_flag += ' -n %d --cpus-per-proc %d' % (n_procs, n_threads)
+            map_flag += ' -host %s' % hosts_string
+
+
+        # Additional (debug) arguments to prun
+        if True or self._verbose:
```
iparask: This is always true, right? If this is the intended behavior, I think removing the `if` makes sense, otherwise removing `True` makes sense

src/radical/pilot/agent/rm/base.py
iparask: This is going to be extremely useful for integration tests now that I think about it! Do you think we can expand it for the rest of the RMs?

src/radical/pilot/agent/scheduler/archive/continuous_summit.py
```diff
@@ -128,15 +125,14 @@ def initialize_child(self):
         # during agent startup.  We dig them out of the config at this point.
         #
         # NOTE: this information is insufficient for the torus scheduler!
-        self._pilot_id = self._cfg['pilot_id']
-        self._lrms_info = self._cfg['lrms_info']
-        self._lrms_lm_info = self._cfg['lrms_info']['lm_info']
-        self._lrms_node_list = self._cfg['lrms_info']['node_list']
+        self._pilot_id              = self._cfg['pilot_id']
```
iparask: DIdn't we say that the `ContinuousSummit` scheduler is not needed? Am I forgetting something?

src/radical/pilot/agent/scheduler/archive/continuous_summit.py
```diff
@@ -557,29 +574,23 @@ def _find_resources(self, node, requested_cores, requested_gpus,
...
+            num_procs['gpus'] = alloc_gpus / gpu_chunk
 
         # Find normalized lfs, cores and gpus
-        if requested_lfs:
-            alloc_lfs = num_procs * lfs_chunk
-        if requested_cores:
-            alloc_cores = num_procs * core_chunk
-        if requested_gpus:
-            alloc_gpus = num_procs * gpu_chunk
+        if requested_cores: alloc_cores = num_procs['cores'] * core_chunk
```
iparask: I am a bit confused. Why do you divide it if you are going to multiply it at the end with `core_chuck` and render both operations unnecessary? Unless it is for rounding reasons, then I can see why it is needed

src/radical/pilot/agent/scheduler/base.py
```diff
@@ -217,6 +218,13 @@ def __init__(self, cfg, session):
         rpu.Component.__init__(self, cfg, session)
 
 
+        tmp = os.environ.get('RP_UNIFORM_WORKLOAD', '').lower()
```
iparask: Why do we care if the workload is uniform or not?

src/radical/pilot/agent/scheduler/base.py
```diff
@@ -653,6 +640,10 @@ def schedule_cb(self, topic, msg):
                 with self._wait_lock:
                     self._wait_pool.remove(unit)
 
+            else:
```
iparask: Why it should break if the workload is uniform?

src/radical/pilot/agent/lm/prte.py
```diff
@@ -0,0 +1,284 @@
...
+        if not stdbuf_cmd:
+            raise Exception("Couldn't find (g)stdbuf")
+        stdbuf_arg = "-oL"
+
+        # Base command = (g)stdbuf <args> + prte + prte-args + debug_args
+        prte_args = '--report-uri -'
+        cmdline   = '%s %s %s %s ' % (stdbuf_cmd, stdbuf_arg, prte, prte_args)
+
+        # Additional (debug) arguments to prte
+        if os.environ.get('RADICAL_PILOT_PRUN_VERBOSE'):
```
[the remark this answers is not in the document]
andre-merzky: I made this consistent with line 29 now - but indeed, the actual value does not matter at this point.  It's not a documented env var, and only used for debug settings...

src/radical/pilot/agent/lm/prte.py
andre-merzky: where?  But since this is a classmethod, we have no `self` in the first place.

src/radical/pilot/agent/lm/prte.py
andre-merzky: Ah, thanks - I removed the `True` ...  Thanks!

src/radical/pilot/agent/rm/base.py
andre-merzky: Yes, we should be able to expand it to replace all RMs in a test setup! :-)

src/radical/pilot/agent/scheduler/base.py
andre-merzky: Assume you have 1k uniform tasks.  You try to schedule task 1, and it fails.  There is no need to try to schedule task 2 under the same conditions, as it will fail as well.  Or task 3, ...  Knowing that the workload is uniform allows to optimize this, and we only test one task.

In the general case we don't know about uniformity, so this optimization should never apply - but for testing its useful, thus this undocumented env setting to enable this.

src/radical/pilot/agent/scheduler/base.py
andre-merzky: see above :-)  A new iteration cannot yield a different result.

src/radical/pilot/agent/lm/prte.py
iparask: Here: ` 'version_info': {name: prte_info}` I was not able to find where name was set up in the method. That is why I asked.

src/radical/pilot/agent/scheduler/base.py
iparask: Ha! I see. I did not think of that specific corner case!

src/radical/pilot/agent/lm/prte.py
iparask: @andre-merzky, if this is my mistake go forward and merge!

andre-merzky: Ah, I see - `name` came from line 35 in this case:
```
    @classmethod
    def lrms_config_hook(cls, name, cfg, lrms, logger, profiler):
```

But looking over it again, we don't actually need to use the name here, the `lm_info` dict is private anyway, so I simplified the dict.  Since that has no functional consequences (the dict entry is informative only), I'll go ahead and merge.

Thanks for your patience! :-)

iparask: :D Then it is done!

PR #1801, opened 2019-02-01
push node_list to mongodb for analytics

andre-merzky:
Push the node list to MongoDB, so that Analytics can evaluate the amount of available resources (the cu slots otherwise only document the amount of *used* resources).


src/radical/pilot/agent/bootstrap_0.sh
```diff
@@ -1527,7 +1527,7 @@ PB1_LDLB="$LD_LIBRARY_PATH"
 #        We should split the parsing and the execution of those.
 #        "bootstrap start" is here so that $PILOT_ID is known.
 # Create header for profile log
-if ! test -z "$RADICAL_PILOT_PROFILE"
+if ! test -z "$RADICAL_PILOT_PROFILE$RADICAL_PROFILE"
```
iparask: Why are you checking twice for the environment variable?

andre-merzky: This checks if either `RADICAL_PILOT_PROFILE` or `RADICAL_PROFILE` is set.  Both are valid by now (per RU changes), so either one should trigger profiling.

mturilli:
Is this ready to be merged?

andre-merzky:
This supports the prelim proposal data, but otherwise is not urgent.   We can probably put it into the next release.

PR #1802, opened 2019-02-03
prepare for saga rename

andre-merzky:
the expected match on RP level to the saga rename - but this one should be backward compatible to avoid breakage.

### DIFF
--- a/src/radical/pilot/agent/executing/shell.py
+++ b/src/radical/pilot/agent/executing/shell.py
@@ -124,8 +124,8 @@ class Shell(AgentExecutingComponent):
                                      # are not yet known
 
         # get some threads going -- those will do all the work.
-        try:    import         saga.utils.pty_shell as sups
-        except: import radical.saga.utils.pty_shell as sups
+        try:    import radical.saga.utils.pty_shell as sups
+        except: import         saga.utils.pty_shell as sups
 
         self.launcher_shell = sups.PTYShell("fork://localhost/")
         self.monitor_shell  = sups.PTYShell("fork://localhost/")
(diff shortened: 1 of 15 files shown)


src/radical/pilot/context.py
```diff
@@ -11,18 +11,20 @@
...
-class Context (saga.Context):
+class Context (rs.Context):
 
     # --------------------------------------------------------------------------
     #
     def __init__ (self, ctype, thedict=None) :
 
         # init the saga.Context
-        self._apitype  = 'saga.Context'
+        self._apitype  = 'radical,saga.Context'
```
iparask: There is a typo here: 
```suggestion
        self._apitype  = 'radical.saga.Context'
```

src/radical/pilot/agent/executing/shell.py
```diff
@@ -124,7 +124,9 @@ def initialize_child(self):
                                      # are not yet known
 
         # get some threads going -- those will do all the work.
-        import saga.utils.pty_shell as sups
+        try:    import         saga.utils.pty_shell as sups
+        except: import radical.saga.utils.pty_shell as sups
```
iparask: Since we are preparing for sage renaming shouldn't it be:
```
        try: import radical.saga.utils.pty_shell as sups
        except: import saga.utils.pty_shell as sups
```
This way if radical.saga is installed everything is great otherwise grab saga?

src/radical/pilot/context.py
andre-merzky: Thanks!

src/radical/pilot/agent/executing/shell.py
andre-merzky: Ack, will switch.

andre-merzky:
both done...

PR #1803, opened 2019-02-04
ensure safe termination on faulty installs

andre-merzky:
see https://github.com/radical-collaboration/extasy-bpti/issues/17


mturilli:
Is this ready to be merged?

andre-merzky:
I think its ok to go, but its waiting for review, and needs testing before merge, as it touches the BS.

PR #1804, opened 2019-02-04
expose application comm channels to CUs

andre-merzky:
This introduces an application level communication channel within the scope of a pilot agent.  An example on how to use a pubsub channel is added.  

This adds little overhead to the agent (two more bridges being started).


andre-merzky:
Please give this  try with V2 of RU and RS.  If that works, we can start thinking about a backport.

andre-merzky:
closing in favor of #1870

Issue #1800, opened 2019-02-01
slurm barfs on unset LFS


andre-merzky:
(repotrted by Ray:

the agent_0.err shows:
```
2019-02-01 10:01:46,829: radical.saga        : MainProcess                     : MainThread     : WARNING : cannot initialize context - no session: {'LifeTime' : '-1', 'Type' : 'ssh', 'UserCert' : '/home1/06078/tg853774/.ssh/id_rsa.pub', 'UserKey' : '/home1/06078/tg853774/.ssh/id_rsa'}
Traceback (most recent call last):
  File "/work/06078/tg853774/stampede2/radical.pilot.sandbox/rp.session.js-169-25.jetstream-cloud.org.yren.017928.0000/pilot.0000/rp_install/bin/radical-pilot-agent", line 71, in <module>
    bootstrap_3(sys.argv[1])
  File "/work/06078/tg853774/stampede2/radical.pilot.sandbox/rp.session.js-169-25.jetstream-cloud.org.yren.017928.0000/pilot.0000/rp_install/bin/radical-pilot-agent", line 42, in bootstrap_3
    if agent_name == 'agent_0': agent = rpa.Agent_0(agent_name)
  File "/work/06078/tg853774/stampede2/radical.pilot.sandbox/rp.session.js-169-25.jetstream-cloud.org.yren.017928.0000/pilot.0000/rp_install/lib/python2.7/site-packages/radical/pilot/agent/agent_0.py", line 118, in __init__
    session=self._session)
  File "/work/06078/tg853774/stampede2/radical.pilot.sandbox/rp.session.js-169-25.jetstream-cloud.org.yren.017928.0000/pilot.0000/rp_install/lib/python2.7/site-packages/radical/pilot/agent/rm/base.py", line 229, in create
    return impl(cfg, session)
  File "/work/06078/tg853774/stampede2/radical.pilot.sandbox/rp.session.js-169-25.jetstream-cloud.org.yren.017928.0000/pilot.0000/rp_install/lib/python2.7/site-packages/radical/pilot/agent/rm/slurm.py", line 20, in __init__
    LRMS.__init__(self, cfg, session)
  File "/work/06078/tg853774/stampede2/radical.pilot.sandbox/rp.session.js-169-25.jetstream-cloud.org.yren.017928.0000/pilot.0000/rp_install/lib/python2.7/site-packages/radical/pilot/agent/rm/base.py", line 110, in __init__
    self._configure()
[...]

PR #1809, opened 2019-02-11
Resolving issue #1800.

iparask:
LFS path is checked if it contains environment variables. Otherwise it is left as is.

I does also some PEP8 changes.


itomaldonado:

andre-merzky:
Hmm, I may be missing something - but I think this will not resolve the issue.  The problem is not that there is no `$` in the path string, but that the path is of `NoneType` and thus not iterable.  Your code would have the same exception on the SAGA level I would assume?  The check for `$` is not needed btw:

```
>>> os.path.expandvars('/etc/')
'/etc/'
>>> os.path.expandvars('')
''
>>> os.path.expandvars(None)
  File "/usr/lib/python2.7/posixpath.py", line 287, in expandvars
    if '$' not in path:
TypeError: argument of type 'NoneType' is not iterable
```

`dict.get('entry', '')` can still return `None` if somewhere in the past happened `dict['entry'] = None`.  The `get()` will only shield against non-existing keys, not against unset values.  I assume this should be fixed with something like:

```
val = dict.get(key)
if val: 
    val = os.path.expandvars(val)
```

or with:
```
val = dict.get(key)
if val is None:
    val = ''
val = os.path.expandvars(val)
```
whichever you prefer.  The first one is shorter though :-)

iparask:
> `dict.get('entry', '')` can still return `None` if somewhere in the past happened `dict['entry'] = None`. The `get()` will only shield against non-existing keys, not against unset values. I assume this should be fixed with something like:

You are right! I did not think that it may have been set somewhere else as `None`. Let me see where it is set, because it shouldn't. And commit again.

andre-merzky:
We may not have control over where it's set in all cases.  If `expandvars` cannot handle `None`, we should shield... :-/

andre-merzky:
PS.: another option of course is to implement a more resilient `expandvars` routine in RU.  Like, one which works for generic strings, not only paths.  We do something like this already [here](https://github.com/radical-cybertools/radical.utils/blob/version/2/src/radical/utils/config.py#L13) - with a somewhat different syntax, but that could be factored out and expanded...

iparask:
It's been read from `agent_0.cfg`. I think that `lfs_path_per_node` is set to None in that file. I need to see what is going on exactly. I'll answer to the ticket or with a new commit.

andre-merzky:
Remember though that agent configs can be user supplied...

PR #1815, opened 2019-02-17
Fix/issue 1814

### LINKED ISSUE #1814: Netiface requirement for 0.10.4
When installing RP from pip on MacOS 10.14 I see the following:
```
radical-pilot 0.50.22 has requirement netifaces==0.10.4, but you'll have netifaces 0.10.9 which is incompatible.
```
Local tests with EnTK perform as expected but I am not sure whether this would be a problem with more "realistic" runs. Is requirement for `==0.10.4` a hard requirement?

andre-merzky:
Fixes #1814


mturilli:
Time to merge this?

andre-merzky:
Yes, this will go into the release (this week).

PR #1819, opened 2019-02-19
Feature/lm tests into feature/summit

iparask:
We need to bring the LM testing and everything to Summit branch.

@andre-merzky can you help with the conflicts? I will try them and see what may be wrong.


itomaldonado:

src/radical/pilot/agent/lm/orte.py
```diff
@@ -262,12 +260,14 @@ def construct_command(self, cu, launch_script_hop):
             node_id = node['uid'].rsplit('_', 1)[-1] 
 
             # add all cpu and gpu process slots to the node list.
+            # pylint: disable=unused-variable
             for cpu_slot in node['core_map']: hosts_string += '%s,' % node_id
             for gpu_slot in node['gpu_map' ]: hosts_string += '%s,' % node_id
             for cpu_slot in node['core_map']: depths.add(len(cpu_slot))
+            # pylint: enable=unused-variable
```
andre-merzky: The unused variable should be replaced by `_`.  We don't need a linter if we are not using it ;-)

src/radical/pilot/agent/lm/orte_lib.py
```diff
@@ -265,16 +264,18 @@ def construct_command(self, cu, launch_script_hop):
...
-            
+
             node_id = node[1].rsplit('_', 1)[-1] 
 
             # add all cpu and gpu process slots to the node list.
+            # pylint: disable=unused-variable
             for cpu_slot in node[2]: hosts_string += '%s,' % node_id
             for gpu_slot in node[3]: hosts_string += '%s,' % node_id
             for cpu_slot in node[2]: depths.add(len(cpu_slot))
+            # pylint: enable=unused-variable
```
andre-merzky: same

src/radical/pilot/agent/rm/base.py
```diff
@@ -197,6 +197,7 @@ def __init__(self, cfg, session):
     @classmethod
     def create(cls, name, cfg, session):
 
+        # pylint: disable=protected-access
```
andre-merzky: What warning is addressed with this actually?  Just `session._log`?  Then this should be fixed by adding a log getter to the session I guess.

src/radical/pilot/agent/rm/fork.py
```diff
@@ -49,8 +49,7 @@ def _configure(self):
             self.cores_per_node = 1
 
         self.node_list  = list()
-        requested_nodes = int(math.ceil(float(self.requested_cores) /
-                                        float(self.cores_per_node ) ) )
+        requested_nodes = int(math.ceil(float(self.requested_cores) / float(self.cores_per_node ) ) )
         for i in range(requested_nodes):
```
andre-merzky: why?  This is now longer than 80?

src/radical/pilot/agent/rm/loadleveler.py
```diff
@@ -58,7 +56,7 @@ class LoadLeveler(LRMS):
         29: {'A':  0, 'B': 30, 'C': 28, 'D': 17, 'E': 26},
         30: {'A':  3, 'B': 29, 'C': 31, 'D': 18, 'E': 25},
         31: {'A':  2, 'B': 28, 'C': 30, 'D': 19, 'E': 24},
-        }
+                     }
```
andre-merzky: Uh?

src/radical/pilot/agent/rm/loadleveler.py
```diff
@@ -146,13 +144,13 @@ class LoadLeveler(LRMS):
         13: {'A':  9, 'B':  5, 'C': 12, 'D': 15},
         14: {'A': 10, 'B':  6, 'C': 15, 'D': 12},
         15: {'A': 11, 'B':  7, 'C': 14, 'D': 13},
-        }
+                        }
```
andre-merzky: ?

src/radical/pilot/agent/rm/loadleveler.py
```diff
@@ -625,7 +623,7 @@ def _bgq_create_sub_block_shape_table(self, shape_str):
 
                 # Calculate the number of nodes for the current shape
                 from operator import mul
-                num_nodes = reduce(mul, filter(lambda length: length != 0, sub_block_shape.values()))
+                num_nodes = reduce(mul, [length for length in sub_block_shape.values() if length != 0])
```
andre-merzky: Whats wrong with filter actually?  It should not be slower than list comprehension really?

mturilli: Ugliness and readability?! https://stackoverflow.com/questions/3013449/list-comprehension-vs-lambda-filter no difference really, just preference. Big fun of list comprehension here :P 💃

andre-merzky: Yeah, both ugliness and readability are very much in the eye of the beholder it seems ;-)

.pylintrc
```diff
@@ -0,0 +1,552 @@
...
+        deprecated-operator-function,
+        deprecated-urllib-function,
+        xreadlines-attribute,
+        deprecated-sys-function,
+        exception-escape,
+        comprehension-escape,
+        C,
+        R,
+        relative-import,
+        bare-except
```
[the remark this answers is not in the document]
iparask: Yes, we are.

.pylintrc
```diff
@@ -0,0 +1,552 @@
...
+# naming-style
+#function-rgx=
+
+# Good variable names which should always be accepted, separated by a comma
+good-names=i,
+           j,
+           k,
+           ex,
+           Run,
+           _
```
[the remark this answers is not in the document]
iparask: I do not know. We can always fix it and remove exceptions

.pylintrc
```diff
@@ -0,0 +1,552 @@
...
+# else.
+single-line-if-stmt=no
+
+
+[LOGGING]
+
+# Logging modules to check that the string format arguments are in logging
+# function parameter format
+logging-modules=logging
+
```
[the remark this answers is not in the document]
iparask: I did not add that there.

.pylintrc
```diff
@@ -0,0 +1,552 @@
...
+redefining-builtins-modules=six.moves,past.builtins,future.builtins,io,builtins
+
+
+[CLASSES]
+
+# List of method names used to declare (i.e. assign) instance attributes.
+defining-attr-methods=__init__,
+                      __new__,
+                      setUp
+
```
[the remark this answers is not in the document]
iparask: We can fix it!

src/radical/pilot/agent/rm/fork.py
iparask: I think we instruct flake8 not to look for 80 characters. I can check

src/radical/pilot/agent/rm/loadleveler.py
iparask: It was the wrong indentation. Based on flake8 and pylint closing should be at the same column as opening.

src/radical/pilot/agent/rm/loadleveler.py
iparask: Same as above

src/radical/pilot/agent/rm/loadleveler.py
iparask: pylint jumped with this

.pylintrc
```diff
@@ -0,0 +1,552 @@
...
+valid-classmethod-first-arg=cls
+
+# List of valid names for the first argument in a metaclass class method.
+valid-metaclass-classmethod-first-arg=mcs
+
+
+[DESIGN]
+
+# Maximum number of arguments for function / method
+max-args=5
```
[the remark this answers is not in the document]
iparask: Then we increase the number. I am not sure though how it is going to help.

src/radical/pilot/agent/lm/__init__.py
```diff
@@ -2,7 +2,7 @@
...
 __license__   = "MIT"
 
-
-from .base import *
-from .base import LaunchMethod
-
+# pylint: disable=wildcard-import
+from .base import *  # noqa: F403, F401
+from .base import LaunchMethod  # noqa: F403, F401
+# pylint: enable=wildcard-import
```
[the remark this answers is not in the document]
iparask: Yes, we can do something like that. I am not sure whether we can say that this is priority or not. If we have a continuously failing Travis, because this is not fixed, then I am not sure how automating testing and everything with code quality helps. Then how are you going to enforce it for new pushes?

Nonetheless, it should be a priority for version 2

src/radical/pilot/agent/lm/archive/mpirun_ccmrun.py
```diff
@@ -82,8 +81,9 @@ def construct_command(self, cu, launch_script_hop):
                 hostlist.append(node['name'])
         hosts_string = ",".join(hostlist)
 
-        command = "%s %s -np %d -host %s %s %s" % \
-                  (self.ccmrun_command, self.launch_command, len(hostlist), 
+        command = "%s%s -np %d -host %s %s %s" % \
+                  (self.ccmrun_command,
+                   self.launch_command, len(hostlist),
                    hosts_string, env_string, task_command)
```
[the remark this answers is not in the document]
iparask: No, it was not :P

src/radical/pilot/agent/lm/archive/mpirun_rsh.py
```diff
@@ -70,19 +70,6 @@ def construct_command(self, cu, launch_script_hop):
...
-
-            # Create a hostfile from the list of hosts
-            hostfile = self._create_hostfile(hosts, impaired=True)
-            hosts_string = "-hostfile %s" % hostfile
-
-        else:
-
-            # Construct the hosts_string ('h1 h2 .. hN')
-            hosts_string = " ".join(hosts)
-
```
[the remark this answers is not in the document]
iparask: Did I do that? I am not sure who changed this line.

src/radical/pilot/agent/lm/__init__.py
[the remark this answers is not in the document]
andre-merzky: So, I may be missing the point, but I don't think the idea is to configure the tests so that they pass.  If they show problems, then that's *good*, and they should continue to do so until they are fixed.  If we only care about green boxes, then `assert(True)` will do the trick ;-)

So I'd rather use a pylintrc which reflects what we want in terms of code checks, instead of one which is green and disabled in some code only so that the boxes turn green.  What's the point?

I am surprised that this is me saying this, who like tests the least amongst all of us ;-)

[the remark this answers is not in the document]
mturilli: I agree. `from ..some.module.name import *` is commonly deprecated and we should avoid it.

src/radical/pilot/agent/rm/base.py
mturilli: The same as above. Let's use linters to learn better coding practices. I think respecting private variables is good practice. Would you all agree?

src/radical/pilot/agent/rm/fork.py
mturilli: I would keep the 80 characters rule. Even if all our monitors can accomodate longer lines these days, the limit helps to notice when the code becomes unreasonable 'deep'. Would you agree?

andre-merzky: :+1:   There are always exception one can consider - this shouldn't be one.

src/radical/pilot/agent/rm/loadleveler.py
andre-merzky: But this is much worse than it was before.  If this needs fixing, then the *data* need to be shifted, too.

src/radical/pilot/agent/rm/loadleveler.py
andre-merzky: It threw an error on `filter`?  Ugh.  While some people might not like this, its a totally valid construct which has advantages in some use cases, compared to list comprehension.  Unless we understand why pylint complains here, this is a warning I would ignore...

src/radical/pilot/agent/rm/loadleveler.py
iparask: Yes, they should!

src/radical/pilot/agent/rm/loadleveler.py
iparask: It is an anti-pattern: https://docs.quantifiedcode.com/python-anti-patterns/readability/using_map_or_filter_where_list_comprehension_is_possible.html

andre-merzky: I would suggest not blindly converting filter and map to LC.  In the cases in this file its fine - but lets not remove valid, legal, well formatted and performant code just because some anti-pattern guide says so, at least not without a better motivation...

I will leave the current changed in place.

PR #1821, opened 2019-02-21
Feature/jsrun erf

iparask:
This is the PR to move from resource files to explicit resource files


itomaldonado:

src/radical/pilot/agent/lm/jsrun.py
```diff
@@ -27,23 +27,28 @@ def _configure(self):
 
     # --------------------------------------------------------------------------
     #
-    def _create_resource_set_file(self, slots, uid, sandbox):
+    def _create_resource_set_file(self, slots, uid, sandbox, mpi=False,
+                                  gpu=False):
```
andre-merzky: I don't understand why we pass `gpu` and `mpi` flags - what difference does it make?  If GPUs are defined in the slots, they need to go into the resource sets - the flag won't change that.  It seems like `mpi` triggers the specification of `rank` - but as discussed, this has not much to do with the MPI ranks, and I'd argue we should use the same syntax for MPI and non-MPI variants (specifying `rank` for non-MPI is fine).

iparask: I am perfectly fine with that, but:
```bash
bash-4.2$ cat unit.000000.rs 
cpu_index_using: physical
rank: 0: { host: 1; cpu: {0-13}}
rank: 0: { host: 1; cpu: {22-36}}
bash-4.2$ jsrun --erf_input=$pwd/unit.000000.rs /bin/date
ERF error: 1+ ranks are assigned with multiple resource sets.
```
```bash
bash-4.2$ cat unit.000000.rs 
cpu_index_using: physical
1: { host: 1; cpu: {0-13}}
1: { host: 1; cpu: {22-36}}
bash-4.2$ jsrun --erf_input=$pwd/unit.000000.rs /bin/date
Thu Feb 28 10:56:27 EST 2019
Thu Feb 28 10:56:27 EST 2019
```
```bash
bash-4.2$ cat unit.000000.rs 
cpu_index_using: physical
rank: 0: { host: 1; cpu: {0-13}}
bash-4.2$ jsrun --erf_input=$pwd/unit.000000.rs /bin/date
Thu Feb 28 10:57:13 EST 2019
```
```bash
bash-4.2$ cat unit.000000.rs 
cpu_index_using: physical
1: { host: 1; cpu: {0-13}}
bash-4.2$ jsrun --erf_input=$pwd/unit.000000.rs /bin/date
Thu Feb 28 10:57:56 EST 2019
```

In my opinion, for non-MPI unit the ERF file should look like examples 2 and 4, despite the fact that  3 works. I will dot the rank part, and I see how to remove the flags.

vivek-bala: Isn't 2 an example for an MPI task? Do you mean you require 'rank' only if it an MPI task? The two seems to be conflicting, no?

iparask: Each line is an application execution. It executes the application defined in the `jsrun` multiple times with as many processes as the number defined at the beginning. I am not sure it will work correctly with MPI

src/radical/pilot/agent/lm/jsrun.py
```diff
@@ -66,25 +71,55 @@ def _create_resource_set_file(self, slots, uid, sandbox):
...
+                for gpu_set in node['gpu_map']:
+                    gpus_str += ['{' + ','.join(str(gpu) for gpu
+                                      in gpu_set) + '}']
+                cores = ','.join(str(core) for core in cores_str)
+                gpus = ','.join(str(gpu) for gpu in gpus_str)
+                rs_str           += '1: {'
+                rs_str           += ' host: %d;' % node['uid']
+                if cores: rs_str += ' cpu: %s' % cores
+                if gpus : rs_str += '; gpu: %s' % gpus
+                rs_str           += '}\n'
```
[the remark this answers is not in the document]
iparask: I do not understand what you mean. Can you give an example or propose a change?

src/radical/pilot/agent/lm/jsrun.py
vivek-bala: Yes, that's precisely how it should work though. I will launch two processes that both execute /bin/date. 
I think my point is connected with my comment below. I'll add an example there. My suggestion is to use
either rank for both MPI and non-MPI applications or not use at rank at all.

src/radical/pilot/agent/lm/jsrun.py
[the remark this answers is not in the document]
vivek-bala: Sure. See example resource layouts. 'rank' is only useful to specify rank-core binding when they are out of order. See example 9 of https://www.ibm.com/support/knowledgecenter/en/SSWRJV_10.1.0/jsm/10.2/base/erf_format.html. Otherwise with and without 'rank' are equivalent.

With 'rank':
```bash
# MPI task
bash-4.2$ cat erf_1.rs 
cpu_index_using: physical
rank: 0: {host: 1; cpu: {0,1,2,3}}
rank: 1: {host: 1; cpu: {6,7,8,9}}

bash-4.2$ jsrun --erf_input erf_1.rs /ccs/proj/csc190/Hello_jsrun/jsrun_layout
########################################################################

*** MPI Ranks: 2, OpenMP Threads: 4, GPUs per Resource Set: 0 ***
========================================================================
MPI Rank 000 of 002 on HWThread 000 of Node a30n01, OMP_threadID 0 of 4
MPI Rank 000 of 002 on HWThread 001 of Node a30n01, OMP_threadID 1 of 4
MPI Rank 000 of 002 on HWThread 003 of Node a30n01, OMP_threadID 2 of 4
MPI Rank 000 of 002 on HWThread 002 of Node a30n01, OMP_threadID 3 of 4
MPI Rank 001 of 002 on HWThread 007 of Node a30n01, OMP_threadID 1 of 4
MPI Rank 001 of 002 on HWThread 006 of Node a30n01, OMP_threadID 0 of 4
MPI Rank 001 of 002 on HWThread 008 of Node a30n01, OMP_threadID 3 of 4
MPI Rank 001 of 002 on HWThread 009 of Node a30n01, OMP_threadID 2 of 4

# Non-MPI task (i.e. task with 1 process)
bash-4.2$ cat erf_2.rs 
rank: 0: {host: 1; cpu: {0,1,2,3}}

bash-4.2$ jsrun --erf_input erf_2.rs /ccs/proj/csc190/Hello_jsrun/jsrun_layout
########################################################################

*** MPI Ranks: 1, OpenMP Threads: 4, GPUs per Resource Set: 0 ***
========================================================================
MPI Rank 000 of 001 on HWThread 000 of Node a30n01, OMP_threadID 1 of 4
MPI Rank 000 of 001 on HWThread 003 of Node a30n01, OMP_threadID 2 of 4
MPI Rank 000 of 001 on HWThread 001 of Node a30n01, OMP_threadID 0 of 4
MPI Rank 000 of 001 on HWThread 002 of Node a30n01, OMP_threadID 3 of 4
```

Without rank:
```bash
# MPI task
bash-4.2$ cat erf_3.rs 
1: {host: 1; cpu: {0,1,2,3}}
1: {host: 1; cpu: {6,7,8,9}}

bash-4.2$ jsrun --erf_input erf_3.rs /ccs/proj/csc190/Hello_jsrun/jsrun_layout
MPI Rank 001 of 002 on HWThread 007 of Node a30n01, OMP_threadID 1 of 4
MPI Rank 001 of 002 on HWThread 009 of Node a30n01, OMP_threadID 2 of 4
MPI Rank 001 of 002 on HWThread 008 of Node a30n01, OMP_threadID 3 of 4
MPI Rank 001 of 002 on HWThread 006 of Node a30n01, OMP_threadID 0 of 4
########################################################################

*** MPI Ranks: 2, OpenMP Threads: 4, GPUs per Resource Set: 0 ***
========================================================================
MPI Rank 000 of 002 on HWThread 001 of Node a30n01, OMP_threadID 3 of 4
MPI Rank 000 of 002 on HWThread 000 of Node a30n01, OMP_threadID 0 of 4
MPI Rank 000 of 002 on HWThread 002 of Node a30n01, OMP_threadID 1 of 4
MPI Rank 000 of 002 on HWThread 003 of Node a30n01, OMP_threadID 2 of 4

# Non-MPI task (i.e. task with 1 process)
bash-4.2$ cat erf_4.rs 
cpu_index_using: physical
1: {host: 1; cpu: {0,1,2,3}}

bash-4.2$ jsrun --erf_input erf_4.rs /ccs/proj/csc190/Hello_jsrun/jsrun_layout
########################################################################

*** MPI Ranks: 1, OpenMP Threads: 4, GPUs per Resource Set: 0 ***
========================================================================
MPI Rank 000 of 001 on HWThread 000 of Node a30n01, OMP_threadID 0 of 4
MPI Rank 000 of 001 on HWThread 003 of Node a30n01, OMP_threadID 1 of 4
MPI Rank 000 of 001 on HWThread 002 of Node a30n01, OMP_threadID 2 of 4
MPI Rank 000 of 001 on HWThread 001 of Node a30n01, OMP_threadID 3 of 4
```

IMO we don't really require both 'rank' and without 'rank' modes. Its enough to keep one.
That said, I don't think we need 'rank' since we won't be doing out of order rank-core binding since we need application knowledge for that. My suggestion would to be go with no specification about the rank (i.e. 'without rank' in my above example which is equivalent to the 'else' case you have in the implementation).

Just for completion what I mean by out of order rank-core binding:
We will always assign resources to rank 0, then to rank 1 and so on. 'rank' is only useful
when we want to assign resources to rank j before rank i where j>i. (example 9 of https://www.ibm.com/support/knowledgecenter/en/SSWRJV_10.1.0/jsm/10.2/base/erf_format.html)

Let me know if you see something I am missing.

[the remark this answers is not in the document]
iparask: I do not think that `Specific ordering` requires out of order necessarily. I prefer the specificity of the ranks because it leaves no room for errors in interpretation. I can remove them if it causes so much trouble.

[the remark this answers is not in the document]
vivek-bala: No it doesn't. Purpose of rank is to specify an order. We are choosing it to be a particular order (ascending order), that's all. I do think we are over-specifying it. My suggestion would be to not use 'rank'. 

Either way I would recommend to use one approach for mpi and non-mpi tasks.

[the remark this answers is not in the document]
iparask: So @andre-merzky any opinions on this? It will affect also the way test for #1822 change with the ERF format!

mturilli:
@iparask Are we ready to merge?

src/radical/pilot/agent/lm/jsrun.py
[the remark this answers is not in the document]
mturilli: @andre-merzky Any opinion? I am happy with keeping as general as possible

src/radical/pilot/agent/lm/jsrun.py
mturilli: @iparask Following yesterday discussion on Slack, did we all converge on this?

src/radical/pilot/agent/lm/jsrun.py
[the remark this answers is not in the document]
iparask: Based on last evenings Slack heated discussion, we are staying with the explicit rank style files.

iparask:
Not yet, I still have to work on one of the comments @andre-merzky and @vivek-bala left.

src/radical/pilot/agent/lm/jsrun.py
```diff
@@ -66,25 +70,34 @@ def _create_resource_set_file(self, slots, uid, sandbox):
...
-            rs_str           += 'RS %d: {'  % rs_id
-            rs_str           += ' host: %d' % node['uid']
-            if cores: rs_str += ' cpu: %s'  % cores
-            if gpus : rs_str += ' gpu: %s'  % gpus
-            rs_str           += ' }\n'
-            rs_id            += 1
+            if node['gpu_map']: 
+                node_map = zip(node['core_map'], node['gpu_map'])
+            else:
+                node_map = zip(node['core_map'], *node['gpu_map'])
```
andre-merzky: What does this do?

src/radical/pilot/agent/lm/jsrun.py
```diff
@@ -66,25 +70,34 @@ def _create_resource_set_file(self, slots, uid, sandbox):
...
-                                               in  node['gpu_map']])
-
-            rs_str           += 'RS %d: {'  % rs_id
-            rs_str           += ' host: %d' % node['uid']
-            if cores: rs_str += ' cpu: %s'  % cores
-            if gpus : rs_str += ' gpu: %s'  % gpus
-            rs_str           += ' }\n'
-            rs_id            += 1
+            if node['gpu_map']: 
+                node_map = zip(node['core_map'], node['gpu_map'])
```
andre-merzky: This assumes that both maps have the same length?  If so, is that guaranteed?  If not,  what happens to the superfulous nodes which get dropped in the zipped result?

src/radical/pilot/agent/lm/jsrun.py
```diff
@@ -66,25 +70,34 @@ def _create_resource_set_file(self, slots, uid, sandbox):
...
+
+                cores = ','.join(str(core * 4) for core in map_set[0])
+
+                rs_str += 'rank: %d: {'  % rank
+                rs_str += ' host: %s;'  % str(node['uid'])
+                rs_str += ' cpu: {%s}'  % cores
+                try:
+                    if map_set[1]:
+                        gpus  = ','.join(str(gpu) for gpu in map_set[1])
+                        print gpus
```
andre-merzky: This looks like a debug print.

src/radical/pilot/agent/lm/jsrun.py
```diff
@@ -99,14 +112,14 @@ def construct_command(self, cu, launch_script_hop):
...
-        task_procs     = cud.get('cpu_processes', 0)
-        task_env       = cud.get('environment') or dict()
-        task_args      = cud.get('arguments')   or list()
-        task_argstr    = self._create_arg_string(task_args)
-        task_sandbox   = ru.Url(cu['unit_sandbox']).path
+        slots        = cu['slots']
+        cud          = cu['description']
+        task_exec    = cud['executable']
+        task_env     = cud.get('environment') or dict()
+        task_mpi     = bool('mpi' in cud.get('cpu_process_type', '').lower())
```
andre-merzky: task_mpi should not be needed anymore?

iparask: Yeah...

src/radical/pilot/agent/lm/jsrun.py
iparask: Hmmm you are right. I went with the assumption that we have GPUs per process. I'll fix it. Sorry

andre-merzky: We might have, with our recent discussion on only allowing uniform processes and all - but in the current state I am not sure the scheduler honors that, so lets stay on the save side for the time being...

Thanks!

src/radical/pilot/agent/lm/jsrun.py
iparask: It might not be needed. It creates a zip even when the second iteratable is emtpy.

iparask:
I did all the changes as requested. The unit tests still pass. Can you take a look?

andre-merzky:
@vivek-bala : are your comments addressed?

andre-merzky:
Thank you! :-)

PR #1822, opened 2019-02-21
Feature/summit jsrun tests

vivek-bala:
Opening a PR so everyone can take a look at the test suite. Let me know what you think.

There is one aspect of the test that I am still determining how to do: tests where we run MPI application where each process needs to use a distinct node, we need to assert that in fact two different nodes were used (tests 5 and 6). If you have any ideas, feel free to share.


itomaldonado:

tests/test_launch_methods/test_jsrun/jobs.sh
```diff
@@ -0,0 +1,83 @@
...
+}
+
+assert() {
+    actual=$1
+    expected=$2
+    tid=$3
+
+    cp $actual results_$d/$tid/jsrun_output_dup.log
+    # Remove Node ID from actual output
+    sed -i 's/Node ....../Node /g' results_$d/$tid/jsrun_output_dup.log
```
mturilli: Why do you remove the IDs?

tests/test_launch_methods/test_jsrun/jobs.sh
```diff
@@ -0,0 +1,83 @@
...
+    omp_env=$2
+    cmd=$3
+    exp_out=$4
+    tid=$5
+
+    # Create required folders
+    rm results_$d/$tid -rf
+    mkdir results_$d/$tid
+
+    # Record resource file, omp env, and cmd in log
```
mturilli: What omp stands for?

tests/test_launch_methods/test_jsrun/test_1/cmds
```diff
@@ -0,0 +1,3 @@
+# Task needs to be assigned node 1, physical core 1, HW thread 1
+export OMP_NUM_THREADS=1
+jsrun -a 1 -U resource-file /ccs/proj/csc190/Hello_jsrun/jsrun_layout
```
mturilli: Add newline at te end of file?

tests/test_launch_methods/test_jsrun/jobs.sh
```diff
@@ -0,0 +1,83 @@
...
+#BSUB -o jsrun_test.%J
+#BSUB -e jsrun_test.%J
+#BSUB -N rct-tests@googlegroups.com
+
+# CONFIG
+path=`pwd`
+d=`date +%D | sed 's/\//_/g'`
+host=`/bin/hostname`
+#dest="rct-tests@googlegroups.com"
+dest="vivek.bala@rutgers.edu"
```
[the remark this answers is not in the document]
iparask: I agree with Matteo

tests/test_launch_methods/test_jsrun/jobs.sh
iparask: It is for OPENMP

tests/test_launch_methods/test_jsrun/jobs.sh
```diff
@@ -0,0 +1,83 @@
...
+
+    # Record resource file, omp env, and cmd in log
+    echo $res_set >> results_$d/$tid/full_output.log
+    echo $omp_env >> results_$d/$tid/full_output.log
+    echo $cmd >> results_$d/$tid/full_output.log
+
+    # Set omp env, create resource file and run cmd
+    $omp_env
+    echo $res_set >> $path/resource-file
+    cmd=`echo $cmd | sed "s#resource-file#$path/resource-file#g"`
```
iparask: Since there is a resource file in each folder why you write it to a different file and then remove it?

tests/test_launch_methods/test_jsrun/test_5/cmds
```diff
@@ -0,0 +1,3 @@
+# Task needs to be assigned nodes 1 and 2, physical core 1 of each node, HW thread 1 of each physical core
+export OMP_NUM_THREADS=1
+jsrun -a 1 -U resource-file /ccs/proj/csc190/Hello_jsrun/jsrun_layout
```
iparask: You are missing `verbose`, otherwise the node name will never appear in the output

iparask:
Why don't you take the node list (`$LSB_HOST`) and keep the unique values apart from the first?

tests/test_launch_methods/test_jsrun/jobs.sh
andre-merzky: Presumably because the differ from run to run, and are thus difficult to verify against a static output?

andre-merzky:
@vivek-bala : action needed wrt. comments.

tests/test_launch_methods/test_jsrun/jobs.sh
vivek-bala: Yes it stands for OpenMP. The value is being extracted from the scripts and executed in the current session.

tests/test_launch_methods/test_jsrun/jobs.sh
vivek-bala: The content of the resource file is appended to the output log inside each folder. The resource file itself
is created in ``path`` and then deleted at the end of the individual test.

tests/test_launch_methods/test_jsrun/jobs.sh
vivek-bala: Yes, that's correct. The node IDs differ between runs. The same can hold for HW threads in some tests.

tests/test_launch_methods/test_jsrun/test_5/cmds
vivek-bala: I am not sure I follow. These were the exact commands that were run and I do get the node name/ID. Let me know if you think I missed something.

tests/test_launch_methods/test_jsrun/test_1/cmds
vivek-bala: Done. Thanks.

tests/test_launch_methods/test_jsrun/jobs.sh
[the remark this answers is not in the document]
vivek-bala: Done. Makes sense. Thanks.

tests/test_launch_methods/test_jsrun/jobs.sh
iparask: Yes, but why are you doing all that, when you can just  use the `res_set` file as input?

tests/test_launch_methods/test_jsrun/test_1/res_set
```diff
@@ -0,0 +1 @@
+RS 0: { host: 1, cpu: 0}
```
mturilli: Should we use ERF?

tests/test_launch_methods/test_jsrun/jobs.sh
vivek-bala: Fair point. Changed and tested.

tests/test_launch_methods/test_jsrun/test_1/res_set
vivek-bala: Yes I agree. @iparask let me know how we can proceed on that.

iparask: @vivek-bala I'll make a commit now that you are done

tests/test_launch_methods/test_jsrun/jobs.sh
```diff
@@ -0,0 +1,79 @@
...
+    expected=$2
+    tid=$3
+
+    cp $actual results_$d/$tid/jsrun_output_dup.log
+    # Remove Node ID from actual output
+    sed -i 's/Node ....../Node /g' results_$d/$tid/jsrun_output_dup.log
+    if ! diff $expected results_$d/$tid/jsrun_output_dup.log; then
+        op="$tid failed: Output location = results_$d/$tid/jsrun_output.log, Diff location = results_$d/$tid/full_output.log"
+        echo "Diff between expected and actual (bar node ID):" >> results_$d/$tid/full_output.log
+        diff $expected results_$d/$tid/jsrun_output_dup.log >> results_$d/$tid/full_output.log
```
iparask: Can you put the absolute path here? I am not sure from the email where the log file is. As a result, I cannot check it

tests/test_launch_methods/test_jsrun/test_1/cmds
iparask: I suggest the following:
```suggestion
jsrun -a 1 -U resource-file /ccs/proj/csc190/Hello_jsrun/jsrun_layout | sort
```

This way all outputs will be sorted and easier to compare with the file you produce. Especially with MPI execution, where the print order is not guaranteed.

iparask:
May also I suggest that it moves under the `test/lm/` folder as integration tests?

vivek-bala:
Moved the tests and made some changes to the test assertions that are more extensible

iparask:
This is an integration test, right? May I suggest then that there is a folder named `integration` under `lm` and `test_jsurn` to be named just `jsrun`?
It will fall very nicely with the `unit` folder that denotes the unit tests!

vivek-bala:
Thanks Giannis. Good point. Moved it into another folder integration tests. I renamed some folders with 'test_' so that they can be auto-discovered by pytest.

tests/test_launch_methods/test_jsrun/jobs.sh
mturilli: @vivek-bala any follow up on this?

andre-merzky:
Merging now.  I think that Ioannis' request for absolute paths is still open, but I don't want to delay further.

Thanks all!

PR #1828, opened 2019-03-08
Feature/travis

andre-merzky:
This should have been merged a while ago... :/


tests/api/test_pilot_manager.py
```diff
@@ -19,14 +18,14 @@
 if DBURL is None:
     print "ERROR: RADICAL_PILOT_DBURL (MongoDB server URL) is not defined."
     sys.exit(1)
-    
-DBNAME = os.getenv("RADICAL_PILOT_TEST_DBNAME")
+
+DBNAME = os.getenv("RADICAL_PILOT_TEST_DBNAME", 'test')
```
iparask: Apparrently, MLab does not work the way it used anymore. Also, I do not see where the DBURL is set. Is it fair to assume that this is taken care of?

tests/go_test.py
```diff
@@ -27,106 +28,112 @@ def unit_state_change_cb (unit, state) :
...
-except Exception as e:
-    print "TEST FAILED"
-    raise
-
-finally:
-    # Remove session from database
-    session.close()
+
+def test_go():
+
```
iparask: If this test is disabled, shouldn't we remove it? Also, is there a use case that requires `go`? I'm asking out of curiosity

iparask:
Also Travis fails because: `E   RuntimeError: Couldn't create new session (database URL 'mongodb://rp:rp@ds015335.mlab.com:15335/rp' incorrect?): too many namespaces/collections`.

I think the DB should be changed to something we have access and own.

andre-merzky:
check with rm_test branch

tests/go_test.py
andre-merzky: `go` as in `GlobusOnline` :-)  Its disabled because we don't have an installtion available, but I would not remove it as long as we have the code.

tests/api/test_pilot_manager.py
andre-merzky: Out of scope for this PR.  Yes, we need to find a MongoDB servcice replacement... :/

andre-merzky: Yes, URL setting should be taken case of.

tests/api/test_pilot.py
```diff
@@ -69,68 +68,68 @@ def test__pilot_wait(self):
...
+        pilot.wait(timeout=300)
+        assert pilot.state == rp.FAILED, "State is '%s' instead of 'Failed'." % pilot.state
 
-        cpd = radical.pilot.ComputePilotDescription()
+        cpd = rp.ComputePilotDescription()
         cpd.resource = "local.localhost"
         cpd.cores = 100000000000  # This should fail - at least in 2014 ;-)
         cpd.runtime = 1
-        cpd.sandbox = "/tmp/radical.pilot.sandbox.unittests"
+        cpd.sandbox = "/tmp/rp.sandbox.unittests"
```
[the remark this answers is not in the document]
andre-merzky: The changes really only try to get the test code closer to pep-8, as we agreed to do with all code we files in a PR for some reason.  I feel like it moves the PR out of scope if the result it to attempt to fix additional semantic or contextual issues in that code, too.

tests/go_test.py
mturilli: I would probably rename it `globus_online_test.py`. Go is slightly overloaded these days ;)

andre-merzky:
`travis.yaml` is indeed conflicting with the one from `feature/rm_tests`,  but we'll merge this one for now anyway, and will replace it with the new one once that is merged.  The latter will take a while, since that branch comes in via the Summit changes, and those are not yet up for release.

andre-merzky:
Travis integration works.  Tests fail, but that is addressed in other branches.

PR #1829, opened 2019-03-12
Feature/expanded tests

### LINKED ISSUE #1827: Expand CI tests
Based on:
- https://docs.google.com/spreadsheets/d/1Mmw3Fw9Yt5hUo5FLenqeJRM5F3OBfyM4-hRH-aPxJ4g/edit#gid=0
- https://docs.google.com/document/d/1N2x08FkDsui6jnHazIBrfYdQ8GGyDTBGqA887F_wpTo/edit

The two matrices should be reduced to a comprehensive one to be run as:
- Jsrun standalone test (as actually done via at)
- As integration test with RP via at too.

iparask:
The purpose of this PR is to check whether it is in accordance with issue #1827 and the integration tests of @vivek-bala 

It fixes #1827


itomaldonado:

tests/test_lm/test_integration/test_jsrun/test_3/res_set
```diff
@@ -0,0 +1,2 @@
+cpu_index_using: physical
+rank: 0: { host: 1; cpu: {0,4,8,12,16,20,24,28}}
```
mturilli: Could this also be:
`rank: 0: { host: 1; cpu: {1,2,3,4,5,6,7,8}}`

tests/test_lm/test_integration/test_jsrun/test_3/cmds
```diff
@@ -0,0 +1,3 @@
+# Task needs to be assigned node 1, physical core 1, HW thread 8
```
mturilli: I am not sure this is possible: 8 HW threads = 2 cores as each core has up to 4 HWT, right?

tests/test_lm/test_integration/test_jsrun/test_4/cmds
```diff
@@ -0,0 +1,3 @@
+# Task needs to be assigned node 1, physical core 1, HW thread 8, 1 GPU
```
mturilli: See my previous comments about cores/HWT

tests/test_lm/test_integration/test_jsrun/test_2/cmds
```diff
@@ -1,3 +1,3 @@
-# Task needs to be assigned node 1, physical cores 1 and 2, HW thread 1 of each physical core
+# Task needs to be assigned node 1, physical core 1, HW thread 1, 1 GPU
```
vivek-bala: We should stick to one convention:

"Physical core 1" means "physical core with (or at) index 1"
"1 GPU" refers to quantity. I suggest changing this to "GPU 1" (or GPU 0 - whichever is appropriate)

tests/test_lm/test_integration/test_jsrun/test_3/cmds
vivek-bala: "physical core 1, HW thread 8" -- I don't think this is too useful.
"physical cores 1-8, HW thread 1 of each physical core" maybe? * (marked for reference)

tests/test_lm/test_integration/test_jsrun/test_4/cmds
vivek-bala: Same as above (marked *)

tests/test_lm/test_integration/test_jsrun/test_5/cmds
```diff
@@ -1,3 +1,3 @@
-# Task needs to be assigned nodes 1 and 2, physical core 1 of each node, HW thread 1 of each physical core
+# Task needs to be assigned node 1, physical core 8, HW thread 1, 0 GPUs
```
vivek-bala: Same as above (marked *)
Indices start from 0 or 1? Just to make sure we are consistent.

tests/test_lm/test_integration/test_jsrun/test_6/cmds
```diff
@@ -1,3 +1,3 @@
-# Task needs to be assigned nodes 1 and 2, physical cores 1 and 2 of each node, HW threads 1 and 2 of each physical core
-export OMP_NUM_THREADS=2
+# Task needs to be assigned node 1, physical core 8, HW thread 1, 1 GPU
```
vivek-bala: Same as above (marked *)

tests/test_lm/test_integration/test_jsrun/test_7/cmds
```diff
@@ -0,0 +1,3 @@
+# Task needs to be assigned node 1, physical core 8, HW thread 8
```
vivek-bala: Same as above (marked *)

tests/test_lm/test_integration/test_jsrun/test_8/cmds
```diff
@@ -0,0 +1,3 @@
+# Task needs to be assigned node 1, physical core 1, HW thread 1
```
vivek-bala: Same as above (marked *)

tests/test_lm/test_integration/test_jsrun/test_2/cmds
iparask: okay

tests/test_lm/test_integration/test_jsrun/test_3/res_set
iparask: No, unfortunately it is not. Regardless the index we use, we have to refer to cores based on their logical ID. This was communicated to me by ORNL, when we were discussing the SMT levels. 
Quote from a ticket with ORNL:
>  This seems a little misleading, as I'd expect "physical" to accept numbers 0-41 and virtual to take 0-167 (when in SMT4 mode). Instead, in SMT1 it seems like the 0-167 designation remains but only the "first" slice on each core is honored. I'll have to check into that.

As a result we need to keep the logical numbering until ORNL gets back at me.

iparask:
@vivek-bala can you also check if the outputs are correct? I would like to avoid running them before you say yes. Fixing a test to run correctly, does not mean that the test is correct. I want the test to be correct based on the convention you've created.

tests/test_lm/test_integration/test_jsrun/test_3/cmds
iparask: I wanted to mention 4 threads per process. If the comment is misleading, I'll change it.

tests/test_lm/test_integration/test_jsrun/test_4/cmds
iparask: Will fix

iparask:
Also, while reviewing keep in mind the [test configuration spreadsheet](https://docs.google.com/spreadsheets/d/1Mmw3Fw9Yt5hUo5FLenqeJRM5F3OBfyM4-hRH-aPxJ4g/edit#gid=0)

tests/test_lm/test_integration/test_jsrun/test_3/cmds
iparask: It means 1 process 8 threads. Neither of your suggestions are correct. I will change all the comments and push

iparask:
On a more general note, I will put the comments based on the test configuration and how that is translated. Is that okay with everyone?

tests/test_lm/test_integration/test_jsrun/test_5/cmds
iparask: The node list starts from 1, CPUs and GPUs from 0

tests/test_lm/test_integration/test_jsrun/test_2/cmds
iparask: @vivek-bala, does the check look for a specific GPU index or the number of GPUs?

tests/test_lm/test_integration/test_jsrun/test_2/res_set
```diff
@@ -0,0 +1,2 @@
+cpu_index_using: physical
+rank: 0: { host: 1; cpu: {0}; gpu: {1}}
```
vivek-bala: Do gpu indices start with 1? Any specific reason it is 'gpu: {1}' and not 'gpu: {0}'?

vivek-bala:
Yes, I'm checking that as well. The comments in the cmds will be helpful, I will be able to confirm once you have iterated on them.

tests/test_lm/test_integration/test_jsrun/test_2/res_set
iparask: They start by `0`, but what does it matter when verification is based on the number of resources used and not which resources?

vivek-bala: Correct, we only test for the number of cpus/gpus assigned and not their index. I will start make my initial suggestion.

Two reasons:
* We simply remain consistent between cpu and gpu. Going with the first gpu (as we do for cpu) makes sense than picking a random number. There should be a reason on why we chose a particular configuration to test.
* When/if we add tests to check the index of the cpu/gpu assigned, you will have to go in and change the comments as well as the tests (or you write your test to check if gpu 1 was acquired -- again why gpu 1?).

To me, this is trivial and starting from the first index makes sense. Testing if we got gpu 1 ( or any other cpu/gpu) can be an additional test, of course.

vivek-bala: FWIW only test 2 tries to assign gpu 1. Rest all assign gpu 0.

vivek-bala:
All tests look correct. I see that the multi-node tests have been removed (originally test 5 and 6). Should we add that to the list of tests (with the new test ids)? or was there a particular reason to remove the multi-node tests?

I have a suggestion regarding test 6 and 8. From the spreadsheet, we have
```
test id - 6
processes: 8 
threads: 1
GPUs: 1

test id - 8
processes: 8 
threads: 4
GPUs: 1
```

This gives the impression of "threads per process" and "GPUs per process". In the test though, we are actually using it as "threads per process" and "GPUs per jsrun". See [here](https://github.com/radical-cybertools/radical.pilot/blob/feature/expanded_tests/tests/test_lm/test_integration/test_jsrun/test_6/res_set#L2). 

Can I suggest that we test for "GPUs per process"? (of course we can't more than 6 processes per node). The resource set for test 6 will look like:
```
cpu_index_using: physical
rank: 0: { host: 1; cpu: {0}; gpu: {0}}
rank: 1: { host: 1; cpu: {4}; gpu: {1}}
rank: 2: { host: 1; cpu: {8}; gpu: {2}}
rank: 3: { host: 1; cpu: {12}; gpu: {3}}
rank: 4: { host: 1; cpu: {16}; gpu: {4}}
rank: 5: { host: 1; cpu: {20}; gpu: {5}}
```
I'm happy to make the change if so. Or maybe we add this as a new test?

iparask:
I'll add them! The spreadsheet says 1 GPU, so I assumed one GPU in total. I'll a test with 1 GPU per process, with 6 processes.

tests/test_lm/test_integration/test_jsrun/test_3/cmds
mturilli: Do we have two processes? Is each process placed onto a separate physical core?

tests/test_lm/test_integration/test_jsrun/test_2/res_set
mturilli: For consistency, I would also start both CPU and GPU from 0 (or both from 1 that would also be consistent but potentially confusing as it might indicate that the number starts from 1)

mturilli:
@iparask to remove non-uniform tests and merge the PR

PR #1840, opened 2019-03-27
FIx issue 1072

mturilli:
Updated documentation for virtualenv but the section about troubleshooting seems badly outdated:
- Pip is now v 19.0.3 so I am guessing everything about pip having to be 1.4.1 is to be deleted.
- We write that we do not support conda while I think we do.


itomaldonado:

andre-merzky:
Thanks!

PR #1846, opened 2019-04-03
fix var expansion

andre-merzky:
This uses the RU backport of `expand_env` to get RP to work on Tiger (don't want to wait for the release) - see radical-cybertools/radical.utils/pull/161


andre-merzky:
Thanks @iparask !

PR #1848, opened 2019-04-09
Fix/issue 1834

vivek-bala:
This is PR compares expected string with actual string (output) obtained when running the test cmds.
This PR is not completed as the current string parsing does not succeed for all tests. Just opening for discussion.


itomaldonado:

PR #1862, opened 2019-05-23
Fix/titan ve

andre-merzky:
update ve creation script, update titan config to new ve location


bin/radical-pilot-create-static-ve
```diff
@@ -112,11 +112,16 @@ mkdir -p "$prefix"
...
-curl -k -L -O "$VIRTENV_TGZ_URL" 2>&1 | progress
-tar zxmf "$VIRTENV_TGZ"
-VIRTENV_CMD="$PYTHON virtualenv-1.9/virtualenv.py"
+
+VIRTENV_CMD="$(which virtualenv 2>/dev/null)"
+if test -z "$VIRTENV_CMD"
+then
+    VIRTENV_TGZ="virtualenv-1.9.tar.gz"
+    VIRTENV_TGZ_URL="https://pypi.python.org/packages/source/v/virtualenv/$VIRTENV_TGZ"
+    curl -k -L -O "$VIRTENV_TGZ_URL" 2>&1 | progress
```
iparask: Is progress installed everywhere? For example, I do not have it in any of my VMs, so if I tried to create a static VE this would fail.

andre-merzky: See https://github.com/radical-cybertools/radical.pilot/blob/devel/bin/radical-pilot-create-static-ve#L49 ;-)

PR #1867, opened 2019-06-11
Feature/srun

andre-merzky:
This implements the srun launch method, and a noop scheduler.


src/radical/pilot/agent/lm/srun.py
```diff
@@ -0,0 +1,133 @@
...
+                    break
+
+                for _ in node['core_map']:
+                    hostlist.append(node['name'])
+
+                for _ in node['gpu_map']:
+                    hostlist.append(node['name'])
+
+            placement = ""
+            if uniform and len(hostlist) > 1:
```
lee212: In the description, `srun` has the condition: `IF    task <= nodesize    OR    task is uniformel`. `task <= nodesize` is implicitly applied? because I don't find it.

src/radical/pilot/configs/resource_princeton.json
```diff
@@ -51,25 +52,26 @@
         "local"                      :
         {
             "job_manager_endpoint"    : "slurm://localhost/",
+            "job_manager_hop"         : "fork://localhost/",
```
lee212: Out of curiosity, what does `job_manager_hop` do? If it is a placeholder for schema and hostname of `js_url`, empty `job_manager_hop` still has the same behavior? like falling into `elif '+' not in js_url.schema:` condition in `session.py` if it is missing.

andre-merzky: Ah, good question!

This is somewhat of a historic remnant.  In some cases we had to  use resource URLs to access a batch systems of a cluster via another gateway machine.  In those cases it was difficult to derive the endpoint for file transfer and command submission (where to run `qsub`) from the URL available for the actual resource, and we used the `js_hop` to hardcode that.

By now the code to derive the URL  is more clever, and we handle gateway machines differently, so that setting is not used much - but the code is still there and probably should be cleaned out.

The only really useful setting is for localhost and testing (like here), where I can configure RP to use local `slurm` for submission, but to access the login node via ssh.  That then works even when I am on the loginn node, where the heuristics would think ssh is not needed.  That way we can test ssh (and other remote access mechanisms) when running on localhost.

src/radical/pilot/agent/lm/srun.py
andre-merzky: Yes, indeed - good catch.  This code does not attempt to place the task on a specific node in the single node case, but leaves the placement to srun (or rather slurm).  The same as for non-uniform multi-node tasks.  I  agree that this is inconsistent - I'll add the placement for single-node tasks (which are also uniform by definition, so that just means removing the check for hostlist length).  Thanks!

PR #1868, opened 2019-06-11
Feature/gpu devices

### LINKED ISSUE #1655: GPU-CUs placement on nodes with multiple gpus
On Bridges there are multiple GPU devices per node. When multiple CUs are executed on the same node and each requires one GPU, RP schedules based on the available GPU devices. This, though, assumes that the executed task is able to find a free GPU device and execute there.

I think RP should be able to submit the units to different GPU devices. That can be done by setting the environment variable `CUDA_VISIBLE_DEVICES` to the number of GPU device the CU will use.

iparask:
This branch tries expand the CU environment with the GPU device(s) id(s) that the CU is using. The expansion should be moved to the scheduler.
It fixes #1655


iparask:
@andre-merzky, I would say that `_schedule_units` makes more sense. The slots have already been set and we can still set the environment variable. I think it is relatively straight forward to implement it for non-MPI units, but I'm not sure how does it translate for multinode units. In addition, if a launch method does not need such a variable, it can discard it and not include it.

Does it make sense to you?

andre-merzky:
> I would say that `_schedule_units` makes more sense

more sense than what?

I agree on the multinode tasks: if a task has GPU-0 on node-0, and GPU-1 on node-1, then a task-specific env setting does not get us far.  Hmpf, then we need to put it into the various launch methods after all.  Ugh, that will be messy... :/

iparask:
> more sense than what?
Yeah.... :)

I am not so sure. It can also go in the scheduler where we search for resources for non mpi units. That ensures that it is consistent between launch methods, and as soon as we find a use case that spans over nodes for GPUs we see then. 

As far as I know, none of the use cases I work for need multinode GPU usage. I remember there was one that did that, but I'm not sure what it was supposed to do.

andre-merzky:
Yeah, agree - we may want to wait for multi-node/multi-GPU use cases to finalize this.  For example, the scheduler may not set the env in that case, and leave it to the LM to do so if needed.

In that case in turn, the scheduler base class looks good again, don't you think?

iparask:
Yes, I agree! I was thinking of doing it in the `_alloc_nompi` method, but that does not exist in the base class at all. The next possible candidates are `_try_allocation` or `_schedule_units` but those two are agnostic of whether the unit is MPI or not.

Do you think it is correct semantically?

andre-merzky:
Yes, I think agnostic is fine.  For example, in `schedule_units` we could do:
```python
if unit.get('slots'):
  gpu_maps = list()
  for slot in unit['slots']:
    if slot['gpu_map'] not in gpu_maps:
      gpu_maps.append(slot['gpu_map'])
  if len(gpu_maps) == 1:
    # uniform GPU requirements
    unit['description']['environmen']['CUDA_DEVICES'] = \
        ','.join(gpi_map[0] for gpu_map in gpu_maps[0])
```
or something like this.

iparask:
I went to `_try_allocation` instead the `schedule_units` for two reasons. All schedulers use ` try_allocation` and only the `Continuous` overwrites it. Including it in `_schedule_units` would mean I should have add it in `_schedule_cb` and all the classes that overwrite those methods that would create a larger code replication.

andre-merzky:
Hmm, I am confused - I don't see where the Continuous scheduler overloads `_schedule_units`?  But `try_allocation` though is.  But you placed it in `allocate_slot` in the continuous (instead of try_allocation) - which is called multiple times for multi-process units, so the env selection is done again and again.   Moreover, `unit['slots']` is not assigned at this point. 

BTW: the `gpu_maps.append()` line has an indentation error in both cases.    Did you try to run this? ;-)

In summary: I think `_try_allocation` makes sense.  The only scheduler overloading it is the `NOOP` scheduler - and that does not define GPU maps in the first place.

iparask:
> Hmm, I am confused - I don't see where the Continuous scheduler overloads `_schedule_units`? But `try_allocation` is neither. But you placed it in `allocate_slot` in the continuous - which is called multiple times for multi-process units, so the env selection is done again and again. Moreover, `unit['slots']` is not assigned at this point.
I meant that only the Continuous scheduler overwrites `_try_allocate`.

> 
> BTW: the `gpu_maps.append()` line has an indentation error in both cases. Did you try to run this? ;-)
Oops. I fixed it.

It's missing a unit test

andre-merzky:
This has been iterated more than enough IMHO (and sorry for me picking on so many things) - I think it should be merged.  Also, other PRs now stall and depend on this one.  @iparask : opinion?

iparask:
Go with it!

PR #1872, opened 2019-06-16
Feature/bulk cb

andre-merzky:
This PR implements RP support for bulk callbacks.  Those are disabled by default, but can be enabled by `export RP_USE_BULK_CB=TRUE`.  The RP script in `examples/error_handling.py` shows an exemplary use of that feature.

@mturilli : Would you mind looking through the code and let me know where you think the changed behavior warrants new unit tests?  Once those are then added, I'll push the PR to @iparask for review.  Thanks :-)


andre-merzky:
@mturilli : ping

src/radical/pilot/compute_pilot.py
```diff
@@ -156,10 +153,12 @@ def _update(self, pilot_dict):
 
 
             try:
-                assert(rps._pilot_state_value(target) - rps._pilot_state_value(current)), \
-                            'invalid state transition'
-            except:
-                self._log.error('%s: invalid state transition %s -> %s', 
+                state_diff = rps._pilot_state_value(target) - \
+                             rps._pilot_state_value(current)
+                assert(state_diff), 'invalid state transition'
```
iparask: Hello @andre-merzky , I tried:
```python
>>> assert(2)
>>> try:assert(2)
... except: print 'False'
...
>>> try:assert(-1)
... except: print 'False'
...
>>> try:assert(0)
... except: print 'False'
...
False
```

Will the above behave as expected?

iparask: Shouldn't it be:
```python
assert(state_diff == 1)
```

andre-merzky: Ah, indeed - the test should be `assert(state_diff >= 1`, will fix!

andre-merzky: `== 1` was correct - no gaps are allowed at that point.

src/radical/pilot/unit_manager.py
```diff
@@ -278,66 +280,74 @@ def _pilot_state_cb(self, pilot, state):
...
+                                              'uid'   : {'$in'     : uids}},
+                                             {'$set'  : {'control' : 'umgr'}},
+                                             multi=True)
+                to_restart = list()
+                for unit in units:
+
+                    unit['state'] = rps.FAILED
+                    if not unit['description'].get('restartable'):
+                        self._log.debug('unit %s not restartable', unit['uid'])
+                        continue
```
iparask: Here you want to go to the next iteration of the look or exit the loop completely? If the former wouldn't an `else` statement make sense?

andre-merzky: Semantically there is no difference, and I am not a fan of nesting, specifically in Python.

iparask:
@andre-merzky, may I suggest to avoid small changes in log messages and comments if they are not necessary for the PR? Because those changes trigger flake8 and pylint for untouched files and it is very difficult to see why they are not passing.

andre-merzky:
> @andre-merzky, may I suggest to avoid small changes in log messages and comments if they are not necessary for the PR? Because those changes trigger flake8 and pylint for untouched files and it is very difficult to see why they are not passing.

The first commit in this PR is from August 2018, that is well before we agree on our `every file touched passes pylint` rule.  Well, that's my excuse for this PR at least, lets see what I can come up with for the others... ;-)

andre-merzky:
This is now ready for review.  Note that it currently needs an 
```sh
export RP_USE_BULK_CB=True
```
to show any effect.  `examples/error_handling.py` is a good test how bulk callbacks would be used.  For bulks to be active, configure RP bridges to use bulks, or deploy radical-cybertools/radical.utils/pull/218 which sets a decent default bulk size.

setup.py
```diff
@@ -139,7 +139,7 @@ def get_version(_mod_root):
 
 
 # ------------------------------------------------------------------------------
-# check python version. we need >= 2.7, <3.x
+# check python version. we need >= 3.5
```
mturilli: Should we use 3.6?

src/radical/pilot/compute_pilot.py
```diff
@@ -137,12 +137,16 @@ def __str__(self):
...
-            self._log.error("[Callback]: pilot '%s' failed - exit", self.uid)
+        uid   = self.uid
+        state = self.state
+
+        self._log.info("[Callback]: pilot %s state: %s.", uid, state)
+
+        if state == rps.FAILED and self._exit_on_error:
+            self._log.error("[Callback]: pilot '%s' failed (exit)", uid)
 
             # There are different ways to tell main...
```
mturilli: Is this still the case or do we have just one way now?

setup.py
andre-merzky: Ack, I changed that.

src/radical/pilot/compute_pilot.py
andre-merzky: The other ways are commented out below.  All of them have different trade-offs, that's and I want to keep the different options around in case we need to switch.  The current version is very reliable, but leads to a confusing `KeyboardInterrupt`, and uses an undocumented API call.  `raise` is unreliable, `os.kill` is very reliable, but kills the parent thread w/o warning or trace, `sys.exit` would be perfect, but internally uses signals which are unreliable in threaded applications.

andre-merzky:
Thanks for looking through this, @mtitov  and @mturilli !  Let me know if you want a walkthrough - the PR is very old by now and collected some cruft...

mtitov:
> Well spotted! Should we do a global search and replace for this?

addressed here #2083

PR #1874, opened 2019-06-16
Fix/cuda

andre-merzky:
This adds cuda support - but branch needs some cleaning up before merge


bin/radical-pilot-create-static-ve
```diff
@@ -112,11 +112,16 @@ mkdir -p "$prefix"
 cd $prefix
 
 echo -n "install virtualenv "
-VIRTENV_TGZ="virtualenv-1.9.tar.gz"
-VIRTENV_TGZ_URL="https://pypi.python.org/packages/source/v/virtualenv/$VIRTENV_TGZ"
-curl -k -L -O "$VIRTENV_TGZ_URL" 2>&1 | progress
-tar zxmf "$VIRTENV_TGZ"
-VIRTENV_CMD="$PYTHON virtualenv-1.9/virtualenv.py"
+
+VIRTENV_CMD="$(which virtualenv 2>/dev/null)"
```
iparask: Isn't this change coming from devel?

src/radical/pilot/agent/executing/popen.py
```diff
@@ -201,6 +201,8 @@ def _handle_unit(self, cu):
             # Start a new subprocess to launch the unit
             self.spawn(launcher=launcher, cu=cu)
 
+            time.sleep(1)
```
iparask: Why is this delay needed? This also assumes that all CUs will be more than a second.

bin/radical-pilot-create-static-ve
andre-merzky: I re-synced with `feature/summit`, this is clean now.

src/radical/pilot/agent/executing/popen.py
andre-merzky: Thanks, removed.

tests/test_lm/test_unit/test_jsrun.py
```diff
@@ -46,18 +46,20 @@ def test_create_resource_set_file(mocked_init, mocked_method, mocked_raise_on):
 
 # ------------------------------------------------------------------------------
 #
+def _create_resource_set_file(self, slots, uid, sandbox):
+    return 'rs_layout_cu_%06d' % (int(uid.split('_')[1]))
 @mock.patch.object(JSRUN, '__init__', return_value=None)
 @mock.patch.object(JSRUN, '_configure', return_value='jsrun')
 @mock.patch.object(JSRUN, '_create_resource_set_file',
-                          return_value='rs_layout_cu_000000')
+                           _create_resource_set_file)
```
mturilli: Is this a repetition?

tests/test_summit/test_rm_create_on_localhost.py
```diff
@@ -53,6 +54,8 @@ def test_rm_create_on_localhost():
     import pprint
     pprint.pprint(lrms.lrms_info)
     assert lrms.lrms_info == {'agent_nodes'     : {},
+                              'gpus_per_node'   : 6,
+                              'cores_per_node'  : 20,
```
mturilli: Should this be 40?

mturilli:
Why we don't have Travis to test this?

tests/test_lm/test_unit/test_jsrun.py
andre-merzky: No - basically it mocks the member method `_create_resource_set_file` with the local method of the same name defined just above the call (the previous mock was wrong).  Thanks for checking! :-)

tests/test_summit/test_rm_create_on_localhost.py
andre-merzky: Good point - I think the test case still comes from summitdev.  The actual values don't matter though, as they are part of the test case.  As long as the values match, the method is doing the right thing.

andre-merzky:
Uh, I don't know.  @iparask : any idea?

PR #1877, opened 2019-06-16
Integration/molssi

andre-merzky:
This PR adds an exemplary molssi integration script.


andre-merzky:
@mturilli : please review - thanks

PR #1878, opened 2019-06-16
Fix/hombre 2

andre-merzky:
This PR simplifies (and in special cases fixes) the HOMBRE scheduler.


andre-merzky:
Pfft, no idea why I opened this PR twice - sorry...

PR #1882, opened 2019-06-20
Fix/cuda

andre-merzky:
Bugger, I closed #1874 by accident it seems.  Well, here it is again...


andre-merzky:
All comments from #1874 are addressed by now.

src/radical/pilot/agent/agent_0.py
```diff
@@ -351,29 +354,36 @@ def _start_sub_agents(self):
...
-                slots['gpu_threads'  ] = 0
-                slots['nodes'        ] = [{'name'    : node[0],
-                                           'uid'     : node[1],
-                                           'core_map': [[0]],
-                                           'gpu_map' : [],
-                                           'lfs'     : {'path': '/tmp', 'size': 0}
-                                          }]
+                bs_name = "%s/bootstrap_2.sh" % (self._pwd, sa)
+                ls_name = "%s/%s.sh" % (self._pwd, sa)
+                slots = {
```
iparask: Here you give a slot to the agent, right?

andre-merzky: Indeed, in case a subagent needs to be placed on a compute node.  The RM will make sure that the respective node is blocked from the scheduler.

Issue #1881, opened 2019-06-18
create list of files to be covered by unit testing


andre-merzky:

andre-merzky:
(roughly) sorted by priority:
```
src/radical/pilot/states.py
src/radical/pilot/utils/component.py

src/radical/pilot/agent/agent_0.py
src/radical/pilot/agent/agent_n.py
src/radical/pilot/worker/update.py

src/radical/pilot/session.py
src/radical/pilot/pilot_manager.py
src/radical/pilot/unit_manager.py
src/radical/pilot/compute_pilot.py
src/radical/pilot/compute_unit.py

src/radical/pilot/umgr/scheduler/base.py
src/radical/pilot/umgr/scheduler/round_robin.py
src/radical/pilot/umgr/scheduler/backfilling.py

src/radical/pilot/agent/lm/base.py
src/radical/pilot/agent/lm/fork.py
src/radical/pilot/agent/lm/mpiexec.py
src/radical/pilot/agent/lm/prun.py
src/radical/pilot/agent/lm/aprun.py
src/radical/pilot/agent/lm/ssh.py

src/radical/pilot/agent/rm/base.py
src/radical/pilot/agent/rm/slurm.py
src/radical/pilot/agent/rm/lsf.py
src/radical/pilot/agent/rm/fork.py
src/radical/pilot/agent/rm/torque.py
src/radical/pilot/agent/rm/pbspro.py

src/radical/pilot/agent/scheduler/base.py
src/radical/pilot/agent/scheduler/hombre.py
src/radical/pilot/agent/scheduler/continuous.py
src/radical/pilot/agent/scheduler/continuous_ordered.py

src/radical/pilot/agent/executing/base.py
src/radical/pilot/agent/executing/popen.py
src/radical/pilot/agent/executing/shell_fs.py

src/radical/pilot/pmgr/launching/base.py
src/radical/pilot/pmgr/launching/default.py
```

iparask:
Okay! Thank you! Should we use the structure of `feature/rm_test` branch or what exists in devel?

AymenFJA:
Thanks @andre-merzky .

andre-merzky:
> Okay! Thank you! Should we use the structure of `feature/rm_test` branch or what exists in devel?

Well, I guess reuse is great, but happy to leave this up to the two of you to judge in terms of effort...

iparask:
@andre-merzky is prun the same as `prte.py`?

[...]

PR #1884, opened 2019-06-20
Feature/rm tests

iparask:
@andre-merzky what do you think? I'm trying to finish the base class of `lm`


src/radical/pilot/states.py
```diff
@@ -41,13 +41,14 @@
         CANCELED               :  5}
 _pilot_state_inv   = {v: k for k, v in _pilot_state_values.iteritems()}
 _pilot_state_inv_full = dict()
-for s,v in _pilot_state_values.iteritems():
+for st,v in _pilot_state_values.iteritems():
     if v in _pilot_state_inv_full:
```
andre-merzky: why the change from s to st?

tests/test_states.py
```diff
@@ -0,0 +1,65 @@
...
+def test_pilot_state_value():
+    from radical.pilot.states import _pilot_state_value
+    assert(_pilot_state_value('NEW')                    == 0)
+    assert(_pilot_state_value('PMGR_LAUNCHING_PENDING') == 1)
+    assert(_pilot_state_value('PMGR_LAUNCHING')         == 2)
+    assert(_pilot_state_value('PMGR_ACTIVE_PENDING')    == 3)
+    assert(_pilot_state_value('PMGR_ACTIVE')            == 4)
+    assert(_pilot_state_value('DONE')                   == 5)
+    assert(_pilot_state_value('FAILED')                 == 5)
+    assert(_pilot_state_value('CANCELED')               == 5)
```
andre-merzky: To be honest, I don't find these tests particularly useful.  The numbers we assign are not really special, so these tests basically assert that `a=b` results in `a=b` - duh ;-)  Not that I have an educated opinion what types of tests would make sense in this context, just saying...

iparask: Yeah, but this is the first file in issue #1881. And apparently it is the easiest. I agree with you it is a bit weird to test files that create constants only.

src/radical/pilot/states.py
iparask: The linter was complaining because `s` is used also as a parameter to a function.

andre-merzky: Why the hell is it complaining about that?
```
   list.append(s)
   list[len(list)] = s
```
would be equivalent, but it would not complain about the latter?  That does not make sense...  Lets not follow the linter blindly, please...

tests/test_states.py
andre-merzky: There are a couple of methods there which are supposed to ensure correct state progression.  Those are rather central to how we define and ensure states.  The `collapse` is tested, we may want to add test to the `progress` versions, too?

tests/test_states.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+# ------------------------------------------------------------------------------
+#
+def test_pilot_state_collapse():
+    from radical.pilot.states import _pilot_state_collapse,_pilot_state_values
+
+    for key in _pilot_state_values.iterkeys():
+        assert(_pilot_state_collapse([key]) == key)
+# ------------------------------------------------------------------------------
```
andre-merzky: Looking over this again, I don't think this test covers the right cases.  The method is supposed to select the most current state out of a list of states.  Assume a unit gets state updates out of order (as it happens) - what is our best guess in what the state the unit is in?

```
[NEW, UMGR_STAGING_INPUT, FAILED, UMGR_STAGING_INPUT_PENDING] -> FAILED
[NEW, EXECUTING, UMGR_STAGING_INPUT, UMGR_STAGING_INPUT_PENDING] -> EXECUTING
```
etc.

iparask: Aha! Thank you! I will update the test with a few test cases then that have states out of order and provide the last one!

tests/test_states.py
iparask: Okay!

src/radical/pilot/states.py
iparask: The issue was this:
```
for s,v in _pilot_state_values.iteritems():
    if v in _pilot_state_inv_full:
        if not isinstance(_pilot_state_inv_full[v], list):
            _pilot_state_inv_full[v] = [_pilot_state_inv_full[v]]
        _pilot_state_inv_full[v].append(s)
    else:
        _pilot_state_inv_full[v] = s


def _pilot_state_value(s):
    return _pilot_state_values[s]
```
And this was the linter's message:
`W: 53,23: Redefining name 's' from outer scope (line 44) (redefined-outer-name)`

We can take it out if you think this is too much.

tests/test_states.py
andre-merzky: Thank you, @iparask !

src/radical/pilot/states.py
andre-merzky: yes, thank you.

tests/test_lm/test_unit/test_rsh.py
```diff
@@ -62,6 +62,7 @@ def test_construct_command(mocked_init,
     component.launch_command = 'rsh'
 
     for unit, result in test_cases:
+        print unit['uid']
         if result == "ValueError":
```
andre-merzky: This can probably go?

tests/test_states.py
andre-merzky: Should this still be addressed or  will that go into a different PR?  If the latter, please open a ticket to keep track of it.

tests/test_states.py
andre-merzky: Should this still be addressed or  will that go into a different PR?  If the latter, please open a ticket to keep track of it.

andre-merzky:
@iparask : ping

tests/test_lm/test_unit/test_rsh.py
iparask: Yes!

tests/test_states.py
iparask: Fixing it!

iparask:
I think it is ready. Please check once more.

iparask:
Hello @andre-merzky, would you like to have another look?

tests/test_cases/unit.000002.json
```diff
@@ -28,7 +28,8 @@
         "lm": {
             "spark"  : ["/spark-submit    test.py", null],
             "ibrun"  : "RuntimeError",
-            "orte"   : "RuntimeError"
+            "orte"   : "RuntimeError",
```
mturilli: Should we discontinue ORTE altogether?

andre-merzky:
Thanks for all the work!

PR #1889, opened 2019-06-27
Fixing travis to separate test lint form testing

iparask:
The proposed change does the following three things.

1. Sets up devel as the testing branch
2. Lints only for the changed python files
3. Separates testing from linting tests, which caused travis to succeed when tests fail.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,3 @@
-
 language: python
 sudo: disabled
 
@@ -6,7 +5,7 @@ python:
   - "2.7"
 
 branches:
-  - feature/lm_tests
+  - devel
 
 os:
   - linux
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -18,16 +17,26 @@ env:
...
+    - DIFF=`git diff --name-only --diff-filter=b $OLD...$NEW`
+    - DIFF=$(echo $DIFF | grep -o -e '\b[^ ]*.py\b')
     - LOC=/home/travis/virtualenv/python2.7  # Location where VE is created on travis
-    - CMD1="coverage run -m pytest -vvv tests/;flake8 tests/;pylint tests/"
-    - CMD2="flake8 --config=.flake8 $LOC/lib/python2.7/site-packages/radical/pilot/agent/lm/;flake8 --config=.flake8 $LOC/lib/python2.7/site-packages/radical/pilot/agent/rm/;"
-    - CMD3="pylint $LOC/lib/python2.7/site-packages/radical/pilot/agent/lm/;pylint $LOC/lib/python2.7/site-packages/radical/pilot/agent/rm/"
+    - CMD_PYTEST="coverage run -m pytest -vvv tests/"
+    - CMD_FLAKE8="if ! test -z \"$DIFF\"; then flake8 --config=.flake8 $DIFF;else echo 'Nothing to pep8'; fi"
+    - CMD_PYLINT="if ! test -z \"$DIFF\"; then pylint $DIFF;else echo 'Nothing to lint';fi"
+    - CMD_PYTEST="coverage run -m pytest -vvv tests/;flake8 tests/;pylint tests/"
```
andre-merzky: Why is this redefining `CMD_PYTEST` from above?  It also duplicates the test specific linters below I believe?

andre-merzky:
Ironically, the tests seems to fail because of DBURL settings...

.travis.yml
iparask: crap... This should be removed. Thank you for catching it!

PR #1892, opened 2019-06-28
Fix/gpu chunk

andre-merzky:
the `_find_resources` method missed to clean out partial chunks from the set of found cores and gpus.  This patch fixes this problem.


src/radical/pilot/agent/scheduler/continuous.py
```diff
@@ -468,14 +490,15 @@ def _alloc_nompi(self, unit):
...
-                                                    lfs_chunk=lfs_chunk,
-                                                    core_chunk=core_chunk,
-                                                    mem_chunk=mem_chunk)
+                                                         requested_cores=total_cores,
+                                                         requested_gpus=total_gpus,
+                                                         requested_lfs=total_lfs,
+                                                         requested_mem=total_mem,
+                                                         core_chunk=core_chunk,
+                                                         partial=False,
+                                                         gpu_chunk=1,
```
iparask: Why is `gpu_chunk` set to 1?

andre-merzky: because we do not yet really have anything like GPU threads (and per our discussion on task description its unclear if that will ever happen), so we have a different co-locality notion than for cores (where threads require multiple cores in one chunk, i.e. on the same node).

andre-merzky:
@iparask : please complete review - thanks

andre-merzky:
> Can we get travis to flake8 and pylint for the continuous scheduler and pass? 

Yep, I can take care of that.

andre-merzky:
Done.

mturilli:
@iparask , @AymenFJA : we have several errors in the tests, both in terms of code bugs (e.g., `test_jsrun.py, line 51,  fixture 'mocked_raise_on' not found`) and actual failed asserts. Should we ignore these tests at the moment as 'not ready' for production?

mturilli:
@iparask , @AymenFJA : I noticed that our mongodb on one.radical chokes with   `E -        [ftdc] Assertion: 13538:couldn't open [/proc/30927/stat] Too many open files src/mongo/util/processinfo_linux.cpp 76` This is a fatal error that crashes our server. Log inspection shows that this happens after running multiple times your RP tests. Upon further testing, it seems that you tests (possibly when they fail or always) do not close their connections to mongodb. Connections pile up until the server is crashed with the above error. Could you please investigate? Note that once the server crashes, all your tests that depend from mongodb fail because of the server unavailability. You may want to fail ‘gracefully’ when that happens because the test itself was never run and so its failure tells us nothing about RP.

AymenFJA:
> @iparask , @AymenFJA : we have several errors in the tests, both in terms of code bugs (e.g., `test_jsrun.py, line 51, fixture 'mocked_raise_on' not found`) and actual failed asserts. Should we ignore these tests at the moment as 'not ready' for production?

I don't think we should ignore the testing, we can converge on the code again and try to fix it, to be 'ready' for the production stage.

mturilli:
> I don't think we should ignore the testing,

I meant for this specific PR. Otherwise, I agree with you: we should get this to work reliably. Personally, I think these tests are great and will make a huge different for the next release alongside the upcoming integration tests.

PR #1894, opened 2019-07-01
Fix/jsrun

andre-merzky:
This PR fixes a  syntax error in the jsrun LM which crept in via a recent merge, and also fixes the evaluation of RP profile settings in the executors, re-enabling task level profiling,


andre-merzky:
> I did not find where _profile is created in RP code.

The `self._prof` member is present in all components, and is defined in the [component base class](https://github.com/radical-cybertools/radical.pilot/blob/devel/src/radical/pilot/utils/component.py#L324) (along with a logger and reporter)

PR #1897, opened 2019-07-02
Fix/summit prte 2

andre-merzky:
This PR changes the way the PRTE DVM reports the HNP URI.  That was previously parsed out of the hnp process output - but that parsing yields an incorrect result in some debug modes.  The URI is now communicated via the FS, which is slower but more reliable.   The DVM setup now also uses a hostfile instead of relying on LSF env information for node selection.

### DIFF
--- a/src/radical/pilot/agent/lm/prte.py
+++ b/src/radical/pilot/agent/lm/prte.py
@@ -39,7 +39,7 @@ class PRTE(LaunchMethod):
             raise Exception("Couldn't find prte")
 
         # Now that we found the prte, get PRUN version
-        out, err, ret = ru.sh_callout('prte_info | grep "Open RTE"', shell=True)
+        out, _, _ = ru.sh_callout('prte_info | grep "Open RTE"', shell=True)
         prte_info = dict()
         for line in out.split('\n'):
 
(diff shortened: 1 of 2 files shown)


andre-merzky:
@iparask : please review - thanks

src/radical/pilot/agent/lm/prte.py
```diff
@@ -81,41 +96,38 @@ def lrms_config_hook(cls, name, cfg, lrms, logger, profiler):
...
+                        if '://' in line:
+                            dvm_uri = line.strip()
+                            break
 
-            elif 'DVM ready' in line:
-                if not dvm_uri:
-                    raise Exception("VMURI not found!")
+            except Exception as e:
+                logger.debug('DVM check: %s' % e)
+                time.sleep(0.5)
```
mturilli: See my comment to this code on PR #1902

andre-merzky: Ack, same reply.

andre-merzky:
> Can we get travis to check flake8 and pylint on this file? the rm_tests branch also has a test for this launch method. It would be nice to make sure that it still passes.

Sure - pylint and flake8 both pass on the prte LM.

Thanks both!

PR #1899, opened 2019-07-03
let base class know about uniform workloads

andre-merzky:
This is crucial for optimal hombre performance.


andre-merzky:
@mturilli : please review

PR #1900, opened 2019-07-03
fix GPU maps

andre-merzky:
This fixes two small errors related to `CUDA_VISIBLE_DEVICE` settings (slot dict was incorrectly accessed) and gpu mapping in the hombre scheduler (slot dict was incorrectly created).


iparask:
Did you branch out of the `feature/gpu_devices` branch or `devel`? It seems like the former, more than the later.

andre-merzky:
@iparask : this PR is by now fairly small and contained - it basically just moves code around and fixes the affected parts in the hombre scheduler.  Can you please give it another checkup, so that we can close this out?  Thanks!

iparask:
I assume flake8 and pylint passes, right?

andre-merzky:
Pylint is not yet there, I'll have to address this in a separate commit - apologies...

PR #1901, opened 2019-07-03
fix jsrun env export and summit agent config

andre-merzky:
jsrun fails if an -E export is not set

### DIFF
--- a/src/radical/pilot/agent/lm/jsrun.py
+++ b/src/radical/pilot/agent/lm/jsrun.py
@@ -2,6 +2,7 @@
 __copyright__ = "Copyright 2016, http://radical.rutgers.edu"
 __license__   = "MIT"
 
+import os
 
 import radical.utils as ru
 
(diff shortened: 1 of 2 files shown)


andre-merzky:
@mturilli : please review

src/radical/pilot/agent/lm/jsrun.py
```diff
@@ -120,6 +121,11 @@ def construct_command(self, cu, launch_script_hop):
         env_list   = self.EXPORT_ENV_VARIABLES + task_env.keys()
         env_string = ' '.join(['-E "%s"' % var for var in env_list])
 
+        # jsrun fails if an -E export is not set
+        for var in env_list:
+            if var not in os.environ:
+                os.environ[var] = ''
+
```
mturilli: This was commented out in a subsequent PR. Was that intentional?

andre-merzky: Also a good catch, thanks for being thorough! 
 
Yes, that was intentional.  The above indeed ensures that the `-E` directive in jsrun does not screw our job when variables are unset - but I later found some cases where even valid env settings could kill our job, specifically related to module loads (`LD_LIBRARY_PATH`).  I thus disabled some parts of the env handling completely until we have a better understanding of this problem, but I wanted to keep the original code around.

PR #1902, opened 2019-07-03
Fix/summit prte

andre-merzky:
prte fixes for summit

The PR is a little bigger than it should be as it already contains the jsrun env fixes from #1901


iparask:
This also does not look like a branch from devel. @andre-merzky is this intentional or just a simple mistake?

andre-merzky:
> This also does not look like a branch from devel. @andre-merzky is this intentional or just a simple mistake?

mistake - too many small things going on, and it's not always easy to branch commits out into independent branches.  Apologies...

andre-merzky:
The PR text so explicitly states: `The PR is a little bigger than it should be as it already contains the jsrun env fixes from #1901` :-)

src/radical/pilot/agent/scheduler/continuous.py
```diff
@@ -544,13 +563,16 @@ def _alloc_mpi(self, unit):
...
         requested_gpus   = cud['gpu_processes']
+        gpu_chunkg       = cud['gpu_threads']
         lfs_per_process  = cud['lfs_per_process']
         mem_per_process  = cud['mem_per_process']
 
         # make sure that processes are at least single-threaded
         if not threads_per_proc:
             threads_per_proc = 1
 
+        if not gpu_chunk:  gpu_chunk  = 1
```
iparask: There is another PR that introduces this, right?

andre-merzky: Yes, some PRs are overlapping by now.  That will be resolved when merging back from devel though...

examples/config.json
```diff
@@ -268,27 +268,18 @@
...
-    "ornl.summit"  : {
-        "project"  : "BIP178",
-        "queue"    : "batch",
-        "schema"   : "local",
-        "cores"    : 84,
-        "gpus"     : 12
-    },
-
     "local.summit" : {
         "project"  : "BIP178",
```
mturilli: This requires a costly work around from the user. E.g., sed -i 's/BIP178/geo111/g' [...]/lib/python2.7/site-packages/radical/pilot/configs/resource_ornl.json Do we have a path towards eliminating this hard coded allocation value? (also in #1904)

examples/hello_rp.sh
```diff
@@ -0,0 +1,98 @@
+#!/bin/sh
```
mturilli: I think the name of the example is too generic and confusing as we have already `helloworld_mpi.py` and `helloworld_threads.py`. Maybe `helloworld_cu_startup.py`? Don't like it much but I did not want to complain without leaving a positive suggestion :)

src/radical/pilot/agent/lm/prte.py
```diff
@@ -81,41 +96,38 @@ def lrms_config_hook(cls, name, cfg, lrms, logger, profiler):
...
-
-                dvm_uri = line.strip()
-                logger.info("prte uri: %s" % dvm_uri)
+            time.sleep(0.1)
+            try:
+                with open(furi, 'r') as fin:
+                    for line in fin.readlines():
+                        if '://' in line:
+                            dvm_uri = line.strip()
+                            break
```
mturilli: This is ugly :) Is there a way to get notified/pushed instead of trying 100 times and then assume sleep a bit more if nothing has happened?

src/radical/pilot/agent/lm/prte.py
```diff
@@ -139,6 +151,7 @@ def _watch_dvm():
                 # swallowed, the next `prun` call will trigger
                 # termination anyway.
                 os.kill(os.getpid())
+                raise RuntimeError('PRTE DVM dies')
```
mturilli: s/dies/died/ ?

src/radical/pilot/configs/agent_summit.json
```diff
@@ -34,15 +34,16 @@
...
-            "target": "node",
-            "components": {
-                "AgentExecutingComponent" : {"count" : 1}
-            }
-        }
+        "AgentStagingOutputComponent" : {"count" : 1},
+        "AgentExecutingComponent"     : {"count" : 1}
     }
+    
+  # "agents": {
```
mturilli: This assumes our RU comment-augmented json right?

src/radical/pilot/configs/resource_local.json
```diff
@@ -116,7 +116,7 @@
         "lfs_path_per_node"           : "/tmp",
         "lfs_size_per_node"           : 1024,
         "cores_per_node"              : 8,
-        "gpus_per_node"               : 1
+        "gpus_per_node"               : 2
```
mturilli: Is this going to create a problem if `localhost` has indeed only 1 GPU as most of the "usual" localhost have?

src/radical/pilot/configs/resource_ornl.json
```diff
@@ -318,7 +318,7 @@
         "agent_config"                : "default",
         "agent_scheduler"             : "CONTINUOUS",
         "agent_spawner"               : "POPEN",
-        "agent_launch_method"         : "PRTE",
+        "agent_launch_method"         : "JSRUN",
```
mturilli: Is this intentional? I was confused by this being a PRTE PR.

examples/config.json
andre-merzky: At the moment we do not have the ability to fix that.  I don't like the fact that the examples depend on that configuration file in general, tbh - it makes it more difficult to run examples for end users, and as you observe, the data get out of sync anyway.  I think the main reason to have that config in place is *for us* to be able to easily run the examples, and to avoid the need to do the same changes in several files.

I see two approaches to alleviating the issue.  First, I think that some example configurations should become part of the resource config, like default access schema, number of cores to use for examples, etc.  Secondly, a user configuration file would make sense,  which lists preferred access schema for all resources, preferred queues, user names, allocations, etc.

Both are fairly straight forward to implement - but both require significant coding I think.  The configuration restructuring in v.2 will make this (somewhat) simpler though.

examples/hello_rp.sh
andre-merzky: hehe.  Well, the script probably should go toward integration testing really, it is not very illustrative of a user workload, and thus not a good example per se.  OTOH, it does a thorough job on inspecting task setup and placement, and can be used to confirm scheduler placement (which is what I mainly use it for) - but I did not have a better place to put it.  `tests/scripts/` maybe?

src/radical/pilot/agent/lm/prte.py
andre-merzky: This is ugly, but unfortunately more reliable than parsing the (buffered) output stream of the HNP process.  The latter also reveals the URL eventually, but specifically if debug settings are enabled, the URL is well hidden in lots of noisy output, and we would need a more clever parser.

In principle we should be able to wait for a `DVM ready` message on the HNP output stream and should then expect to just find a valid URL in that file.  That actually was implemented before.  Alas, that also turned out to be not very reliable :-/

Summary: yes, ugly, but I don't see a better way right now.

src/radical/pilot/agent/lm/prte.py
andre-merzky: thanks, fixed.

src/radical/pilot/configs/agent_summit.json
andre-merzky: That is correct - all json's consumed by RP are read via `ru.read_json` which makes sure of that.

src/radical/pilot/configs/resource_local.json
andre-merzky: It should not - but let me convert it.

BTW, might be good to know: if `RADICAL_DEBUG` is set, you can request any number of cpus, cores etc. from localhost - the FORK RM will consider the machine to be anything you like.  But since this is not the default setting (and should not be), let me fix this.  Thanks.

andre-merzky: BTW: we configure localhost also to have 8 cores, which is not common - but let's leave it until it becomes a problem, ok?

src/radical/pilot/configs/resource_ornl.json
andre-merzky: Ah, good catch - but yes, this is intentional.  The `agent_launch_method` is used to place sub-agents onto compute nodes, and jsrun is a good choice for doing so, in case we use agent configurations where the PRTE DVM and the HNP also runs on a compute node - in that case, we cannot use prun to place the sub-agent there, as the DVM would not be alive at that point.

src/radical/pilot/agent/lm/prte.py
mturilli: Thanks. Maybe something to escalate to the PRTE developers?

examples/hello_rp.sh
mturilli: I was just thinking about naming but yes, `tests/scripts/` seems good. @lee212 maybe a useful starting point for some of your integration tests?

PR #1904, opened 2019-07-03
resolve cherry-picked env fix

andre-merzky:


andre-merzky:
No, the experiment branch should never be released.  Also, the fix is in that branch, and this PR cherry-picked it out of there... ;-)

andre-merzky:
Done.

examples/config.json
```diff
@@ -276,14 +276,6 @@
...
-    "local.summit" : {
-        "project"  : "BIP178",
-        "queue"    : "batch",
-        "schema"   : "local",
-        "cores"    : 84,
-        "gpus"     : 12
-    },
-
     "ornl.summit_prte" : {
         "project"  : "BIP178",
```
mturilli: This requires a costly work around from the user. E.g., `sed -i 's/BIP178/geo111/g' [...]/lib/python2.7/site-packages/radical/pilot/configs/resource_ornl.json` Do we have a path towards eliminating this hard coded allocation value?

src/radical/pilot/agent/lm/jsrun.py
```diff
@@ -117,8 +116,15 @@ def construct_command(self, cu, launch_script_hop):
...
-        env_string = ' '.join(['-E "%s"' % var for var in env_list])
+        env_string = ''
+      # task_env   = cud.get('environment') or dict()
+      # env_list   = self.EXPORT_ENV_VARIABLES + task_env.keys()
+      # env_string = ' '.join(['-E "%s"' % var for var in env_list])
+      #
+      # # jsrun fails if an -E export is not set
+      # for var in env_list:
+      #     if var not in os.environ:
+      #         os.environ[var] = ''
```
mturilli: This is commented out because `jsrun` does not fail anymore without -E?

andre-merzky: This is the counterpart to the comment in the other PR: here we disable `-E` completely in the lines before because it did not play well with some `module load` directives.  For now we rely on env settings in the unit shell script and on jsrun to export the right environment by default.  This may need more work later, but for now I would like to gather some data on how this behaves for actual workloads.

examples/config.json
andre-merzky: Well, this is the example config, not the resource config - so the user should not need to change the resource config file (unless I am missing something).

But FWIW, settings in the resource config can be overwritten by placing a file in `~/.radical/pilot/configs/resource_ornl.json` - any keys defined there would overwrite the installed resource config.

PR #1906, opened 2019-07-04
install events.md

andre-merzky:
it is needed by RA for session checks


andre-merzky:
@iparask : please review - thanks

PR #1907, opened 2019-07-04
fix some tests

andre-merzky:


andre-merzky:
This fixes some of the tests which went out of sync with code.  The remaining tests seem to fail because of resource limitations.  The hombre test fails because, ironically, hombre has been fixed - the test encoded incorrect behavior.  I am hesitant to fix that test as it will likely need to be rewritten with the other scheduler tests anyway...  Opinion?

andre-merzky:
@iparask @AymenFJA : please review - thanks

PR #1909, opened 2019-07-06
Rc/v0.70.0

andre-merzky:


setup.py
```diff
@@ -233,31 +251,27 @@ def read(fname):
...
-  # },
-  # 'upload_sphinx'      : {
-  #     'upload-dir'     : 'docs/build/html',
-  # },
-    # This copies the contents of the examples/ dir under
-    # sys.prefix/share/$name
-    # It needs the MANIFEST.in entries to work.
-  # 'data_files'         : makeDataFiles('share/%s/examples/' % name, 'examples'),
     'data_files'         : df,
+    'cmdclass'           : {'upload': RunTwine},
```
mturilli: Did not find what this is.

andre-merzky: calling `python setup.py upload` then pushes the module to pypi when this is defined, removing some manual steps in the release process.  Twine is the python tool which handles the pypi upload and some checks while doing so.

andre-merzky:
Apologies, this PR pointed  to the wrong target: it is a snapshot of `devel` which forms the release candidate and thus goes toward `master`.  I created this PR so that the travis tests are triggered.

PR #1913, opened 2019-07-12
Updating how `get_mpi_info` checks for Open MPI

iparask:
This proposes a change in the LM base class to fix how `get_mpi_info` checks for OpenMPI versions.


src/radical/pilot/agent/lm/base.py
```diff
@@ -167,9 +169,11 @@ def create(cls, name, cfg, session):
             return impl(name, cfg, session)
 
         except KeyError:
+            # pylint: disable=protected-access
             session._log.exception("LM '%s' unknown or defunct" % name)
 
         except Exception as e:
+            # pylint: disable=protected-access
```
iparask: @andre-merzky, I am not sure why it uses the session log and not its own. Would it make sense to change it?

andre-merzky: because we have quite a number of logfiles already, and I did not want to go a granularity beyond one per component.  The LM is not a separate component after all, so can well reuse the logger of it's parent component (the executor).

andre-merzky: This should use `self._log` as defined in the ctor instead of adding the pyling exception.

src/radical/pilot/agent/lm/base.py
```diff
@@ -86,6 +87,7 @@ def __init__(self, name, cfg, session):
         self.env_removables = []
 
         self._configure()
+        # pylint: enable=protected-access
```
andre-merzky: It's a shame that python has no granularity beyond `public/private`...  The session logger is the right thing to use here, but the logger should not be part of the application facing API.  Something like a C++ `friend` is what we would need here.  So we add two linter lines for one code line :-P  Oh well...

src/radical/pilot/agent/lm/base.py
andre-merzky: Ah, I missed that  this is in `create()`, which is a `@classmethod`, so there is not even something like `it's own logger`, since we have no `self` here.  We would need to create a new logger every time this method is called, only to log one message...

iparask: Ah I see. I thought that when create is called, the constructor is also called. Thank you for explaining that to me!

andre-merzky: Ah, understood - welcome!  A class method really is like a free function which happens to live in the namespace of the class, but does not share any data with any class instance.  It only makes sense for methods which are semantically bound to the class, but cannot be member methods - like, in this case, a factory method...

PR #1916, opened 2019-07-12
Archiving ORTE

### LINKED ISSUE #1915: Archive ORTE LM
We should archive the ORTE launcher since it is not used anymore

iparask:
This PR fixes #1915.

It archives ORTErun and ORTELIB, and removes their respective tests


andre-merzky:
RU's json parser supports comments - would it make sense to just comment-out the ORTE test cases?  I don't mind much if not, git is archiving them after all.  Thoughts?

iparask:
I did not know that! I do not mind commenting out the JSON entry as much as what we should do about the test. In this case here the test is deleted, so I am not sure it makes sense to have the JSON entry as well. If we are to archive the test then commenting the JSON entry makes sense.

andre-merzky:
> I did not know that!

That's the main reason why we wrote a wrapper around `json.read()` ;-)

>  I do not mind commenting out the JSON entry as much as what we should do about the test. In this case here the test is deleted, so I am not sure it makes sense to have the JSON entry as well. If we are to archive the test then commenting the JSON entry makes sense.

Well, the test itself is easily recreated (it is very similar to the other tests) - but the created ORTE command lines are non-trivial to re-create.  I mostly care because we need to add tests for PRTE, and those command lines will look very similar :-D.   So, I mostly care about the test cases captured in the json files.  But either way, I don't feel strong about it, no worries - so either way works for me...

PR #1919, opened 2019-07-15
Feature/rm integration tests

lee212:
I've made 1st attempt for rm integration test. I'd like to ask reviews, and further commit to fix/update will follow.


tests/test_rm/integration/test_rm_int.py
```diff
@@ -0,0 +1,178 @@
...
+    import mock
+except ImportError:
+    from unittest import mock
+
+# ------------------------------------------------------------------------------
+#
+def setUp(resource):
+
+    session      = rp.Session()
+    cfg          = session.get_resource_config(resource=resource)
```
iparask: Can I ask that the sessions are mocked? I believe that reading just the configuration should be enough. @andre-merzky can we read a config file somehow by providing the resource label without having to create a session?

tests/test_rm/integration/test_rm_int.py
```diff
@@ -0,0 +1,178 @@
...
+                              'cores_per_node'  : 8,
+                              'lfs_per_node'    : {'path': "/tmp", 'size': 1024},
+                              'lm_info'         : {'version_info'    : {'FORK':
+                                  {'version': '0.42', 'version_detail': 
+                                  'There is no spoon'}}},
+                              'mem_per_node'    : 0,
+                              'name'            : 'Fork',
+                              'node_list'       : [['localhost', 'localhost_0']]}
+    lrms.stop()
+    session.close()
```
iparask: This can be moved in `tearDown`.

tests/test_rm/integration/test_rm_int.py
```diff
@@ -0,0 +1,178 @@
...
+    tearDown()
+
+
+@mock.patch.object(PBSPro, '_parse_pbspro_vnodes', return_value=['nodes1', 'nodes2'])
```
iparask: I do not think `_parse_pbspro_vnodes` should be mocked. I suggest that `subprocess.check_output` gets mocked as done [here](https://github.com/radical-cybertools/radical.pilot/blob/devel/tests/test_rm/test_unit/test_pbspro.py#L126) and let PBSPro to run fully.

iparask:
Some final thoughts: I like where this is heading and I would way that it is an overall good job!

tests/test_rm/integration/test_rm_int.py
andre-merzky: This being an *integration* test, shouldn't those values be actually read from the resource instead of being mocked?

tests/test_rm/integration/test_rm_int.py
andre-merzky: Well, one can read the resource config via `ru.read_json(filename)`.  The filename is easy to derive from the resource label, and so is the config section to be used.  But we don't have a separate method.

andre-merzky: Well, the session has a `_load_resource_configs(self)` method - but that expects an initialized session to exist...

andre-merzky:
@lee212 : do you have an estimated timeline for an iteration on this branch/PR?

andre-merzky:
@lee212 : Hey Hyungro, what is the status/timeline for this?

andre-merzky:
@lee212 , @mturilli : I would like to discuss the trajectory for this PR on the next devel call - is that ok with both of you?

andre-merzky:
Postponed to march

andre-merzky:
This PR has not been touched for a while - @lee212: what is the trajectory?

andre-merzky:
This PR has not been touched for a while - @lee212: what is the trajectory?

andre-merzky:
@lee212 : ping: any plans on how to proceed?

andre-merzky:
@lee212 : ping: any plans on how to proceed?

andre-merzky:
@iparask : can you have a look at the branch and let me know if you see a way forward?

iparask:
I think we can use most of the work @lee212 has done here. It needs some changes to bring it to the new testing guidelines, but I like the approach.

I propose to merge this branch with the one I am working on and close this PR. Then we can delete this branch as it will not be needed.

andre-merzky:
Thanks both!  @iparask , feel free to proceed as suggested.

PR #1922, opened 2019-07-18
Fix/expand rcfg

andre-merzky:
With this PR, the resource config can contain entries like:
```
"default_workdir" = "$SCRATCH/%(pd.project)s/sandboxes/"
```
which are expanded from the remote environment and from pilot description keys.  This should address the  shared FS selection problem where the correct path depends on the used allocation.


andre-merzky:
closed in favor of #1958

PR #1923, opened 2019-07-18
Feature/test schedulers

andre-merzky:
@iparask : is this ready for merge?


iparask:
I would say no. As you can see it does not provide correct test coverage. I'll reopen this PR when it is ready.

andre-merzky:
Ack. You may want to make sure though that this becomes mergeable before the Py3 transition.

PR #1925, opened 2019-07-30
Hotfix/tiger cfg

andre-merzky:
This hotfix sets the correct scheduler on tiger (cont), and fixes the VE location (on scratch).  Both are needed to get tiger working w/o additional user configuration.


src/radical/pilot/configs/resource_princeton.json
```diff
@@ -64,21 +64,21 @@
...
-        "pre_bootstrap_0"             : ["module purge", 
+        "pre_bootstrap_0"             : ["module purge",
                                          "module load intel/18.0/64/18.0.3.222",
                                          "module load intel-mpi/intel/2018.3/64",
                                          "module load intel-python/2.7"
                                         ],
-        "default_remote_workdir"      : "$HOME",
+        "default_remote_workdir"      : "/scratch/gpfs/$USER/",
+        "virtenv"                     : "/scratch/gpfs/$USER/ve.rp",
+        "virtenv_mode"                : "use",
```
lee212: Out of curiosity, what's different between 'create' and 'use'? Does 'create' create a new virtualenv if it doesn't exist whereas 'use' uses existing virtualenv?

mturilli:
Note that linting fails. I know it is a hotfix for a user so I approve it but we might want to try to keep linting happy in EnTK as it should be less demanding than for RP (maybe? :))

src/radical/pilot/configs/resource_princeton.json
andre-merzky: Good question :-)  `create` will create the ve if it does not exist - any subsequent uses of `create` will reuse the thus created ve.  `use` expects the ve to already exist, and will never attempt to create it.  We use this on resources where the automated ve creation fails.  In those cases, the user has to create the ve manually in advance.  On tiger, we don't have virtualenv installed, but also can't fetch virtualenv from the bootstrapper as the compute node have no direct internet link - thus the manual creation and `use`...

andre-merzky:
> Note that linting fails. I know it is a hotfix for a user so I approve it but we might want to try to keep linting happy in EnTK as it should be less demanding than for RP (maybe? :))

Ah, good catch - the PR should have been toward `master`, not `devel` - thanks!  Then the linter will only check the changed json file, and should not fail.  I'll reroute and let the tests run again...

andre-merzky:
linters are happy now :-)

PR #1926, opened 2019-07-31
Hotfix/conda ve

andre-merzky:
Unfortunately, tiger needs another hotfix: our users use Anaconda, but that does not play well with our static ve creations script: it does not like the way we install virtualenv.  Alas, the fix is a little more involved than calling `conda install -y virtualenv`, since (a) the conda command was recently moved to a shell function which is not available in scripts (this was a really stupid decision IMHO), and (b) different versions of conda use different env variables to point to the conda executable (*sigh*).  Either way: this patch makes sure that virtualenv is installed in the used conda env, and then uses that to create our static ve.

### DIFF
--- a/bin/radical-pilot-create-static-ve
+++ b/bin/radical-pilot-create-static-ve
@@ -4,7 +4,7 @@ script="$0"
 prefix="$1"
 arg="$2"
 
-echo 
+echo
 echo "script : $script"
 echo "prefix : $prefix"
 echo "arg    : $arg"
@@ -26,7 +26,7 @@ help(){
 
     err="$1"
     ret=0
-    
+
     if ! test -z "$err"
     then
         ret=1
@@ -62,7 +62,7 @@ progress(){
 
 if test "$prefix" = "-h"
 then
-    help 
+    help
 fi
 
 if test -z "$prefix"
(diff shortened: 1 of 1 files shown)


bin/radical-pilot-create-static-ve
```diff
@@ -111,11 +111,27 @@ fi
...
+    CONDA="$(which conda 2>/dev/null | head -n 1)"
+    test -z "$CONDA" && CONDA="$_CONDA_EXE"
+    test -z "$CONDA" && CONDA="$CONDA_EXE"
+    if test -z "$CONDA"
+    then
+        echo "conda is not functional"
+        exit 1
+    fi
+    echo -n "install conda virtualenv "
+    "$CONDA" install -y virtualenv 2>&1 | progress || exit 1
```
iparask: This is where the virtualenv is created, right? It should be `conda create` instead of `install`

andre-merzky: Ah, no - this is where `virtualenv` is installed in the conda env, so that the remainder of the script can then use `virtualenv` to create the pilot VE.  Yeah, one can also use conda to actually *create* the ve - but I wanted to keep the changes minimal, and not start testing how the pilot likes a such created VE.  With the current fix, the VE created by the script remains exactly as before.

iparask: okay!

PR #1932, opened 2019-08-07
Feature/coschedule

andre-merzky:
This is a first attempt at a co-locating scheduler.  It expects unit descriptions to have tags like

```
cud.tags = {'colocate': {'bag' : 'bag_1',
                         'size': 3}}
```

This will inform the scheduler that the workload features a bag of tasks, called `bag_1` which consists of three tasks.  The scheduler will collect those three tasks and place them all on the same node (or not at all, if the bag cannot be placed).  Untagged units are scheduled individually, as before.


andre-merzky:
@mturilli : I'll merge this as experimental feature, lets see where it goes.

PR #1933, opened 2019-08-07
Hotfix/pilot term

andre-merzky:
We know 4 different protocols to request termination from a thread in abnormal conditions: exception, `interrupt_main`, signals and `sys.exit()`.  None of them seems to be very reliable.  We used to use an exception to inform the application about pilot errors - but that became unstable over the last weeks, in that the exceptions did not arrive in the main thread anymore.  This PR switches to `interrupt_main()`, which is fairly brutal, but works.  The main problem with it is that the application cannot catch it, really - but since it is only issued when the pilot sets `terminate_on_error` (default: `yes`), this seems acceptable.

### DIFF
--- a/src/radical/pilot/compute_pilot.py
+++ b/src/radical/pilot/compute_pilot.py
@@ -3,15 +3,12 @@ __copyright__ = "Copyright 2013-2016, http://radical.rutgers.edu"
 __license__   = "MIT"
 
 
-import os
-import sys
 import copy
 import time
 import threading
 
 import radical.utils as ru
 
-from . import utils     as rpu
 from . import states    as rps
 from . import constants as rpc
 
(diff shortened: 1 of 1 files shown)


andre-merzky:
Sorry Matteo, I again forgot to target `master` for this hotfix.  The original diff was:

```
-          # ru.cancel_main_thread('int')
-            raise RuntimeError('pilot %s failed - fatal!' % self.uid)
+            ru.cancel_main_thread('int')
+          # raise RuntimeError('pilot %s failed - fatal!' % self.uid)
```
but it's now bigger because I did some linting on that file.  There remain linter complains about `protected access`.  I can fix those, but that's quite some work and beyond what I would consider prudent for a hotfix: we would need to introduce some `protected` accessors (python has no native notion of protected access).

mturilli:
Thanks @andre-merzky , I think this is already an improvement. Should we merge?

andre-merzky:
ack - thanks, Matteo.

PR #1938, opened 2019-08-13
Increased default runtime to 30 minutes in getting_started.py

lee212:
Some resources i.e. stampede2 KNL requires extra time for package compilation e.g. zmq using Intel Compiler and the getting_started.py example may need increased default runtime for initial bootstrapping.


andre-merzky:
Thanks @lee212 !

quick note: we use the `hotfix/...` branches (which should usually target `master`)to trigger new releases immediately upon merge - otherwise it's just a `fix/...`.  I assume you did not want a new release for this PR, right?

lee212:
Will use `fix/...` in the next time. Thanks!

PR #1942, opened 2019-08-13
archiving orte executor

### LINKED ISSUE #1941: Archive stale orte.py and related code

andre-merzky:
This fixes #1941


src/radical/pilot/agent/executing/base.py
```diff
@@ -59,16 +62,13 @@ def create(cls, cfg, session):
         from .shell_fs import ShellFS
         from .abds     import ABDS
 
-        try:
-            from .orte     import ORTE
-        except:
-            pass
+      # from .orte     import ORTE
```
mturilli: Are we sure we want to leave around the commented code?

andre-merzky: It is the same scheme we used in other, similar parts of the code - see https://github.com/radical-cybertools/radical.pilot/blob/devel/src/radical/pilot/agent/lm/base.py#L120 for example.

andre-merzky:
I cleaned the archived code to remove some valid linter warnings - but the remaining ones are just due to the fact that this is archived code, and I would ignore them.

We probably should purged all archived files after 1.0 release.

Issue #1930, opened 2019-08-06
Add methods in scheduler base class


iparask:
We discussed to add methods in the Scheduler base class with a `NotImplemented` runtime error.

PR #1948, opened 2019-08-20
Adding not implemented methods in scheduler base class

iparask:
This PR adds two methods in the base class that need to be implemented in all derived classes. As a result the base class raises an error they are not implemented.


iparask:
It fixes issue #1930

andre-merzky:
Great - thanks!

PR #1950, opened 2019-08-21
saga needs help for tiger special case

andre-merzky:
we need to trigger a special case in SAGA, and that means we need the hostname in the job service URL.

This will be a hotfix release.


andre-merzky:
See also https://github.com/radical-cybertools/radical.saga/pull/734

PR #1951, opened 2019-08-23
move metric definitions from RA to RP

andre-merzky:
This moves some RP specific metric definitions from RA to RP.


src/radical/pilot/utils/prof_utils.py
```diff
@@ -89,20 +89,16 @@
...
                              {ru.EVENT: 'prte_sending_launch_msg'}],
             'prte_phase_2': [{ru.EVENT: 'prte_sending_launch_msg'},
-                             {ru.EVENT: 'prte_running'           }],
-            'exec_app'    : [{ru.EVENT: 'prte_running'           },
-                             {ru.EVENT: 'prte_iof_complete'      }],
-            'prte_phase_3': [{ru.EVENT: 'prte_iof_complete'      },
+                             {ru.EVENT: 'app_start'              }],
+            'exec_app'    : [{ru.EVENT: 'app_start'              },
+                             {ru.EVENT: 'app_stop'               }],
+            'prte_phase_3': [{ru.EVENT: 'app_stop'               },
```
mturilli: Seeing that in our group we tend to call the user-facing code 'application' and the executable of a CU 'kernel', should we name this `krnl` or `kern` instead of `app`?

andre-merzky: How about `task`?

mturilli: We went for cmd

mturilli:
I am happy to keep `radical-pilot-prte2prof` separated at the moment but we should ask Thomas whether the traces will make their way into the official release. If so, I would add this to our profiler so to avoid a 2-step profiling for prrte-based sessions. If we agree, should we open a ticket to track the adoption of the traces by PRRTE?

andre-merzky:
> I am happy to keep `radical-pilot-prte2prof` separated at the moment but we should ask Thomas whether the traces will make their way into the official release. If so, I would add this to our profiler so to avoid a 2-step profiling for prrte-based sessions. If we agree, should we open a ticket to track the adoption of the traces by PRRTE?

Agree on all that.  Only that I would move the trace parsing into the PRTE launch method and do that on the fly when we capture the traces, so that we keep prte related code close together.

src/radical/pilot/utils/prof_utils.py
andre-merzky: True - but at the moment our traces have `app_*` events.  But I don't mind changing that right now - the experiments are done, and we are the only ones using the `app/cmd` traces right now.

examples/00_getting_started.py
```diff
@@ -58,10 +58,7 @@
         pd_init = {'resource'      : resource,
                    'runtime'       : 30,  # pilot runtime (min)
                    'exit_on_error' : True,
-                   'project'       : config[resource]['project'],
-                   'queue'         : config[resource]['queue'],
-                   'access_schema' : config[resource]['schema'],
-                   'cores'         : config[resource]['cores'],
+                   'cores'         : 4 * 41 * 14,
```
mturilli: Is this intended? I don't think so as we used the config file so to avoid resource-specific default configs?

examples/00_getting_started.py
```diff
@@ -88,9 +85,9 @@
             cud = rp.ComputeUnitDescription()
             cud.executable       = '/bin/date'
             cud.gpu_processes    = 0
-            cud.cpu_processes    = 1
-            cud.cpu_threads      = 1
-            cud.cpu_process_type = rp.POSIX
+            cud.cpu_processes    = 90
+            cud.cpu_threads      = 4
+            cud.cpu_process_type = rp.MPI
```
mturilli: This also seem unintended

examples/lm_task.sh
```diff
@@ -0,0 +1,85 @@
+#!/bin/sh
+# test the correct startup of mixed OpenMP / MPI / CPU / GPU tasks
```
mturilli: Should this be in examples if it is a test?

src/radical/pilot/agent/lm/base.py
```diff
@@ -51,9 +53,9 @@ class LaunchMethod(object):
     #        cud.pre_exec, and then should also export those.  That would make
     #        our launch script ore complicated though...
     EXPORT_ENV_VARIABLES = [
-        'LD_LIBRARY_PATH',
-        'PATH',
-        'PYTHONPATH',
+      # 'LD_LIBRARY_PATH',
```
mturilli: Might this be related to the issues we discussed about missing envs on the compute nodes?

src/radical/pilot/agent/lm/prte.py
```diff
@@ -56,22 +56,34 @@ def lrms_config_hook(cls, name, cfg, lrms, logger, profiler):
...
-                fout.write('%s slots=%d\n' % (node[0], lrms.cores_per_node * lrms.smt))
+                fout.write('%s slots=%d\n' % (node[0],
+                                              lrms.cores_per_node * lrms.smt))
 
         pre   = os.environ['PRRTE_PREFIX']
         prte += ' --prefix %s'     % pre
         prte += ' --report-uri %s' % furi
         prte += ' --hostfile %s'   % fhosts
-      # prte += ' --mca plm_rsh_no_tree_spawn 1'
+        prte += ' --pmca orte_state_base_verbose 1'  # prte profiling
```
mturilli: Do we want this 'always on'? Maybe enable it only when `RADICAL_PROFILE="True"`?

src/radical/pilot/agent/lm/prte.py
```diff
@@ -276,16 +289,16 @@ def construct_command(self, cu, launch_script_hop):
             map_flag += ' -host %s' % hosts.rstrip(',')
 
         # Additional (debug) arguments to prun
-        debug_string = ''
+        debug_string = '-verbose'
```
mturilli: See comment in #1946

src/radical/pilot/agent/scheduler/base.py
```diff
@@ -423,6 +427,8 @@ def slot_status(self):
         Returns a multi-line string corresponding to the status of the node list
         '''
 
+        return '-'
```
mturilli: Too quick and dirty for release?

src/radical/pilot/agent/scheduler/base.py
```diff
@@ -471,16 +477,21 @@ def _schedule_units(self, units):
...
-      # # sort units by size
-      # def _sort(a,b):
-      #     da = a['description']
-      #     db = b['description']
-      #     va = da['cpu_processes'] * da['cpu_threads'] + da['gpu_processes']
-      #     vb = db['cpu_processes'] * db['cpu_threads'] + db['gpu_processes']
-      #     return cmp(vb, va)
-      # units.sort(_sort)
-
+        # make sure we have a unit size estimate
```
mturilli: We need tests for this method(s)

src/radical/pilot/agent/scheduler/base.py
```diff
@@ -684,6 +699,65 @@ def unschedule_cb(self, topic, msg):
         return True
 
 
+    # --------------------------------------------------------------------------
+    #
+    def set_tuple_size(self, unit):
+        '''
+        if we get resources freed, we dig through the list of waiting tasks to
+        see which we can schedule.  When failing to schedule a task, it makes in
+        general no sense to keep trying to scheduler tasks of the same size or
```
mturilli: trying to schedule

examples/00_getting_started.py
andre-merzky: Thanks, reverted.

examples/00_getting_started.py
andre-merzky: yep

examples/lm_task.sh
andre-merzky: I changed the text :-)

src/radical/pilot/agent/lm/base.py
andre-merzky: Yes - but that wasn't on this branch... :-/  I'll revert this anyway.

src/radical/pilot/agent/scheduler/base.py
andre-merzky: yes -  this only made sense on the large scale experiments we did.

src/radical/pilot/agent/scheduler/base.py
andre-merzky: I'll revert the scheduler changes - those will be properly applied when merging the `feature/sched_nolock` branch.

PR #1961, opened 2019-09-13
Feature/sched nolock

andre-merzky:
This changes the scheduler in 3 ways:
  - remove need for thread locks (event-driven main loop)
  - optimize iteration over node list (custom iterator with position persistency)
  - fix GPU placement

The latter is achieved by changing the perspective on tasks: instead of attempting to find all resources required by a task, the task is instead sliced into processes with individual (currently homogeneous) resource requirements, 'slots'), and the scheduler allocates those slots individually.  This simplifies the code significantly and fixes problems with GPU, LSF and MEM scheduling.  It also paves the way for non-uniform task layouts (but does not implement that).

There is another optimization pending which depends on https://github.com/radical-cybertools/radical.utils/pull/179 which improves performance over large pilots and large workloads significantly, but that optimization is orthogonal and does not (significantly) affect the changes in this PR.


andre-merzky:
PS.: the PR also includes a `SLEEP` executor which can only, as the name suggest, execute `sleep` commands - but it does so by tracking tasks and their sleep time in memory, and thus can run very many tasks with unlimited  concurrency and very limited overhead, which makes is nice for development, testing and debugging.

src/radical/pilot/agent/scheduler/base.py
```diff
@@ -477,133 +476,316 @@ def _schedule_units(self, units):
...
-                gpu_maps.append(node['gpu_map'])
-
-        if not gpu_maps:
-            # no gpu maps, nothing to do
+        # sort by inverse tuple size to place larger tasks first and backfill
+        # with smaller tasks.  We only look at cores right now - this needs
+        # fixing for GPU dominated loads.
+
+        self._waitpool.sort(key=lambda x: x['tuple_size'][0], reverse=True)
+        scheduled, unscheduled = ru.lazy_bisect(self._waitpool,
```
lee212: `_waitpool` starts with blank, does lazy_bisect() work okay with the empty list?

src/radical/pilot/agent/scheduler/base.py
```diff
@@ -477,133 +476,316 @@ def _schedule_units(self, units):
...
+            # if we have new resources, try to place waiting tasks
+            if resources:
+                a, r = self._schedule_waitlist()
+                if not r: resources  = False
+                if     a: active    += 'w'
 
-        # got an allocation, we can go off and launch the process
-        self._prof.prof('schedule_ok', uid=unit['uid'])
+            # always try to schedule newly incoming tasks
+            a, r = self._schedule_incoming()
```
lee212: Is this required when `resources is False`?

andre-merzky: yes - new tasks might be smaller than previous ones.  So if previous tasks could not be placed and we did not get new resources, we might still be able to place smaller incoming ones.
Thanks for checking!

src/radical/pilot/agent/scheduler/base.py
andre-merzky: Nice catch, I'll add a test in RU :-)

PR #1967, opened 2019-10-01
Rhea migration to Slurm since September 2019

lee212:
Rhea migrated batch schedulers from Moab to Slurm on September 03, 2019, according to [this](https://www.olcf.ornl.gov/for-users/system-user-guides/rhea/running-jobs/#slurm).


andre-merzky:
Thanks!

PR #1974, opened 2019-10-08
Hotfix/pub startup

andre-merzky:
Our code is getting too fast :-)

ZMQ pub channels have a small setup overhead.  While the TCP connection is set up (async!), no subscription can exist (there is not client who could subscribe), and thus early messages on that channel can get filtered out.  Proper synchronization mechanisms are somewhat involved, and we leave that for later protocol redesigns - for now, we introduce a reasonable timeout to resolve this.

This PR also switches back to multipart messages (easier to debug) and adds locks around send/receive of those multipart messages (to ensure atomicity).

### DIFF
--- a/src/radical/pilot/pmgr/launching/default.py
+++ b/src/radical/pilot/pmgr/launching/default.py
@@ -631,12 +631,15 @@ class Default(PMGRLaunchingComponent):
         cmd = "cd %s && tar zchf %s *" % (tmp_dir, tar_tgt)
         self._log.debug('cmd: %s', cmd)
         try:
-            out = sp.check_output(["/bin/sh", "-c", cmd], stderr=sp.STDOUT)
+            out, err, ret = ru.sh_callout(cmd, shell=True)
         except Exception:
-            self._log.exception('callout failed: %s', out)
+            self._log.exception('callout failed')
             raise
-        else:
+
+        if ret:
             self._log.debug('out: %s', out)
+            self._log.debug('err: %s', err)
+            raise RuntimeError('callout failed')
 
         # remove all files marked for removal-after-pack
         for ft in ft_list:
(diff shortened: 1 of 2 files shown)


andre-merzky:
Hey @mturilli @AymenFJA @lee212 @iparask : This is a somewhat urgent PR, does one of you find the time to review it quickly?  Change the diff view to ignore white space to simplify it a bit.  The important change is here: https://github.com/radical-cybertools/radical.pilot/pull/1974/files#diff-20666565344e7fd55f739c499b258961R322 which introduces a delay until sending first message on pubub channels.  It also adds locks around handling of multipart ZMQ messages, and replaces a shell callout which raised in incorrect error on failure.

Thanks.

mturilli:
Note that pylint, pep8 and coverage tests fails.

AymenFJA:
Doing the reviewing now. Thanks @andre-merzky

AymenFJA:
So can I ask why specifically switching back to Multipart messages in ZMQ? is it only because easier to debug? Or do we need some kind of message order preservation that the multipart message offers and other benefits?  Other than this the PR seems fine to me.  

Thanks.

andre-merzky:
> So can I ask why specifically switching back to Multipart messages in ZMQ? is it only because easier to debug? Or do we need some kind of message order preservation that the multipart message offers and other benefits? Other than this the PR seems fine to me.

Very good question - Matteo asked me the same :-)  As far as I have seen, multipart-messages and 'normal' single-part messages are equivalent in performance.  I wasn't sure about that in the beginning, that's why I added both code paths and made it easy to switch.  But yes, the switch back to MP is really just because the debug prints are a bit nicer.  I probably should remove the other code path by now...

Thanks! :-)

PR #1984, opened 2019-11-06
Feature/proc mgmt

andre-merzky:
This PR switches from a watcher and signal based process management to a heartbeat based one, to address several inherent reliability issues with the former approach.  It relies on radical-cybertools/radical.utils/pull/191 for the mechanics.  The most important changes are:

  - Components, Bridges and Agents are now all daemonized processes
  - process management is heartbeat based
  - agent lifetime management is heartbeat based 
  - clearer separation between primary and secondary session  [1]
  - clearer separation of configuration for the individual daemons

Several smaller changes

  - remove last traces of ZMQ channels (moved to RU)
  - improve consistency in variable and ID naming
  - consistent spelling on config options
  - coding guidelines in touched files
 
Review should probably focuse on (in this order)

  - `bin/radical-pilot-{component,bridge,agent}`
  - `src/radical/pilot/session.py`
  - `src/radical/pilot/utils/component.py`
  - `src/radical/pilot/agent/agent*.py`
  - component base classes

I'd be happy to interactively support the review, let me know!

[1]  A primary session connects to MongoDB and spawns bridges / components, a secondary session informs about primary sessions.  This will change further with the removal of MongoDB, and we can reconsider a clearer separation at that point.


andre-merzky:
@mturilli , @iparask : any estimate on when you'll find time for a review?  Can I help?  As said above, I don't mind doing that interactively...

mturilli:
@andre-merzky thank you for your patience. Now that I have done the bulk on the supporting PR in utils, I am moving to this one. Apologies for the time it is taking, they are massive and I am trying my best to give meaningful feedback even if necessarily not as detailed as I would have liked.

bin/radical-pilot-bridge
```diff
@@ -0,0 +1,159 @@
...
+      - name: name of the bridge
+      - kind: type of bridge (`pubsub` or `queue`)
+
+    If the config contains a `heartbeat` section, that section must be formatted
+    as follows:
+
+        {
+          'from'    : 'uid',
+          'pub'     : 'addr_pub',
+          'sub'     : 'addr_sub',
```
mturilli: We use `pub`/`sub` for the heartbeat. Does this mean that this section is only for the `pubsub` type of bridge?

bin/radical-pilot-component
```diff
@@ -0,0 +1,132 @@
...
+
+    # start a non-primary session
+    session = rp.Session(cfg=cfg, _primary=False)
+
+    # create the component and begin to work
+    comp  = rp.utils.Component.create(cfg, session)
+    comp.start()
+
+    # component runs - send heartbeats so that cmgr knows about it
+    hb_pub = ru.zmq.Publisher ('heartbeat', cfg.heartbeat.addr_pub)  #, log=log)
```
mturilli: We wrap main only when demonizing right? Maybe adding it in the docstring of the function?

src/radical/pilot/agent/lm/base.py
```diff
@@ -169,11 +169,13 @@ def create(cls, name, cfg, session):
 
         except KeyError:
             # pylint: disable=protected-access
-            session._log.exception("LM '%s' unknown or defunct" % name)
+            # FIXME
```
mturilli: Should we do this before merging?

src/radical/pilot/agent/lm/base.py
```diff
@@ -169,11 +169,13 @@ def create(cls, name, cfg, session):
...
+            pass
 
         except Exception as e:
             # pylint: disable=protected-access
-            session._log.exception("LM cannot be used: %s!" % e)
+            # FIXME
```
mturilli: Should we do this before merging?

src/radical/pilot/pilot_manager.py
```diff
@@ -77,33 +77,47 @@ def __init__(self, session):
...
+        #
+        name = None
+        if isinstance(cfg, str):
+            name = cfg
+            cfg  = None
+
+        cfg           = ru.Config('radical.pilot.pmgr', name=name, cfg=cfg)
+        cfg.uid       = self._uid
+        cfg.owner     = self._uid
+        cfg.sid       = self._session.uid
```
mturilli: For consistency, should this be session.sid?

src/radical/pilot/umgr/scheduler/backfilling.py
```diff
@@ -110,15 +111,15 @@ def update_pilots(self, pids):
 
                 state = self._pilots[pid]['state']
 
-              # self._log.debug('=== update pilot: %s %s', pid, state)
+              # self._log.debug('update pilot: %s %s', pid, state)
 
                 if  rps._pilot_state_value(state) < _BF_START_VAL:
-                  # self._log.debug('=== early')
+                  # self._log.debug('early')
```
mturilli: Maybe a bit more descriptive debug message?

andre-merzky:
> @andre-merzky thank you for your patience. Now that I have done the bulk on the supporting PR in utils, I am moving to this one. Apologies for the time it is taking, they are massive and I am trying my best to give meaningful feedback even if necessarily not as detailed as I would have liked.

No worries - I know they are large *sigh*

bin/radical-pilot-agent
```diff
@@ -30,43 +29,56 @@ def bootstrap_3(agent_name):
...
     agent = None
 
     try:
-        setproctitle.setproctitle('rp.%s' % agent_name)
+        setproctitle.setproctitle('rp.%s' % aid)
+
+        cfg = ru.Config(path='%s.cfg' % aid)
+        cfg.uid  = aid
+        cfg.aid  = aid
+        cfg.path = os.getcwd()
```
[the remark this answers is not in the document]
andre-merzky: Everything has a `uid`: components, channels, bridges, agents, managers, tasks, sub-agents, yadayada.  That might not always strictly be necessary (like, there is no need to distinguish individual comm channels), but it simplifies profiling, debugging, logging and heartbeats.

Sometimes several UIDs are used in the same scope: like, a config file might contain a agent uid and a session uid - in those cases we use the `pid` and `sid` alternatives to easily distinguish them in that scope.  In the snippet above, we pick an agent UID from such a config file which likely also contains a session and pilot UID, thus the `aid`.  The `cfg.aid = aid` should strictly speaking not be necessary as it should already be set, so I can remove that line.

Having said all this, I try to be careful to not let the alternatives name slip into the class attributes unless needed for clarification.  A pilot may have a `self.sid` to signify the session it is running under, but not a `self.pid` as its own pilot ID - that should remain `self.uid`.  The main reason for this is that any other method (like the profiler) can use `something.uid` reliably, and does not have to distinguish:
```
if isinstance(something, rp.Pilot) uid = something.pid
...
```
which is slow and repetitive.

I see your scattered comments about consistency, I'll look through those cases and clean up.

bin/radical-pilot-bridge
andre-merzky: Yes, heartbeats go over a pubsub channel.

bin/radical-pilot-bridge
```diff
@@ -0,0 +1,159 @@
...
+        > radical-pilot-sub    command foo &              [2]
+        > radical-pilot-pub    command foo bar            [3]
+
+    [1] establishes the pubsub channel 'command'
+    [2] connect to the command channel, subscribe for topic `foo`
+    [3] connect to the command channel, send messages for topic `foo`
+    '''
+
+    # basic setup: uid, logger, process name
+    uid  = cfg.uid
```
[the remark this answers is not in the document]
andre-merzky: There is only one uid in play here, that of the bridge, so `uid` should be fine.

bin/radical-pilot-component
andre-merzky: `main(cfg)` is running in the daemonized process.  I only introduced `wrapped_main` to handle the nesting (indentation).  In general, the `except` should not be too far away from the `try` to keep it coherent and readable, and I found the code which is now in `wrapped_main` too long for that.

That `main` is daemonized should be visible from 
```
ru.daemonize(main=main, args=[cfg], stdout='%s.out' % path,
                                        stderr='%s.err' % path)
```
and from the `ru.daemonize` documentation.  Let me know if you think that's insufficient!

examples/00_getting_started.py
```diff
@@ -67,6 +65,8 @@
         # Launch the pilot.
         pilot = pmgr.submit_pilots(pdesc)
 
+        time.sleep(100)
+
```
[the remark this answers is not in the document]
andre-merzky: Uh, thanks - that needs to go away.  I used it to keep the process alive to attach a debugger...

src/radical/pilot/pilot_manager.py
```diff
@@ -77,33 +77,47 @@ def __init__(self, session):
...
+        #       a predefined config and amed it at the same time.  This might
+        #       be ok for the session, but introduces a minor API inconsistency.
+        #
+        name = None
+        if isinstance(cfg, str):
+            name = cfg
+            cfg  = None
+
+        cfg           = ru.Config('radical.pilot.pmgr', name=name, cfg=cfg)
+        cfg.uid       = self._uid
```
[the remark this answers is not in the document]
andre-merzky: Not needed I think - it's the pmgr's config, so `uid` is contextualized well.

src/radical/pilot/pilot_manager.py
andre-merzky: See above, this should be ok.

src/radical/pilot/umgr/scheduler/backfilling.py
andre-merzky: I'd rather remove them altogether I think - they are only needed for localized debugging...  Thanks!

src/radical/pilot/umgr/scheduler/backfilling.py
```diff
@@ -110,15 +111,15 @@ def update_pilots(self, pids):
...
                     # not eligible, yet
                     continue
 
                 if  rps._pilot_state_value(state) > _BF_STOP_VAL:
-                  # self._log.debug('=== late')
+                  # self._log.debug('late')
```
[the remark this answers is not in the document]
andre-merzky: same

src/radical/pilot/unit_manager.py
```diff
@@ -81,46 +79,48 @@ def __init__(self, session, scheduler=None):
...
         # initialize the base class (with no intent to fork)
-        self._uid    = ru.generate_id('umgr.%(item_counter)04d', ru.ID_CUSTOM,
-                                      namespace=session.uid)
-        cfg['owner'] = self.uid
-        rpu.Component.__init__(self, cfg, session)
-        self.start(spawn=False)
-        self._log.info('started umgr %s', self._uid)
+        self._uid  = ru.generate_id('umgr.%(item_counter)04d', ru.ID_CUSTOM,
+                                    ns=session.uid)
+        cfg.uid       = self.uid
```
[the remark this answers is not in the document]
andre-merzky: and previous reply :-D

andre-merzky:
> No worries - I know they are large _sigh_

BTW: It helps if you configure the github diff view to ignore white space changes.  Also, it is easier to parse the diffs in a side-by-side view: instead of an mix of new/old lines, you get a consistent view of the old code on the left and of the new code on the right side, which makes the code itself much easier to read on major changes like this.

Our policy to cleanup all touch files wrt. coding guide lines is a good one, but it also increases the size of the diffs significantly when many files are touched - and this PR by design touches almost everything in RP :-/  So many changes are really about consistency and comment style, and obfuscate the semantic part - apologies. On the up-side: that process should be mostly complete, and later diffs should be much smaller...

src/radical/pilot/agent/agent_0.py
```diff
@@ -9,140 +9,139 @@
...
 
 
-# this needs git attribute 'ident' set for this file
-git_ident = '$Id$'
-
-
-# ==============================================================================
+# ------------------------------------------------------------------------------
 #
 class Agent_0(rpu.Worker):
```
mturilli: Should this be a ` ``` ` docstring?

src/radical/pilot/agent/agent_0.py
```diff
@@ -180,110 +210,68 @@ def initialize_parent(self):
...
+        self._log.info('pilot state: %s [%s]', state, self._final_cause)
 
-        out = None
-        err = None
-        log = None
+        out, out, err = None, None, None
 
-        try   : out = open('./agent_0.out', 'r').read(1024)
+        try   : out = open('./agent.0.out', 'r').read(1024)
         except: pass
```
mturilli: Should we raise/log a warning?

src/radical/pilot/session.py
```diff
@@ -24,36 +12,40 @@
...
+    it holds :class:`radical.pilot.PilotManager` and
     :class:`radical.pilot.UnitManager` instances which in turn hold
     :class:`radical.pilot.ComputePilot` and :class:`radical.pilot.ComputeUnit`
-    instances.
-    """
+    instances, and several other components which operate on those stateful
+    entities.
+    '''
+
+    # In that role, the session will create a special pubsub channel `heartbeat`
```
mturilli: I would add the following to the previous docstring so that it goes into our documentation automatically.

src/radical/pilot/session.py
```diff
@@ -62,379 +54,174 @@ def __init__(self, dburl=None, uid=None, cfg=None, _connect=True):
...
+              for example in some of it's components.  A secondary session will
+              inherit the original session ID, but will not attempt to create
+              a new DB collection - if such a DB connection is needed, the
+              component needs to establish that on its own.
+        '''
 
-            # we need a dburl to connect to.
-            if not dburl:
-                dburl = os.environ.get("RADICAL_PILOT_DBURL")
+        # NOTE: `name` and `cfg` are overloaded, the user cannot point to
```
mturilli: I would add the following to the previous docstring so that it goes into our documentation automatically.

src/radical/pilot/session.py
```diff
@@ -800,13 +579,20 @@ def list_resources(self):
...
+                resources.append('%s.%s' % (domain, host))
+
+        return sorted(resources)
 
 
     # -------------------------------------------------------------------------
     #
     def add_resource_config(self, resource_config):
-        """Adds a new :class:`radical.pilot.ResourceConfig` to the PilotManager's
+        '''Adds a new :class:`radical.pilot.ResourceConfig` to the PilotManager's
```
mturilli: Along all the file we use ` '''Word ` and ` '''/n ` inconsistently.

src/radical/pilot/utils/component.py
```diff
@@ -4,301 +4,382 @@
...
-from .pubsub     import PUBSUB_BRIDGE  as rpu_PUBSUB_BRIDGE
-
 
 # ------------------------------------------------------------------------------
 #
-class Component(ru.Process):
-    """
-    This class provides the basic structure for any RP component which operates
-    on stateful things.  It provides means to:
+class ComponentManager(object):
```
mturilli: We probably need a docstring here

src/radical/pilot/utils/component.py
```diff
@@ -4,301 +4,382 @@
...
+
+        self._hb_pub = ru.zmq.Publisher('heartbeat',
+                                        self._cfg.heartbeat.addr_pub,
+                                        log=self._log)
+        self._hb_sub = ru.zmq.Subscriber('heartbeat',
+                                         self._cfg.heartbeat.addr_sub,
+                                         topic='heartbeat',
+                                         cb=self._hb_sub_cb, log=self._log)
+
+        # confirm the bridge being usable by listening to our own heartbeat
```
mturilli: What we do if this 'confirmation' fails?

src/radical/pilot/utils/component.py
```diff
@@ -4,301 +4,382 @@
...
-    process.  Components inheriting this class should be fully self sufficient,
-    and should specifically attempt not to use shared resources.  That will
-    ensure that multiple instances of the component can coexist for higher
-    overall system throughput.  Should access to shared resources be necessary,
-    it will require some locking mechanism across process boundaries.
+    # --------------------------------------------------------------------------
+    #
+    def _hb_sub_cb(self, topic, msg):
+        '''
+        keep treack of heartbeats for all bridges/components we know
```
mturilli: treack -> track

src/radical/pilot/utils/component.py
```diff
@@ -4,301 +4,382 @@
...
-        The symmetric teardown methods are called `finalize_child()` and
-        `finalize_parent()`, for child and parent process, repsectively.
+        The symmetric teardown methods are called `finalize()`.
 
         Constructors of inheriting components *may* call start() in their
         constructor.
-        """
+        '''
 
         # NOTE: a fork will not duplicate any threads of the parent process --
```
mturilli: I would add this note to the main docstring

src/radical/pilot/utils/component.py
```diff
@@ -333,88 +416,96 @@ def __init__(self, cfg, session):
...
 
-        # make sure we bootstrapped ok
-        self.is_valid()
-        self._session._to_stop.append(self)
+        sync = mt.Event()
+        self._thread = mt.Thread(target=self._worker_thread, args=[sync])
+        self._thread.daemon = True
+        self._thread.start()
 
+        # TODO: add timeout to shield against failing initlization?
```
mturilli: This does not look 'clean' and I saw it in other places. Should we have a way to pub/sub component initialization? Maybe related to the TODO of line 465?

src/radical/pilot/utils/component.py
```diff
@@ -1298,17 +1033,15 @@ def advance(self, things, state=None, publish=True, push=False, ts=None,
...
+        '''
 
         if not ts:
             ts = time.time()
 
         if not isinstance(things, list):
             things = [things]
 
-        self._log.debug('advance bulk size: %s [%s, %s]', len(things), push, publish)
+        self._log.debug('advance bulk: %s [%s, %s]', len(things), push, publish)
```
mturilli: Do we need this or is it a left over from debugging?

bin/radical-pilot-bridge
mturilli: Can we say this explicitly in the docstring?

andre-merzky: I may have misunderstood you?  Not sure.  That config section:
```
{
          'from'    : 'uid',
          'pub'     : 'addr_pub',
          'sub'     : 'addr_sub',
...
```
is to connect the bridge to our heartbeat monitor.  The bridge itself can be for pubsub channels or queues or whatever.  The heartbeats do not run over a bridge (chicken & egg, you see :-)).

Did that answer make more sense?  I might still be off the mark...

src/radical/pilot/agent/agent_0.py
andre-merzky: Yeah, we can do that.  Will do.

src/radical/pilot/agent/agent_0.py
andre-merzky: No, that is actually not an error - the agent may well have not printed any output, we just check.  Apparently python people like to do it like this, I personally would prefer:
```
if os.file.isfile(fname) and os.file.size(fname): out = open(fname, 'r').read(1024)
```
oh well...

src/radical/pilot/session.py
andre-merzky: Not sure - I would consider the heartbeat management to be an implementation detail, not something the end user would need to know as part of the API doc (it is nowhere exposed in the API).  Do you think differently?

src/radical/pilot/session.py
andre-merzky: I don't think so: the docstring for the `cfg` parameter already mentions the overloading (``` (`str` or `dict`): a named or instantiated configuration```) - I only pointed it out in an extra comment as that is one of the very view places in the API where we overload based on type.

src/radical/pilot/session.py
andre-merzky: Alas we do.  I converted many already (also from `"""` to `'''`, which was used inconsistently), but I am sure many remain...

src/radical/pilot/utils/component.py
andre-merzky: Well, it is not exposed to the end user, and I would not include it in the API documentation.  Having said that, it needs a docstring anyway for whoever reads that code - will do.

src/radical/pilot/utils/component.py
andre-merzky: if heartbeats fail we die...

src/radical/pilot/utils/component.py
andre-merzky: thanks, fixed.

src/radical/pilot/utils/component.py
andre-merzky: Not sure if that is something I would want an end user to be exposed to - it is an implementation level information, isn't it?  We may have a different audience in mind when thinking about docstrings?  I consider what ends up on `readthedocs.net` to be mostly consumed by users, but don't expect it to be sufficient for RCT  developers by any means.  Do we differ here?

src/radical/pilot/utils/component.py
andre-merzky: Ah, thanks, I should remove this comment - that shieding is now in place via this `sync` event you see above, which, well, syncs startup and initialization.

andre-merzky: I am not sure if this is really needed, the heartbeat startup timeout should kick in anyway.  I left it as an open question to revisit eventually in case we run into troubles with faulty init routines.  If you think this needs resolving let me know, and I'll dig through it more thoroughly!

src/radical/pilot/utils/component.py
andre-merzky: Not sure I understand: it is a debug log entry, so it is here for debugging.  It is not semantically meaningful, so we don't *need* this (apart from debugging).  What am I missing?

src/radical/pilot/session.py
mturilli: This would end up in the automatically generated documentation that is intended for developers and necessarily (only) for end users? I see the distinction between API and impl. detail but I also see the need to expose very valuable documentation about our implementation. Maybe something we can get smarter with Sphinx?

src/radical/pilot/utils/component.py
mturilli: Following from my previous comment, I think we agree on the distinction between API and impl. details. I think expanding our view of 'user' including developers would be useful. We have a trove of developer-oriented documentation in the code and I think it would be useful to structure it as part of our documentation.

src/radical/pilot/utils/component.py
mturilli: I noticed that in previous cases you preferred to remove debug statements instead of leaving them around in the code. Maybe this is different? It is logged only when logging is set to DEBUG by the user?

src/radical/pilot/utils/component.py
mturilli: The PR is the first of its kind (bye bye darn mt/mp wrapper, yay!) so let's keep it like that and see how it works. I would maybe add something to the TODO about using different methods than a wait to manage state transitions (in this case transition from initialize to initialized)?

src/radical/pilot/utils/component.py
andre-merzky: Ah, I see.  I sometimes add additional localized debug messages for a specific debugging session and need to clean up afterwards.  Those need to be cleaned out because they don't provide sufficient context to be useful outside of that specific session, add too much noise, or are high-frequency and impact performance (the latter usually I just comment out so they remain somewhat available).  

The one above is a low frequency log which gives reasonable information about task flux for that component, so it can well stay.

Thanks for checking and context! :-)

src/radical/pilot/utils/component.py
andre-merzky: Ack, my thoughts as well.  I'll expand that comment.

andre-merzky: Ah well, I just fixed it - was shorter than adding the comment :-D

bin/radical-pilot-bridge
mturilli: Thanks, this helps. Suggested small change in documentation: ```RCT heartbeats do not run over a bridge so we may want to connect a bridge to an heartbeat monitor. If the config contains a `heartbeat` section, that section must be formatted as follows:``` Would this work? The origin of this comment was that without knowing that implementation detail, I was confused about the relationship between heartbeat and bridge.

andre-merzky: Great, I fixed the doc.

andre-merzky:
TODO AM: expand docstrings

andre-merzky:
unit tests: `start_bridges()` / `start_components()`, check if child is alive.  Check different cfg files.

`components_stop()` / `bridges_stop()` ?  At least document how that should be implemented.  consider testing.

andre-merzky:
> Are exceptions gone for good (I noticed that exceptions.py is gone)?

Yes, they are gone since quite a while, really, exceptions.py wasn't really used anymore, so I cleaned it out...

Thanks!

andre-merzky:
Tests are passing locally now - but pep8 a nd pylint will take a while :-(

PR #1987, opened 2019-11-07
Fix/py3 summit

andre-merzky:
This gets RP to work on summit.  A problem remains: on termination, the logger module will complain as it tries to open log files while some modules are already unloaded.  I would like to wait with investigating this, it might be resolved in the new process mgmt (did not yet test).


examples/config.json
```diff
@@ -149,7 +149,7 @@
     },
 
     "ornl.summit_prte" : {
-        "project"  : "CSC343",
+        "project"  : "geo111",
```
mturilli: I think we should also depart from hard-coding project IDs into our example configs. Should we just pass it with a env and an exception if it is not defined?

andre-merzky: Ha, this surprises me! *you* suggest an env variable (I never though I'd see the day), and *me* saying `nah, not really` - whats wrong?  Is the world coming to an end?? :-)

In all sincerity though, I think env vars are great for settings which vary from run to run, and config files are great for settings which vary from user to user or resource to resource.

Having said that, our `examples/config.json` is clearly not a satisfactory solution.  Apart from the fact that I should not have committed the above, its obvious that it needs to be changed for every user on every tutorial.

I would envision a better solution to have something like a `~/.radical/pilot/examples.cfg` with very much the same settings which we have in the current example config, so that those data are specific to users and tutorials, and only fall back to `examples/examples.cfg` if the above does not exist.  We can then document that behavior in `examples/examples.cfg` and prepare it as a template for the user to place into `~/.radical` as part of the tutorial setup.  Would something like this makes sense IYO, or is it overkill / off target?  I wouldn't mind having a go at it...

mturilli: I think about examples as something a new user wants to try out as sort of RP hello world. I would use the simplest, most self evident possible approach. To me, having to know about, and write or edit my own `~/.radical/pilot/examples.cf` does not seem to fit that requirement. Maybe I am wrong in my approach to the whole examples thing though.

andre-merzky: We can check what other systems are doing.  If I'm not mistaken, many do require changing code or config files before running any example.  FWIW, so does EnTK :-)

A really self sufficient option would be to get rid of the config altogether, and have a small section at the beginning of each example, like

```python
#!/usr/bin/env python

# configuration for target resource
RP_RESOURCE = 'local.localhost'
RP_SCHEMA = 'fork'
RP_QUEUE = None
...
```

That is just a few lines, and easy to change and to reference in the documentation.  We can still document our config system in one of the advanced examples (if we want).

Not saying that this is the best way, but it is a rather simple and self-sufficient option.

andre-merzky: I moved this discussion to #1989

lee212:
`default_remote_workdir` with a project ID changes; for example, I have 3 projects and now I see `radical.pilot.sandbox` on each. `ve.ornl.summit.xx.x` can have multiple copies with different versions. This might become difficult to trace when a user has several projects. I don't think this is a problem but it would be good to have a unified view or a location for sandbox or at least for ve.

mturilli:
There is a trick here. Once the user has privileges on multiple projects, she can decide to use one sandbox location for each run, independent on what project she uses to run her application. E.g., let's assume I have project1 and project2. I can configure RP sandbox path to `/some/path/project1` and then run my application using my project2 allocation by using `{..., project='project2',...}`.

andre-merzky:
We have two options for the default settint: either we pick the project from the pilot description (then we have different locations if different pilot descriptions are used), or we use a hardcoded one (which fails if the user is not a member of that project).   The former seems to be the smaller problem.

An alternative is to use a shared file system path which is (a) accessible under any project ID, and (b) independent of the used project ID.  Alas, summit does not have such a file system, the default permissions close the shared FS path to other projects, and we can't change those permissions on project level.

Ultimately, the user can provide it's own config file in `~/.radical/pilot/configs/resource_ornl.cfg`.  If you place a file there which contains just:

```json
{
    "ornl.summit_prte": {
        "default_pilot_sandbox" : "/tmp"
    }
}
```

then that setting will overwrite the one from the RP installation.  So, if project specific path are confusing, you can set a single path there.

andre-merzky:
BTW: should this discussion be moved to a ticket?

PR #1988, opened 2019-11-09
Fix/py3 frontera

andre-merzky:
Depends on #1987, and adds some minor fixes for Frontera


andre-merzky:
@lee212 : #1987 is merged now, so this one is ready for merge, too.  Hint: if you disable whitespace diff, the diff becomes very small :-)

src/radical/pilot/configs/resource_xsede.json
```diff
@@ -162,7 +162,7 @@
                                         "module unload intel",
                                         "module unload impi",
                                         "module load   intel",
-                                        "module load   python2"
+                                        "module load   python3/3.7.0"
```
lee212: Out of curiousity, 3.5.2 for summit but frontera might have 3.7.0 only?

andre-merzky: yes, exactly!  Summit also has 3.7.0, and it likely also works - but I happen to have tested with 3.5.2 ...

lee212: `module load python/3.5.2` on Summit seems abnormal, it has `python` symlink to python2 whereas  3.7.0 has correctly linking `python` to its python3. This explains that official modules are not fully tested to work as expected. Yes, I agree it wouldn't be much different between 3.5.2 and 3.7.0.

andre-merzky: I don't see a symlink to `python`, but it indeed does not map `python` at all, only `python3`.  Yeah, this might confuse, also within our own stack.  I'll switch to `3.7.0` - thanks!

PR #1998, opened 2019-11-20
Bridges dependencies fixed

iparask:
This PR introduces two changes.

1. Sets conda env python from general 3 to 3.5
2. Loads Anaconda3 in the configuration

This configuration passes all the examples on Bridges. GPU queues to be used.

### DIFF
--- a/src/radical/pilot/agent/bootstrap_0.sh
+++ b/src/radical/pilot/agent/bootstrap_0.sh
@@ -993,7 +993,7 @@ virtenv_create()
     elif test "$python_dist" = "anaconda"
     then
         run_cmd "Create virtualenv" \
-                "conda create -y -p $virtenv python=3.5"
+                "conda create -y -p $virtenv python=3"
         if test $? -ne 0
         then
             echo "ERROR: Couldn't create virtualenv"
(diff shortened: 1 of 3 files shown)


src/radical/pilot/agent/bootstrap_0.sh
```diff
@@ -993,7 +993,7 @@ virtenv_create()
     elif test "$python_dist" = "anaconda"
     then
         run_cmd "Create virtualenv" \
-                "conda create -y -p $virtenv python=3"
+                "conda create -y -p $virtenv python=3.5"
         if test $? -ne 0
```
andre-merzky: Uh, this is tricky - machines might well have only 3.7 installed...

iparask:
If you do `conda create .. python=3` it will install the latest python3 subversion, regardless the version of the case python. The latest on conda-forge is `3.8`. On the other hand, if we setup the first subversion, it will install that. Let me run the tests again without the subversion and see what happens.

andre-merzky:
> If you do `conda create .. python=3` it will install the latest python3 subversion, regardless the version of the case python. The latest on conda-forge is `3.8`. On the other hand, if we setup the first subversion, it will install that. Let me run the tests again without the subversion and see what happens.

Right, that's what I would expect, too.  If that works on bridges I'd much prefer it, there is a decent chance then that it also works on other machines.  Thank for looking into it!

Issue #1990, opened 2019-11-11
00_getting_started.py; Error processing the ERF, missing GPU index


lee212:
STDERR:
```
Error: Line 2:36:(RS GPU): syntax error, unexpected CLOSE_BRACKET, expecting NUM
BER or RANGE or WILDCARD
Error processing the ERF /gpfs/alpine/scratch/hrlee/csc393/radical.pilot.sandbox
/rp.session.login1.hrlee.018211.0000/pilot.0000/unit.000000//unit.000000.rs
```

unit.000000.rs:
```
cpu_index_using: physical
rank: 0: { host: 1; cpu: {0}; gpu: {}}
```

radical-stack:
```
  python               : 3.7.0
  pythonpath           : /sw/summit/xalt/1.1.4/site:/sw/summit/xalt/1.1.4/libexec
  virtualenv           : /autofs/nccs-svm1_home1/hrlee/venv3/entk

  radical.entk         : 0.90.0-v0.72.0-50-g657fbd3@python3-devel
  radical.pilot        : 0.90.1-v0.90.1-50-g7ab71cf@devel
  radical.saga         : 0.90.0-bv0.72.0-44-g67a0318@devel
  radical.utils        : 0.90.3-v0.90.3-94-gcaf4cba@devel
```

andre-merzky:
This is surprising.  Did the jsrun spec change, do you happen to know?

Can you please attach the pilot sandbox and the unit description of the offending unit?

lee212:
[rp.session.login4.hrlee.018214.0003.tar.gz](https://github.com/radical-cybertools/radical.pilot/files/3848467/rp.session.login4.hrlee.018214.0003.tar.gz)

This is running `00_getting_started.py`, so the unit description is:

```
            cud = rp.ComputeUnitDescription()
            cud.executable    = '/bin/date'
            cud.cpu_processes = 1
```

lee212:
Isn't this related? to: https://github.com/radical-cybertools/radical.pilot/blob/88e1761fc124b708d1542981fdaf99dba2f25b33/src/radical/pilot/agent/lm/jsrun.py#L31

andre-merzky:
It is - but according to line 88, we should only trigger the creation of a gpu spec if `gpu_maps` was set by the scheduler.  That should not be the case for non-GPU tasks.

lee212:
PR #2021 fixes this

PR #2001, opened 2019-12-03
Fix for Incorrect gpu_maps #1990

lee212:
Fix for #1990


src/radical/pilot/agent/lm/jsrun.py
```diff
@@ -87,7 +87,7 @@ def _create_resource_set_file(self, slots, uid, sandbox):
                 rs_str += ' cpu: {%s}'  % cores
                 if gpu_maps:
                     gpus  = ','.join(str(gpu) for gpu in gpu_maps.pop(0))
-                    rs_str += '; gpu: {%s}' % gpus
+                    rs_str += '; gpu: {%s}' % gpus if gpus != '' else ''
                 rs_str += '}\n'
```
andre-merzky: Did you find why the `gpu_maps` contained an empty list?

andre-merzky: @lee212 : ping?

lee212: I wonder what expected `gpu_maps` looks like. For example, `[[]]` was presented but `if gpu_maps:` condition couldn't detect this as an empty which I believe this is a valid case, isn't it?

lee212: I do not find why `gpu_map` is presented with an empty list.

andre-merzky:
Thanks for the fix, but I'll close this in favor of #2021 which addresses this on the scheduler level.  This way other LMs will benefit from the corrected `gpu_map`.

PR #2008, opened 2019-12-16
Optimization/pilot staging

andre-merzky:
optimize output staging.  This supports radical.repex.


iparask:
Can you provide me with some context? What is the feature that is supported?

src/radical/pilot/compute_pilot.py
```diff
@@ -582,14 +580,18 @@ def stage_in(self, directives):
...
         '''
-        Stages the content of the staging directive into the pilot's
-        staging area
+        fetch `staging_output.tgz` from the pilot sandbox, and store in $PWD
         '''
 
-        # send the staging request to the pmg launcher
-        self._pmgr._pilot_staging_output(self.as_dict(), directives)
+        try:
+            psbox = self._session.get_fs_dir(self._pilot_sandbox)
```
iparask: Is this on the pilot level or unit level?

andre-merzky:
Yes - additionally to the `pilot.stage_input()` call which we had for a while, this adds the `staging_input` and `staging_output` attributes to a pilot description, very similar to what we have on the unit description.  Those staging directives are honored on pilot submission and termination.

src/radical/pilot/pmgr/launching/default.py
```diff
@@ -241,60 +241,62 @@ def _handle_pilot_output_staging(self, pilot, sds):
...
-            sd['pmgr_state'] = rps.DONE
+            except:
+                self._prof.prof('staging_out_fail', uid=pid, msg=did)
+                sd['state'] = rps.FAILED
 
-            self.publish(rpc.CONTROL_PUBSUB, {'cmd': 'pilot_staging_output_result',
-                                              'arg': {'pilot': pilot,
-                                                      'sds'  : [sd]}})
 
-            self._prof.prof('staging_out_stop', uid=pid, msg=did)
```
iparask: Why is this timestamp removed?

andre-merzky: It's now in line 298, and only recorded on success (as it should be)

iparask: Missed it sorry.

PR #2009, opened 2019-12-16
Fix/prrte opts

andre-merzky:
This addresses PRRTE issues at scale


src/radical/pilot/agent/lm/prte.py
```diff
@@ -74,6 +74,13 @@ def lrms_config_hook(cls, name, cfg, lrms, log, profiler):
         if profiler.enabled:
             prte += ' --pmca orte_state_base_verbose 1'  # prte profiling
 
+        # large tasks imply large message sizes, and we need to account for that
+        # FIXME: we should derive the message size from DVM size - smaller DVMs
+        #        will never need large messages, as they can't run large tasks)
+        prte += ' --pmca ptl_base_max_msg_size %d' % (1024 * 1024 * 1024 * 1)
+
+        # debug mapper problems for large tasks
+        prte += '  -pmca orte_rmaps_base_verbose 100'
```
mturilli: Should this be guarded by `if verbose:`? I am worried that enabling verbose logging with large tasks increase further slows down the system, possibly increase likelihood of failure?

andre-merzky: right, done!

PR #2010, opened 2019-12-16
Fix/idle publish

andre-merzky:
avoid DB pushs on empty loops


mturilli:
Before merging, please note that *tests* fail with travis (not formatting).

andre-merzky:
Thanks - those are the same tests failing as we discussed before I am afraid.

PR #2013, opened 2019-12-16
ensure session dburl trickles down

andre-merzky:
This fixes #2006

Issue #2002, opened 2019-12-05
job tests faulty: need to mock `os.environ`.


andre-merzky:
I had to pass a recent PR despite a failing test.  The test was the ssh launch method which failed because the OS environment for `LD_LIBRARY_PATH` differed from what we encoded in the expected result.  If I change the expected result, then the tests fail when running for me locally.  Ultimately the LM is making a runtime decision based on the actual os env, and thus the result cannot be hard coded this way.  I assume we need to mock the os env for the tests, and this is what this ticket is about.

andre-merzky:
This should be discussed on the next devel call.

mtitov:
@andre-merzky  this issue was resolved indirectly (#2075), should we close it then?

andre-merzky:
I actually don't see how that was resolved - the `LD_LIBRARY_PATH` env setting should still show up in the test results?

mtitov:
That env variables were [set](https://github.com/radical-cybertools/radical.pilot/blob/devel/tests/test_lm/test_unit/test_ssh.py#L51) and then were tested [here as example](https://github.com/radical-cybertools/radical.pilot/blob/devel/tests/test_cases/unit.000003.json#L42)

andre-merzky:
Great, thanks!

PR #2015, opened 2019-12-17
sync umgr and pmgr

andre-merzky:
I merged #2013 before tests passed - mea culpa Matteo!  This fixes the problem which was caused by umgr and pmgr getting out of sync, by ensuring their construction and initialization follows the same sequence.  A diff between `unit_manager.py` and `pilot_manager.py` will motivate the changes in this PR.

### DIFF
--- a/src/radical/pilot/pilot_manager.py
+++ b/src/radical/pilot/pilot_manager.py
@@ -160,7 +160,7 @@ class PilotManager(rpu.Component):
 
     # --------------------------------------------------------------------------
     #
-    # EnTK forks, make sure we don't carre traces of children across the fork
+    # EnTK forks, make sure we don't carry traces of children across the fork
     #
     def _atfork_prepare(self): pass
     def _atfork_parent(self) : pass
--- a/src/radical/pilot/pilot_manager.py
+++ b/src/radical/pilot/pilot_manager.py
@@ -70,6 +70,8 @@ class PilotManager(rpu.Component):
         **Arguments:**
             * session [:class:`rp.Session`]:
               The session instance to use.
+            * cfg (`dict` or `string`):
+              The configuration or name of configuration to use.
 
         **Returns:**
             * A new `PilotManager` object [:class:`rp.PilotManager`].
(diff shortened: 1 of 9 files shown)


andre-merzky:
Hey @mturilli : I fixed some tests (spaces *urgh*), but the tests still  fail because of #2002.  The umgr (point of this PR) seems clean now.

src/radical/pilot/pilot_manager.py
```diff
@@ -150,14 +152,16 @@ def __init__(self, session, cfg='default'):
...
-    def initialize_common(self):
+    def initialize(self):
 
         # the manager must not carry bridge and component handles across forks
         ru.atfork(self._atfork_prepare, self._atfork_parent, self._atfork_child)
 
 
     # --------------------------------------------------------------------------
     #
+    # EnTK forks, make sure we don't carre traces of children across the fork
```
mturilli: Just a typo: carre -> carry

src/radical/pilot/pilot_manager.py
```diff
@@ -198,7 +193,7 @@ def close(self, terminate=True):
         self._terminate.set()
         self._rep.info('<<close pilot manager')
 
-        # we don't want any callback invokations during shutdown
+        # disable callbacks during shutdown
         # FIXME: really?
```
mturilli: Found this funny. Why really?

src/radical/pilot/unit_manager.py
```diff
@@ -152,11 +159,38 @@ def __init__(self, session, scheduler=None):
...
+    #
+    def initialize(self):
+
+        # the manager must not carry bridge and component handles across forks
+        ru.atfork(self._atfork_prepare, self._atfork_parent, self._atfork_child)
+
+
+    # --------------------------------------------------------------------------
+    #
+    # EnTK forks, make sure we don't carre traces of children across the fork
```
mturilli: carre -> carry

src/radical/pilot/pilot_manager.py
andre-merzky: In one way, I don't want to get notified ' you are terminating' when I initiate a termination - that is redundant.  And we need to somewhat carefully check that we don't run circles during termination (callback encountering termination and thus trying to terminate...).  Other than that I am not sure - we should eventually revisit that.

Thanks for the typos, will fix.

PR #2017, opened 2019-12-17
decrease hb interval

andre-merzky:
We recently increased the default HB timeout to cater f or slow filesystems.  In that PR, we also increased the HB interval to keep the number of HB messages  consistent.  That slows down client and agent bootstrapping though, since we usually miss the first heartbeat message on component startup, and the second message now is delayed significantly.

This PR decreases the HB interval to avoid that.  In the long run, we may want to adapt the interval somewhat dynamically, but that is for another day...


andre-merzky:
Yes, the previous PR (which increased the timeout) was specifically for summit.  This PR does not change that timeout, only the message interval.

PR #2018, opened 2019-12-17
improve startup logging

andre-merzky:
increase verbosity of bridge startup


andre-merzky:
Merging before release

PR #2020, opened 2019-12-17
scattered fixes

andre-merzky:
- remove invalid call in agent scheduler
  - silence non-primary sessions


andre-merzky:
> One warning, two tests fail with actual differences in the CMD, not just spaces

One is the `os.environ` problem, the other the PRTE verbosity shield.   Will fix *sigh*.  (There must be better ways to test this stuff...)

PR #2021, opened 2019-12-18
avoid invalid gpu maps

andre-merzky:
This PR replaces #2001 - @lee212, let me please know if this solves your problem.


andre-merzky:
@mturilli , @lee212 : this PR should fix summit - it would be great if you'd find time to review it.  Thanks!

andre-merzky:
Any task with no GPU will fail on summit w/o this fix, so the fix is quickly confirmed with the canonical `00_getting_started.py`.

andre-merzky:
Thanks @mturilli !

PR #2032, opened 2020-01-13
make sure that resource details are preserved

andre-merzky:
This ensures that `resource_details` as recorded by the agent are not overwritten on explicit pilot cancellation.


src/radical/pilot/agent/agent_0.py
```diff
@@ -252,23 +252,6 @@ def finalize(self):
...
-        except: pass
-        try   : log = open('./agent.0.log', 'r').read(1024)
-        except: pass
-
-        self._dbs._c.update({'type': 'pilot',
-                             'uid' : self._pid},
-                            {'$set': {'stdout' : rpu.tail(out),
-                                      'stderr' : rpu.tail(err),
-                                      'logfile': rpu.tail(log)} })
-
```
andre-merzky: This code was removed as it duplicated a DB update done via the update worker.

andre-merzky:
Hi @mturilli : do you have the time for this one also?  Should be quick (I hope).

src/radical/pilot/agent/agent_0.py
mturilli: Are we sure that the other copy of this code is *always* called? Do we have a test for it?

andre-merzky: Yes - 4 lines above the remove part, `update_db()` is *always* called, and `update_db()` *always* runs the same lines (20 lines up in the same file).

PR #2037, opened 2020-01-24
Fixing syntax errors for examples

iparask:
These examples have not been tested. The fixes are syntactic in nature to allow sphinx to correctly build the documentation.

### DIFF
--- a/examples/data_staging/io_staging_pipeline.py
+++ b/examples/data_staging/io_staging_pipeline.py
@@ -19,7 +19,7 @@ def pilot_state_cb (pilot, state):
     if not pilot:
         return
 
-    print "[Callback]: ComputePilot '%s' state: %s." % (pilot.uid, state)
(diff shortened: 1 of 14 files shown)


examples/data_staging/io_staging_pipeline.py
```diff
@@ -143,20 +143,20 @@ def unit_state_cb (unit, state):
 
     except Exception as e:
         # Something unexpected happened in the pilot code above
-        print "caught Exception: %s" % e
+        print( "caught Exception: %s" % e)
         raise
```
mturilli: From here and for all the following `print` statements, should we delete the space after the opening `(`?

andre-merzky:
I did a fair amount of linting on the examples.  While more can (and should) be done, we may first want to weed them out: from a cursory look while linting I am sure that the majority of them does not run anymore, and most should be removed, IMHO.

PR #2044, opened 2020-01-29
revert from bin gtod and bc to awk - if awk is missing, we give up.

### LINKED ISSUE #2012: bc is not universally installed
surprising, but true. We need to cater for that case in the bootstrapper...

andre-merzky:
This fixes #2012


andre-merzky:
Thanks @AymenFJA !

PR #2046, opened 2020-01-31
Update virtualenv command and make p3 a requirement

mturilli:
From python 3.3. onward, the documented way to create a virtualenv is `python3 -m venv ~/ve/`, see https://docs.python.org/3/library/venv.html. While working on Frontera I incurred in issues with using virtualenv, issues that were addressed by using the documented way to create a virtualenv.

### DIFF
--- a/docs/source/installation.rst
+++ b/docs/source/installation.rst
@@ -21,13 +21,14 @@ Software
 
 RADICAL-Pilot requires the following packages:
 
-* Python >= 2.7
-* virtualenv >= 16
-* pip >= 19
+* Python >= 3.7
+* pip >= 20
+
+.. * virtualenv >= 16
 
 or
 
-* Anaconda Python >= 2.7
+* Anaconda Python >= 3.7
 
 All other dependencies are installed automatically by RADICAL-Pilot installer.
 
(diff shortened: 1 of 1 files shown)


docs/source/installation.rst
```diff
@@ -120,15 +121,17 @@ via PyPi
 --------
 .. code-block:: bash
 
-    virtualenv --system-site-packages $HOME/ve
+    python3 -m venv $HOME/ve
     source $HOME/ve/bin/activate
```
andre-merzky: I made the same comment on the EnTK doc fix you proposed: maybe we wait a bit and see how `venv` behaves before documenting it as the default / recommended deployment path?  Not saying it's not the right thing to do, but I'd rather make sure it is stable (it's new-ish code which recently had significant changes, and it is not feature-complete wrt. virtualenv).

mturilli: Fair, answered there. How do we want to track the TODOs? I would usually use a ticket but maybe it is time to give a try to the project feature of github?

andre-merzky: > How do we want to track the TODOs? I would usually use a ticket but maybe it is time to give a try to the project feature of github?

Sure, why not?

andre-merzky:
Thanks @mturilli !

Issue #2049, opened 2020-02-01
RP reports `timeout` even when pilot terminated properly


andre-merzky:
See radical-collaboration/Glatard-Lab/issues/1 for example.  The `timeout` message is confusing and wrong.

andre-merzky:
This was fixed by #2050

PR #2050, opened 2020-02-02
re-order termination

andre-merzky:
`self.terminate` was set too early in the pilot manager, so that the `wait_pilots` call never actually waited for the pilots, but bailed out immediately complaining about timeout.  This PR fixes that - but also cleans up the state reporting done by the pilot, to re-introduce the `rp_state_push` utility to account for clean state pushes on bootstrap failures.


mturilli:
Yes! 👍 👍 👍  We should probably get this out with the next release? :D

andre-merzky:
Yes, absolutely.

andre-merzky:
This fixes #2049 and radical-collaboration/Glatard-Lab/issues/1

andre-merzky:
Thanks @iparask !

PR #2055, opened 2020-02-03
str default should not be `None`

### LINKED ISSUE #2053: All units are in a folder called None.
I executed locally the `00_getting_started.py` example and all the units ( unit shell scripts) went into a folder with the name `None`. The same name was given also to the STDOUT files of the units.

```
$ python -V
Python 3.7.6
$ radical-stack

  python               : 3.7.6
  pythonpath           :
  virtualenv           : /home/iparask/rct_test

  radical.pilot        : 1.0.0-v1.0.0-66-g579be53@devel
  radical.saga         : 1.0.1-v1.0.1-7-g265601d@devel
  radical.utils        : 1.0.0-v1.0.0-31-ga6ecd7a@devel
```

andre-merzky:
This fixes #2053.  I am positively surprised that this went unnoticed before: basically means that units are well isolated even when sharing a sandbox ;-)  Either way, this needs to be released ASAP.  Thanks for opening the ticket, @iparask !

### DIFF
--- a/src/radical/pilot/compute_unit_description.py
+++ b/src/radical/pilot/compute_unit_description.py
@@ -357,16 +357,16 @@ class ComputeUnitDescription(ru.Description):
     }
 
     _defaults = {
-               EXECUTABLE      : None        ,
-               KERNEL          : None        ,
-               NAME            : None        ,
-               SANDBOX         : None        ,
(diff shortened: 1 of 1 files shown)


andre-merzky:
Thanks @iparask !

PR #2062, opened 2020-02-06
fixes in the User Guide and other sections

mtitov:
current section Getting Started

- no LogReporter class (in radical.utils) -> Reporter (since commit a9337f8)
- fix typos and references


mtitov:
Example [00_getting_started.py](https://github.com/radical-cybertools/radical.pilot/blob/devel/examples/00_getting_started.py#L75) contains method call `report.progress_tgt` with parameter `label` which is not in the corresponding module of the installed package of RU (1.0.0)

andre-merzky:
Ah, good catch!  Let me please check how that got out of sync with RU before merging this PR.

mturilli:
I am going to iterate on the text too, will let a message in the PR once I am done. Please feel free to review my editing.

andre-merzky:
The log reporter problem (no `label` arg) is fixed in devel, and since we need to release RP and RU in sync (because of the Description changes), this should not be a problem.  This PR can be merged when ready IMHO.
Thanks!

mturilli:
In the guide we write: "You can control that output with the ``RADICAL_PILOT_VERBOSE`` variable, which can be set to the normal Python logging levels, and to the value ``REPORT`` to obtain well formatted output.  We assume the ``REPORT`` setting to be used when referencing any output in this chapter.". Is this still the case? I see that `verbose  = os.environ.get('RADICAL_PILOT_VERBOSE', 'REPORT')` and `os.environ['RADICAL_PILOT_VERBOSE'] = verbose` are not in `00_getting_started.py` anymore but it is in all the other examples. Grepping on RCT does not show an apparently meaningful use of`REPORT`.

andre-merzky:
@mturilli : Thanks for pointing that out, I'll fix that section.

mturilli:
Thanks @andre-merzky . Note that I have also fixed several other inaccuracies. You may want to have a look at the sections/parts of the text related to `session` and its default behavior. I rewrote them looking at the current definition of `close()` in `src/radical/pilot/session.py`.

mturilli:
In `08_unit_environment.rst` we write: "This makes the environment variable `TEST` available during CU execution. Some other variables, such as the `RP_UNIT_ID` above, are set by RP internally and are here used for demonstration.". Do we have a list of the variables defined by RP that can be used in the unit description?

mturilli:
I have now finished my iteration of this section. @andre-merzky , I will be waiting for your input and then I will approve the PR.

andre-merzky:
Thanks both!  I fixed what I found documented on the logger and reporter settings.

mturilli:
@andre-merzky , I think the diagrams at the end of `developer.rst` are outdated. If you confirm, we should probably update them.

PR #2064, opened 2020-02-08
fix cud schema

andre-merzky:
I made a mistake in specifying staging directives in the task descriptions as lists of strings.  We do indeed mostly use that format, but I forgot that we in fact also support a dict format for specifying more complex staging ops.  This patch fixes that by removing the type spec for staging declarations altogether.  That's not optimal but will have to suffice for now as this needs to be released.  I'll open a ticket to track the creation of a proper value checker.


andre-merzky:
ping @iparask : would you mind having a look at this today?  Let me know if you don't manage though!  Thanks.

iparask:
Hello @andre-merzky, I am ok with this PR. May I suggest that the ticket is opened first, references the PR, and then the PR gets merge?

PR #2073, opened 2020-02-13
fix stack deps

### LINKED ISSUE #2072: Wrong setup requirements versioning
RP's setup file has the wrong minimum version in dependencies. For example, the latest setup file requires `radical.utils>=0.90`, but RP imports `Description` from `radical.utils`, which does not exist until version `1.1`, I believe. As a result this error is produced:
`AttributeError: module 'radical.utils' has no attribute 'Description'`

andre-merzky:
This fixes #2072.


setup.py
```diff
@@ -245,8 +245,8 @@ def run(self):
     'package_data'       : {'': ['*.txt', '*.sh', '*.json', '*.gz', '*.c',
                                  '*.md', 'VERSION', 'SDIST', sdist_name]},
   # 'setup_requires'     : ['pytest-runner'],
-    'install_requires'   : ['radical.utils>=0.90',
-                            'radical.saga>=0.90',
+    'install_requires'   : ['radical.utils>=1.1',
+                            'radical.saga>=1.0',
```
iparask: Shouldn't this be 1.1 also?

andre-merzky: 1.0 will work - there was no change in RS which RP depends on between 1.0 and 1.1.

PR #2077, opened 2020-02-18
fix issue 2075 part 2; fix fork test

andre-merzky:


tests/test_rm/test_unit/test_lsf_summit.py
```diff
@@ -18,7 +18,7 @@
 def test_configure(mocked_init, mocked_raise_on):
 
     # Test 1 no config file
-    os.environ['LSB_DJOB_HOSTFILE'] = 'tests/test_cases/rm/nodelist.lsf'
+    os.environ['LSB_DJOB_HOSTFILE'] = '../../test_cases/rm/nodelist.lsf'
```
andre-merzky: Doesn't this then depend from where the tests are running?  Travis is configured to run via `pytest tests/` - and then those paths were correct I believe - see new failing tests.

If we want to be agnostic to that location, we probably need to do something like

```python
os.environ['LSB_DJOB_HOSTFILE'] = '%s/../../test_cases/rm/nodelist.lsf' % os.path.dirname(__file__)
```

which makes the  path relative to the test case.   Not sure if that's worthwhile?

mtitov: @andre-merzky you are absolutely right, I missed that point, will return it back. BTW to keep the same way of accessing the tests - should this [#L16](https://github.com/radical-cybertools/radical.pilot/blob/8385a7d9a65992b1dca28ec65387c85a5c0fb6ac/tests/test_lm/test_unit/test_common.py#L16) be changed accordingly?

andre-merzky: Nice find :-)  I don't mind either way.  On the one hand it's correct (as its location independent), OTOH consistency is also nice...  So `¯\_(ツ)_/¯`

PR #2095, opened 2020-03-11
Configuration data for Longhorn@TACC

mtitov:


mtitov:
connected to the SAGA PR [#771](https://github.com/radical-cybertools/radical.saga/pull/771)

PR #2096, opened 2020-03-18
Added config parameters for Longhorn: cores_per_node, gpus_per_node

mtitov:
Should some additional information (e.g., about GPUs) be added to the config? (regarding related [issue](https://github.com/2019-ncovgroup/DrugWorkflows/issues/11))


andre-merzky:
Yes, it helps when the actual configuration cannot be picked up correctly.  Its one more thing to maintain though - but, fortunately, machines don't change those values too often (only during upgrades, if at all).

PR #2100, opened 2020-03-22
Feature/test schedulers

andre-merzky:
This PR adds tests for agent scheduler implementations.

@AymenFJA , @iparask : I opened the PR since the work on this seems to stall - or is it completed?  Can you please clarify the state of affairs, and we'll discuss at the next devel meeting on if and how to proceed.  Thanks.


iparask:
Yes, I will have a look Monday morning.

AymenFJA:
Hey @andre-merzky , apologies! we are missing `test_yarn` and `test_torus`. I stopped working on it due to my thesis and graduation process back then. I will start working on it, also I agree, let's discuss it in the devel meeting.

andre-merzky:
> Hey @andre-merzky , apologies! we are missing `test_yarn` and `test_torus`. I stopped working on it due to my thesis and graduation process back then. I will start working on it, also I agree, let's discuss it in the devel meeting.

The yarn scheduler is archived, so we don't need that.  The torus scheduler should probably be archived, too, as that is specific to IBM Blue Genes, and those machines are not around much anymore.  So that should not be a problem.  Also I would assume that a subset of tests would also be useful - no need to wait for completeness :-)

iparask:
They are out of sync quite a lot. I am not sure it makes sense to merge them so that they fail.

We need to bring the tests back up to date and then move forward

andre-merzky:
Thanks both for the feedback!

iparask:
@andre-merzky can you please review?

andre-merzky:
@iparask : instructions unclear :-)  You asked for review but closed it?

PR #2102, opened 2020-03-22
Fix/srun placement

andre-merzky:
fix srun placement.  Note that Longhorn complains (non-fatally) about unsupported cpu bindings - but the placement seems to behave better than the previous implementation even there.


andre-merzky:
This is not the last iteration of the srun launch method, I am sure - but it is less broken than the current one...

PR #2103, opened 2020-03-22
Hotfix/cobalt

andre-merzky:
This PR adds preliminary support for Cobalt (and thus Theta).

Unrelated, the PR also moves the pilot sandboxes on comet from home into scratch.  This change should not have happened on this branch, apologies, but it is small and I thus left it here...


mturilli:
Tests fail with what seem to me meaningful errors

PR #2104, opened 2020-03-22
Hotfix/stampded ibrun

andre-merzky:
This gets the ibrun launch method back into working order on Stampede2.  The PR is needed for RX experiments to function properly: the Gromacs binary used by Srinivas does not like other launch methods.  The PR basically catches up with the changed slots structure.

@mtitov : I assigned this to you, too - but I can't remember if we looked at the slot structure in our previous calls, so let me know if you want to chat about it!


andre-merzky:
@iparask : any idea where the travis test link went?

andre-merzky:
@iparask : ping

iparask:
I think we need to close the PR and open it again. I do not know what has happened.

andre-merzky:
Going to merge, knowing that the ibrun work on Longhorn will iterate on this code before the next release.

PR #2105, opened 2020-03-22
Feature/cpi

andre-merzky:
This PR provides a mechanism to define a master/worker style task overlay for the pilot agent.  The use case here is ScaleMS, where the ScaleMS workflow enactor is expected to manage individual *custom* worker instances within the pilot which perform fine- or coarse grained function calls on behalf of the application.  Another use case turns out to be efficient load balancing for the Covid-19 workflows, and since the mechanism has been tested there, I consider it ok for an initial release.

Please have a look at the included examples, and happy to discuss in more detail in the upcoming devel call.


andre-merzky:
> I see what you are trying to do here and it is interesting. I am wondering though how is this not already supported by the application level communication channels? I would think that with the communication channels, it would have been somewhat straightforward to implement a master-worker application.

The main difference is that the application submits only the master task - the workers are spawned by the master *while running on the pilot*.  The PR basically gives us dynamic workloads, where the workload tasks themself can spawn new tasks, and don't have to ping back to the application to perform that action.

To do so, the master base class hooks into the agent scheduler's input queue and pushes CU messages directly to the stager, skipping all umgr states and activities.

src/radical/pilot/agent/executing/popen.py
```diff
@@ -185,7 +202,7 @@ def spawn(self, launcher, cu):
 
             # Create string for environment variable setting
             env_string = ''
-            env_string += '. %s/env.orig\n'               % self._pwd
+          # env_string += '. %s/env.orig\n'               % self._pwd
```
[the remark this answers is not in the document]
andre-merzky: No, it's not - but this approach alas fails sometimes.  We are still working on clean separation of task environments, and this PR, alas, will not cover it.

src/radical/pilot/task_overlay/master.py
```diff
@@ -0,0 +1,183 @@
...
+                time.sleep(0.1)
+
+        elif uids:
+            while True:
+                with self._lock:
+                    states = [self._workers[uid]['state'] for uid in uids]
+                if states.count('ACTIVE') == len(states):
+                    return
+                time.sleep(0.1)
+
```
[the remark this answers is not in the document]
andre-merzky: Oh, thanks for noting that - I need to look into this a bit.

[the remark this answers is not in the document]
andre-merzky: This has been addressed now.

src/radical/pilot/agent/bootstrap_0.sh
```diff
@@ -6,6 +6,20 @@ export PS1='#'
 unset PROMPT_COMMAND
 unset -f cd ls uname pwd date bc cat echo
 
+
```
mturilli: If the following is useful for debugging, should we enable it only when we debug?

src/radical/pilot/configs/agent_default.json
```diff
@@ -36,19 +36,19 @@
     #
     "bridges" : {
         "agent_staging_input_queue"  : { "kind"      : "queue",
-                                         "log_level" : "error",
+                                         "log_level" : "debug",
```
mturilli: Should 'debug' be the default level in our releases? I would assume it should be 'error'?

mturilli:
Tests fail in a way that seem meaningful to me.

src/radical/pilot/agent/bootstrap_0.sh
andre-merzky: The overhead is near zero, so I'd rather leave it in instead of adding verbosity on the shell level (the bootstrapper does not look at the log level at all right now).

src/radical/pilot/configs/agent_default.json
andre-merzky: Right, will revert to `error`.

PR #2111, opened 2020-04-01
Hotfix/launcher termination

andre-merzky:
Houston, we have a problem...

I was looking into the thread limit problem we recently faced on Frontera, and found that in many cases, the pmgr.launching component did not terminate.  I checked on other machines and found the same on Summit, but not on localhost.  It turned our to be related to the SAGA adaptor used for the pilot launching.  What happened is the following:

Sometime in June 2019, a fix was made to the Python threading module to get rid of those pesky `exception in bootstrap_inner` messages which randomly appeared on thread termination, and which shows random stack traces.  You may remember those - or not if your brain does the sane thing and blends out stuff like that.  Either way, while I applaud the patch of fixing that problem, it changed the code path taking during thread termination *for daemon threads*.  

Sometime in December, we changed our process management to be heartbeat based.  For that to be functional, we changed *all* threads to daemon threads (process termination will not wait for daemon threads to complete).  Specifically, the threads used by SAGA to watch process health are now daemon threads.

Along with those changes, we got rid of quite some code which manages thread termination, as daemon threads *always* terminate - that's the definition of daemon threads.

Now, it happens that the RS threads use a **private signal** `self._stop = threading.Event()` to communicate with the main thread and to allow for early termination, for example on job service instance close or garbage collection, i.e., before the actual process termination kicks in and kills all daemon threads.

The Python code changes I mentioned above are related to a lock management problem.  That lock is now cleanly released for all threads on termination, either explicit, daemonized, or garbage collected.  The change however now triggers a code path when terminating daemonized, garbage collected threads to call a private **function** `self._stop()` -- Oh oh..

So, when our SAGA threads which have a `_stop` Event are daemonized, and when we close the job service, then the threads get garbage collected eventually, and then the threading module will call `self._stop()` **on that `threading.Event` instance** which our implementation re-defines, and kaboom, and the fucking daemon(!) thread will just hangs and will not terminate, and so won't the pmgr.launching component.

The fix is easy: rename `self._stop` to something else.  We use `self._term` in RP and RU, so `self._term` it is.  Why I am telling you all this?  Because (a) it was a nightmare to debug and find, and I wanted to leave a trace of it, but also (b) because it means that we can't rely on daemon threads to die on termination and may need to re-introduce some sanity checks.  Urgh.

Last but not least, you may want to check on the different login nodes of Frontera and Summit if you left rp instances behind.  The following commands should help:
```
$ name=$(uid -un)
$ ps -f -u $name | grep rp.pmgr_launching.0
```

You may want to kill those (which will also tear down some other dependent processes and threads:
```
$ ps -f -u $name | grep rp.pmgr_launching.0 | grep -v grep | cut -c 9-15 | xargs -n 1 -t kill
```

This PR contains some debugging support and cleanup to make observing this problem easier, and will be accompanied by two similar PRs on RS and RU level.

### DIFF
--- a/examples/00_getting_started.py
+++ b/examples/00_getting_started.py
@@ -89,7 +89,7 @@ if __name__ == '__main__':
         # Submit the previously created ComputeUnit descriptions to the
         # PilotManager. This will trigger the selected scheduler to start
         # assigning ComputeUnits to the ComputePilots.
-        umgr.submit_units(cuds)
+      # umgr.submit_units(cuds)
 
         # Wait for all compute units to reach a final state (DONE, CANCELED or FAILED).
         umgr.wait_units()
--- a/examples/00_getting_started.py
+++ b/examples/00_getting_started.py
@@ -89,7 +89,7 @@ if __name__ == '__main__':
         # Submit the previously created ComputeUnit descriptions to the
         # PilotManager. This will trigger the selected scheduler to start
         # assigning ComputeUnits to the ComputePilots.
-      # umgr.submit_units(cuds)
+        umgr.submit_units(cuds)
 
         # Wait for all compute units to reach a final state (DONE, CANCELED or FAILED).
         umgr.wait_units()
--- a/src/radical/pilot/agent/launch_method/prte.py
+++ b/src/radical/pilot/agent/launch_method/prte.py
@@ -73,17 +73,17 @@ class PRTE(LaunchMethod):
         prte += ' --report-uri %s' % furi
         prte += ' --hostfile %s'   % fhosts
 
-        if profiler.enabled or True:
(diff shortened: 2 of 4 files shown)


andre-merzky:
This became more urgent now, first users are hitting thread limits on summit, presumably due to this (the user has 17 launchers still active on the same login node).  It would be great if any one of you can give this (and the related PR's) a quick today, if you can.  Feel free to ping me if you want to discuss.

andre-merzky:
Thanks Matteo.  Ack on the improving.  The main point really would be that we should consider shielding all daemon terminations again.  But this one place will at least trigger on general problems on RP, RS and RU layers as the launcher uses all three of them.

PR #2112, opened 2020-04-03
support for Princeton's Traverse cluster

andre-merzky:


src/radical/pilot/agent/agent_0.py
```diff
@@ -121,7 +121,14 @@ def _hb_term_cb(self):
     #
     def _connect_db(self):
 
-        # TODO: this needs to evaluate the bootstrapper's HOSTPORT
+        hostport = os.environ.get('RADICAL_PILOT_DB_HOSTPORT')
```
mtitov, line 124: I'd like to pint out 2 things here:
1) this env-variable is also checked later in DBSession, should then it be kept only in one place? I would go for keeping it in DBSession (`cfg.dburl` is not updated later, but I guess it would be ok if it would, thus whatever db-related info comes into DBSession it processes it accordingly);
2) there are 2 db-related env-variables: `RADICAL_PILOT_DBURL` and `RADICAL_PILOT_DB_HOSTPORT`. Should be kept just one? or if it is expected that host and port could be changed during to session give a different name to distinguish that one is static and another is dynamic (?)

andre-merzky: Thanks @mtitov , good points. 

1) You are right, and the DBSession was actually cnnecting to the DB all right - but the agent needs the corrected DBURL in its config, so that it can be passed on to other components, and specifically to the UpdateWorker which uses it's own DB connection (i.e, not DBSession).  Nevertheless, it's not good to have that code in two places.  Now that the agent does the expansion early, we can probably loose the version in DBSession.  I'll check and fix.

2) `RADICAL_PILOT_DB_HOSTPORT` is set by the bootstrapper, but only if a tunnel is actually enabled.  The DBURL itself is passed to the agent via the config file, not via the environment variable.  So the agent obtains those two pieces of information from different channels (tunnel hostport env from bootstrapper, original dburl from config file), and needs to build a new dburl out of those.  That can also happen somewhere else - but I'd rather not move it to the bootstrapper as to avoid parsing json config files on shell level, thus the two different sources.

PR #2116, opened 2020-04-10
fix queue naming, prte tmp dir

andre-merzky:
This contains two small fixes: move PRTE tmp files to, well, `/tmp/` to avoid thrashing the shared FS, and make naming of task overlay channel consistent with use (request instead of work channel).


andre-merzky:
Fixes to the funcs executor are out of scope - I opened #2143 for that.

PR #2117, opened 2020-04-10
Fix/gpu proc count

andre-merzky:
Fix gpu process counting for aprun (`gpu_processes` does not imply additional processes).


andre-merzky:
The remaining pep8 warnings are in the works in the scope of other PRs / branches.

PR #2120, opened 2020-04-13
Feature/test schedulers

iparask:
I'm opening a new PR, so that @andre-merzky can review. A couple of comments from my side.

- Base class test should use json file.
- Continuous scheduler test should be in the same style as the base class test. There is no need for a class.

Everything needs to be linted. That is true for both tests.


andre-merzky:
Another 3 days for review.

andre-merzky:
ping @mturilli , @andre-merzky , @darkwhite29

tests/test_scheduler/test_unit/test_base.py
```diff
@@ -0,0 +1,286 @@
...
+try:
+    import mock
+except ImportError:
+    from unittest import mock
+
+
+# ------------------------------------------------------------------------------
+#
+@mock.patch.object(AgentSchedulingComponent, '__init__', return_value=None)
+def test_change_slot_states(mocked_init):
```
andre-merzky, line 18: I still think that test cases should be external json files.  What will this method look like in 5 years when we added 10 cases?   Who is going to maintain that when the function call or member var name changes?

The code could basically reduce to
```python
for test_case in glob('test_slots_*.json'):
    pre_cond, slots, post_cond = ru.read_json(test_case)
    component = AgentSchedulingComponent()
    component.nodes = pre_cond
    component._change_slot_states(slots=slots, new_state=1)
    assert(component.nodes == post_cond), test_case
```

with no loss of test coverage - but easy to maintain on code changes, and easy to add test cases: just drop a json file.

tests/test_scheduler/test_unit/test_base.py
```diff
@@ -0,0 +1,286 @@
...
+                                            "core_map": [[0]],
+                                            "name": "a",
+                                            "gpu_map": None,
+                                            "uid": 1,"mem": None}],
+                                 "lm_info": "INFO",
+                                 "gpus_per_node": 6,
+                                 })
+@mock.patch.object(AgentSchedulingComponent, '_change_slot_states',
+                   return_value=True)
+def test_try_allocation(mocked_init, mocked_schedule_unit, mocked_handle_cuda,
```
andre-merzky, line 156: Same here: unless I misunderstand the purpose of the method, it is testing the most basic task.  Which is fine - but we should make it easy to add test cases.

andre-merzky:
@iparask : ping?

iparask:
Hello @andre-merzky, sorry for the delay. What do you think now?

andre-merzky:
Thanks @iparask !  Should we do the same for the other descriptions?

iparask:
@andre-merzky , do you mind reviewing this PR? It is open for so much time it starts to not make sense. Thank you!

PR #2122, opened 2020-04-14
SRUN latest updates

mtitov:
Current implementation is sufficient to run executables with `srun` (latest tests were conducted at Frontera)


mtitov:
> Please be aware that several slots could point to the same node. I am not sure if that would make a difference here? For examples, for a 4-process task, two slots may end up on node, 1, one slot on node 2, and the last slot on node 3.

In general I agree, but I see that the same approach is used in `slurm` (https://github.com/radical-cybertools/radical.saga/blob/devel/src/radical/saga/adaptors/slurm/slurm_job.py#L556), and that's why think to keep it

andre-merzky:
> In general I agree, but I see that the same approach is used in slurm (https://github.com/radical-cybertools/radical.saga/blob/devel/src/radical/saga/adaptors/slurm/slurm_job.py#L556), and that's why think to keep it

I see.  Well, SAGA always considers full nodes, and so that is assumed to hold there.  For RP, tasks may be located on parts of a node (or parts of several nodes).  Maybe the code also does the expected, I don't really now - I wanted to point out so that the `n_nodes` counter in line 89 will, in the above case, count 4 nodes, where the task actually has for *processes* (spanning 3 nodes).  It will never undercount, but may overcount nodes (possibly beyond the number of allocated nodes).  If `srun` does not mind, then I don't either :-)

mtitov:
@andre-merzky it is not actually overcount, it just makes sure that defined number of nodes would be able to process `n_tasks` (e.g., `5 tasks, 4 cpn -> 2 nodes`) and it wouldn't go beyond what is set by `slurm`. It is as an upper threshold

> If `srun` does not mind, then I don't either :-)

it doesn't :)

andre-merzky:
I think we still misunderstand each other.  Let me try to dig though that.  The code is:

```python
        n_nodes = int(math.ceil(float(n_tasks) /
                                self._cfg.get('cores_per_node', 1))) \
                  if not slots else len(slots['nodes'])
```

It's a bit convoluted (should not use single line if/else), but it basically is this:

```python
if slots:
    n_nodes = len(slots['nodes'])
else:
   cpn = self._cfg.get('cores_per_node', 1)
    n_nodes = n_tasks / cpn
```

The second case is similar to SAGA, and I don't see a problem with that.

The first case though is different.  Assume that the pilot has 2 nodes and 8 cores per node.  Assume that we want to run a task which uses 12 cores, which easily fits.  The scheduler would schedule that task across both nodes, creating 12 process slots: 

```
   slot 0: [node 0, core 0]
   slot 1: [node 0, core 1]
   ...
   slot 8: [node 1, core 0]
   slot 0: [node 1, core 1]
   ...
```
So, `len(slots)` would be `12`.  But we don't want 12 nodes (and indeed don't have 12 nodes), but only need 2 nodes.  Is `srun` ok with the `n_nodes` spec?  It should be able to derive the *actual* node numbers from the allocation size and from number of tasks etc, but I am not sure it does?

Assuming the task uses 12 cores, then the scheduler would

mtitov:
Andre, thank you for clarification, I missed that you pointed on the second part of `n_nodes` definition. Made some changes accordingly: key point there is `n_nodes = len(set(nodelist))` for `slots` case

PR #2123, opened 2020-04-15
First commit to fix #2121

### LINKED ISSUE #2121: RP needs to handle sub-projects on summit
Some users are assigned to sub-projects (like `CHM155_001`) and need to be able to submit to it.  OTOH, the file system scratch dir is still located under the main project ID (`CHM155` in this case).  RP cannot separate the two.

Assuming the underscore separation is general (and I do not yet know a counter example, so lets assume it is), we should be able to derive the main project ID from it, and use it toward the file system interactions.

Alas, this is somewhat urgent.  @iparask : would you find the time today to look into this?  Otherwise I can do this first thing tomorrow morning.

iparask:
@andre-merzky, I think this is it. If you can wait one more day, I will include a test.

### DIFF
--- a/src/radical/pilot/session.py
+++ b/src/radical/pilot/session.py
@@ -758,6 +758,8 @@ class Session(rs.Session):
                             expand['pd.%s' % k.lower()] = v
                     sandbox_raw = sandbox_raw % expand
 
+                if '_' in sandbox_raw and 'summit' in resource:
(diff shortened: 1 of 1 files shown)


andre-merzky:
Yeah, I think this is the most reasonable approach right now - thanks.

iparask:
I haven't been able to write a test yet. Feel free to move this PR forward

PR #2124, opened 2020-04-20
Update bootstrap script

mtitov:


mtitov:
@andre-merzky ok, ack p.s. that is my passion to pack multiline expressions whenever it is possible :)

andre-merzky:
Ah, you would like my dotfiles and git configs :-D  But for scripts we share with others, that might not be a good idea :-)

PR #2127, opened 2020-04-22
add `deactivate` support for tasks

andre-merzky:
This patch lets the bootstrapper create a `deactivate` script during
startup, which essentially mimics the venv `deactivate` command.  Alas,
the latter is not available in forked processes, so the executor cannot
deactivate the venv cleanly for the tasks.  With this script in place,
a task can add

    `. $RP_PILOT_SANDBOX/deactivate`

to the list of `pre_exec` commands to restore the original values for
`PATH`, `PYTHONPATH` and `PYTHONHOME`.  Any following `pre_exec` which
loads a venv or conda env should then not be affected by the pilot's
venv.  We might want to make this the default behavior at some point.

This patch also contains some other cleanups which got carried over by
cherry-picking the deactivate patch from an unrelated branch.


src/radical/pilot/unit_manager.py
```diff
@@ -996,14 +996,19 @@ def wait_units(self, uids=None, state=None, timeout=None):
...
         with self._units_lock:
             states = [self._units[uid].state for uid in uids]
 
+        sdict = {state: states.count(state) for state in set(states)}
+        for state in sorted(set(states)):
+            self._rep.info('\t%-10s: %5d\n' % (state, sdict[state]))
+
+        if to_check: self._rep.warn('>>timeout\n')
+        else       : self._rep.ok  ('>>ok\n')
+
```
andre-merzky, line 1011: Oh, another unrelated change which would have warranted it's own PR - sorry!  It triggers the reported to print a state summary after `wait_units()`.

src/radical/pilot/agent/bootstrap_0.sh
```diff
@@ -146,6 +135,39 @@ export PYTHONNOUSERSITE=True
...
+                                        | cut -f 3 -d ' ' \
+                                        | cut -f 1 -d '=' \
+                                        | uniq)
+
+    # capture the original values in `deactivate`.  If no original value exists,
+    # unset the variable
+    for var in $changed
+    do
+        val=$(grep -e "^export $var=" env.orig | cut -f 2- -d '=')
+        if test -z "$val"
```
mtitov, line 161: `if ! test -z "$val"` (currently it does opposite: unsets if original values exists, and export if not)

mtitov:
In general, that's what I'd like to discuss - how to organize the environment for bootstrapping. I would propose to have some "isolated" (or "clean") environment, and only let for example `PATH`s & `PYTHON`s env variables pass through, everything else should be set by `pre_bootstrap_0/1` parameter(s) -> e.g. run the script with `env -i <cmd>` (should be researched more) and additionally to consider unalias processes (e.g., `unalias python`). And I would think that we wouldn't need to have to track changes in env-variables if we would have "isolated" environment.

p.s. related to these changes, if I'm not mistaken, `deactivate` file doesn't consider changes in env variables from `bootstrap_2.sh` (and specifically what will be set by `pre_bootstrap_1` -> `$PREBOOTSTRAP2_EXPANDED` internal variable).

p.p.s the usage of `deactivate` will be added later, right?

src/radical/pilot/agent/bootstrap_0.sh
andre-merzky: Ah, thanks, nice catch

andre-merzky:
We need to let more than `PATH` and `PYTHONPATH` through the bootstrapper, and specifically do not know in advance what we need to let through.  For example, the prte launch method needs the prte module loaded, which may need a specific compiler, but also sets other env variables like `PRTE_DIR`.  Similar for other launch methods, etc.  Those settings can vary from machine to machine, and those need to be visible to the different agent components.  Also, `PATH` and `PYTHONPATH` create most of the problems for task level virtualenvs, so the reset is needed for them anyway.

Ack on the `pre_bootstrap_0/1` settings.  They will be caught in many cases (those are usually a subset of `pre_bootstrap_0`), but that should be made more explicit.

As for usage: at the moment we can add the deactivate call manually to the task description, to experiment how it behaves, specifically in cases where the application breaks otherwise.  An automated use makes sense once we see that this is stable and has no side effects.

src/radical/pilot/agent/bootstrap_0.sh
andre-merzky: I realized that this wasn't correct anyway: an empty value does not mean that the var did not exist.  This is thus fixed differently now - can you have a look?

andre-merzky:
Yeah, I am also not happy, yet.  Thanks for checking.

src/radical/pilot/agent/bootstrap_0.sh
```diff
@@ -157,12 +157,14 @@ create_deactivate()
     # unset the variable
     for var in $changed
     do
-        val=$(grep -e "^export $var=" env.orig | cut -f 2- -d '=')
-        if test -z "$val"
+        orig=$(grep -e "^export $var=")
```
mtitov, line 160: string was overcut.. missed the variable `env.org` -> `orig=$(grep -e "^export $var=" env.orig)`

andre-merzky: Yeah, saw that too - I pushed before the test run completed... *blush*

mtitov:
@andre-merzky ack the code correctness

BTW as a follow up question: env variable `$RP_PILOT_SANDBOX` is set/managed somewhere? cannot check it..

andre-merzky:
`RP_PILOT_SANDBOX`: yes, it is added [here](https://github.com/radical-cybertools/radical.pilot/pull/2127/files#diff-a629ccc4c2ef78f37f7e121382ce2677R213), which adds it to the tasks execution shell script,

PR #2133, opened 2020-04-27
Updates in IBRun LM (module itself + tests)

mtitov:


src/radical/pilot/agent/launch_method/ibrun.py
```diff
@@ -41,38 +39,52 @@ def construct_command(self, cu, launch_script_hop):
...
         slots        = cu['slots']
         cud          = cu['description']
+
         task_exec    = cud['executable']
-        task_cores   = cud['cpu_processes']  # FIXME: handle cpu_threads
         task_args    = cud.get('arguments') or []
         task_argstr  = self._create_arg_string(task_args)
+        task_env     = cud.get('environment') or dict()
+
+        n_tasks      = cud['cpu_processes']  # FIXME: handle cpu_threads
```
andre-merzky, line 48: We may want to move the task to `FAILED` then if threads are not handled yet.  An
```python
assert(cud.get('cpu_threads', 1) == 1), 'no thread support in IBRUN launch method'
```
would do the trick - the executor will catch the except and set the task state etc.  (Or a `ValueError` I guess...).  Or maybe I misunderstand the `FIXME` and the task will still run as expected?

src/radical/pilot/agent/launch_method/ibrun.py
```diff
@@ -41,38 +39,52 @@ def construct_command(self, cu, launch_script_hop):
...
-        offsets = list()
+        # Usage of env variable TACC_TASKS_PER_NODE is purely for MPI tasks,
+        #  and threads are not considered (info provided by TACC support)
+        n_node_tasks = int(task_env.get('TACC_TASKS_PER_NODE') or
+                           self._cfg.get('cores_per_node', 1))
+
+        if not self._node_list:
+            raise RuntimeError('attribute "_node_list" (from RM) is not set')
+        if not slots.get('nodes'):
+            raise RuntimeError('unit.slots.nodes is not set')
```
andre-merzky, line 58: I'm fine with the check, but I wonder if this case can legally happen.   The `NOOP` scheduler should not set `slots` in the first place, but then this test would result in an `AttributeError`.  All other schedulers should set `nodes` if they are able to place the task - and if not, the task would not arrive here.  Am I missing a case?

You test for `core_map` in the slots in the loop below, and I have the same question there.  Can that legally happen?  Of not, maybe use an assert so that we can optimize it away.  

The `NOOP` scheduler may not assign slots - in that case, IBRUN should still be able to run it, but we would leave placement to the batch systemin that case, and not specify `offsets` at all.  I don't think we need to cover it in this PR, but I wanted to mention it - maybe add a `FIXME`?

src/radical/pilot/agent/launch_method/ibrun.py
```diff
@@ -41,38 +39,52 @@ def construct_command(self, cu, launch_script_hop):
...
-        cpn     = self._cfg.cores_per_node
-        index   = 0
-        offsets = list()
+        # Usage of env variable TACC_TASKS_PER_NODE is purely for MPI tasks,
+        #  and threads are not considered (info provided by TACC support)
+        n_node_tasks = int(task_env.get('TACC_TASKS_PER_NODE') or
+                           self._cfg.get('cores_per_node', 1))
+
+        if not self._node_list:
+            raise RuntimeError('attribute "_node_list" (from RM) is not set')
```
andre-merzky, line 56: We check the validity of the node list during the [RM initialization](https://github.com/radical-cybertools/radical.pilot/blob/devel/src/radical/pilot/agent/resource_manager/base.py#L116), and thus this line is really just checking if the constructor gets that node list from the configuration.  Should we move that test into the constructor, if you think we should have it?

andre-merzky:
(Quick note: no need to set the `layer=RP` flag on tickets in this repo - that is meant for tickets which need fixing in another layer than where they were opened, and then we usually reference a ticket in that other repo to track)

andre-merzky:
@iparask : the tests don't seem to kick in again - any idea what's up?

src/radical/pilot/agent/launch_method/ibrun.py
mtitov: That `FIXME` was there already, I thought just to keep it for later (should I remove it?). With `cpu_threads` would be better to just ignore it (e.g., not to change the app code after switching between LMs)

src/radical/pilot/agent/launch_method/ibrun.py
mtitov: I like idea with `assert`, will update accordingly

In general all that checks are of general purpose - if we use something then need to check that it has valid values, yeah, I didn't keep in mind the origins (didn't consider schedulers), I'll re-check it tho.

And now I think that with nodes-related parameters I did cut the possibility it could be just empty lists (?) Should I update it as:
`if self._node_list is not None` and `if slots.get('nodes') is not None` ?

src/radical/pilot/agent/launch_method/ibrun.py
mtitov: yes, totally agree, thank you! It is redundant to have this validity check here (will update this as well)

src/radical/pilot/agent/launch_method/ibrun.py
andre-merzky: Great, thanks.  In general I think we should use asserts more frequently to assure our assumptions, I have been slowly introducing them in some code paths...

src/radical/pilot/agent/launch_method/ibrun.py
andre-merzky: Ah, makes sense.  Can you clarify what happens in the current code when a task requests, say, 2 threads per process?  I assume ibrun will ignore that, but will also not pin the process to specific cores so that the process can indeed then create those processes?  If so, the `FIXME` can indeed be removed.

mtitov: Yeah, can confirm that threads will be ignored, and no process pinning to specific cores (offsets are at process/task level and threads are not considered)

iparask:
No, I do not know. Let me bring them back and make sure that all PR trigger the tests.

Issue #2126, opened 2020-04-20
RP testing [LM]


mtitov:
Hi everyone,

having this ticket for discussions on updates in [RP testing for agent's LM](https://github.com/radical-cybertools/radical.pilot/wiki/RP-testing:-proposal-on-structure-reorganization)

everyone is welcome to make changes in proposal or bring up related topics to discuss

iparask:
Thank yo @mtitov ! I'll look at it and let you know.

iparask:
I agree with @mtitov proposal on organizing the test cases and I would like to second it.

mtitov:
@iparask thank you! This update also should follow your proposed guideline, which brings really good points (some of that is missed at LM tests)

andre-merzky:
This looks good to me.  I am a bit worried that we spend too much time organizing tests instead of writing them and getting them to work reliably - but not much :-)

iparask:
My proposal about test coding guidelines are [here](https://github.com/radical-cybertools/radical.pilot/wiki/Tests-Coding-Guidelines). @andre-merzky we can enforce the new style with new tests, and not old. The same procedure with linting. Any opinions?

andre-merzky:
I don't think the guidelines are enforceable the same way we enforce linting, but otherwise I agree.

mtitov:
Corresponding reference was added to [Tests Coding Guidelines](https://github.com/radical-cybertools/radical.pilot/wiki/Tests-Coding-Guidelines)

PR #2134, opened 2020-04-28
Agent stage_in test1

iparask:
This PR includes a test for the agent's stage in component. It also show how a unit test looks like based on the guidelines introduced in #2126.

Thank you

### DIFF
--- a/src/radical/pilot/agent/staging_input/default.py
+++ b/src/radical/pilot/agent/staging_input/default.py
@@ -77,6 +77,10 @@ class Default(AgentStagingInputComponent):
             else:
                 no_staging_units.append(unit)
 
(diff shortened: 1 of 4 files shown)


mtitov:
@iparask, have general comments yet (about test modules organization):
- Should we have methods `setUp` and `tearDown` in a separate module as part of `BaseTest` class (that would be inherited by actual test classes)? (even we have just one module `default` here, that would be a one approach across other components). And then each test class can redefine these methods, but they still should be called inside with `super` func
- Should we test public methods at the first place? test `work` instead of `_work`. I mean both works, but the primary focus better be on public methods

iparask:
Closing to check travis. I will open a new one including @mtitov comments

PR #2135, opened 2020-04-28
Agent stage_in test1

iparask:
This PR includes a test for the agent's stage in component. It also show how a unit test looks like based on the guidelines introduced in #2126.

Thank you
Giannis

From @mtitov:

I have general comments yet (about test modules organization):

Should we have methods setUp and tearDown in a separate module as part of BaseTest class (that would be inherited by actual test classes)? (even we have just one module default here, that would be a one approach across other components). And then each test class can redefine these methods, but they still should be called inside with super func
Should we test public methods at the first place? test work instead of _work. I mean both works, but the primary focus better be on public methods

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 
 language: python
 python:   "3.6"
-root:     disabled
 os:       linux
 dist:     xenial
 
--- a/src/radical/pilot/agent/staging_input/default.py
+++ b/src/radical/pilot/agent/staging_input/default.py
@@ -77,6 +77,10 @@ class Default(AgentStagingInputComponent):
             else:
                 no_staging_units.append(unit)
 
+        unit_testing = os.environ.get('RP_UNIT_TESTING', 'false')
+
+        if unit_testing.lower() == 'true':
+            return staging_units, no_staging_units
 
         if no_staging_units:
             self.advance(no_staging_units, rps.AGENT_SCHEDULING_PENDING,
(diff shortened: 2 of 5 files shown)


iparask:
The build of this PR exists: https://travis-ci.org/github/radical-cybertools/radical.pilot/builds/680674671

For some reason Github does not display it.

andre-merzky:
> Should we have methods setUp and tearDown in a separate module as part of BaseTest class (that would be inherited by actual test classes)? (even we have just one module default here, that would be a one approach across other components). And then each test class can redefine these methods, but they still should be called inside with super func

Again, I want to caution on over-engineering.  If that approach simplifies something, or makes our live easier in some way, then by all means.  Otherwise I'd suggest to wait until we have some component tests implemented to see if this pattern holds.

> Should we test public methods at the first place? test work instead of _work. I mean both works, but the primary focus better be on public methods

Why does it matter if it is private or public?

mtitov:
@andre-merzky 
1) base class with `setUp` and `tearDown` would have procedures that are common for all other tests (e.g., read test cases, remove files like `dummy` for logs, etc.); and each class later could extend it with set up of specific environment and clear it out later at the extended `tearDown`;
2) that's because of purpose of these methods, private (and as in our case - protected) methods are intended to be called inside the class, but public methods that what is provided for "outside", thus tests should cover the correctness of public methods (and that suppose to cover "internal" methods as well)

andre-merzky:
base class for tests: works for me

private / public: I kind of see your point - but our private methods need correctness and tests just as much.  I am not sure if it makes much of a difference.  Ultimately the private ones will be called by the public ones anyway (who else *could* call them?).  So they may appear as covered, but won't have their own unit tests.

But I really don't have a strong opinion to be honest, I should be careful not to bikeshed that discussion :-P  So either way works for me.

mtitov:
yeah, my point was not to put any methods aside :) but just to focus on public methods firstly and then everything else

PR #2144, opened 2020-05-11
Travis stops checking examples and .coverage fix

iparask:
In order to see the coverage changes we need to include the tested folder in the coveragerc script.

I just added those already covered. This will also allow us to see the changes in coverage from PRs


andre-merzky:
Hmm, locally I am running something like `pytest --cov=radical.pilot --cov-report=term-missing tests` and then don't need to specify any folders - coverare is reported for everything in the given namespace.  Would that here also work?  I am worried we (a) need to maintain yet another list, and (b) it breaks with 3.7 ...

iparask:
I can generalize it and check the whole namespace. I do not remember right now why we had it like that. 

> (b) it breaks with 3.7 ...

Why does it break with 3.7?

andre-merzky:
> Why does it break with 3.7?

Because 3.6 is hardcoded in the paths

iparask:
I fixed that. If we take out `.coveragerc` it will only report what ever is covered. Now, it should report for all files. I move the dependency on travis' rc file which needs to be maintain correctly.

iparask:
I commented out the examples until Aymen and Aydin finish their work. Travis was failing because of that linting.

andre-merzky:
I think `pytest --cov=radical.pilot` covers all code which is imported in that module - which is what we really want I guess.  Like, we don't care about code which is in the `archive/` subdirs, or any other files which are not imported.

iparask:
I'm not say no to that, but through codecov we will get the coverage badge and also a PR comment with the coverage change of a PR. I think those two are important.

Issue #2145, opened 2020-05-12
feature request: pilot job name


andre-merzky:
From Wenjie:

Hi Can I change the name of the lsf job showing in the queue?

Say I have two jobs in the queue:

```
JOBID   USER       STAT   SLOTS    QUEUE       START_TIME    FINISH_TIME   JOB_NAME                      
91487   lei        RUN    211      batch       May 11 23:45  May 12 00:45  pilot.0000                    
91467   lei        PEND      -     batch             -             -       pilot.0000
```
They both have the name pilot.0000 and I can't differentiate between them...

PR #2146, opened 2020-05-12
Feature/job name

andre-merzky:
This implements #2145


andre-merzky:
About the linters in `src/radical/pilot/pmgr/launching/default.py`

`  18   0    E0402   Attempted relative import beyond top-level package (relative-beyond-top-level)`
This is wrong, the import is correct

`  672  26   W0212   Access to a protected member _get_session_sandbox of a client class (protected-access)
  941  27   W0212   Access to a protected member _get_resource_sandbox of a client class (protected-access)
  942  27   W0212   Access to a protected member _get_session_sandbox of a client class (protected-access)
  943  27   W0212   Access to a protected member _get_pilot_sandbox of a client class (protected-access)
  944  27   W0212   Access to a protected member _get_client_sandbox of a client class (protected-access)`

They are supposed to be protected members, as they are not part of the public session API.   At the same time, the launcher needs that information.  Python has no notion of a `friend` class or similar, so there is no reasonable way to fix this.  The unreasonable way is to split the Session in a public and private part, which is not something I would favor just to fix these warnings :-/

The example and test linters were supposed to be taken out of the Travis test, IIRC?  We have an ongoing effort to fix both, they won't get done in this PR.

iparask:
> What would be a reasonable test for it?

A meaningful test would be that for different configurations of  resources, rcfgs and pilots to check whether or not the SAGA job description is correct as well as the rest of the thing it configures in that `ret` variable. Would you agree with that?


> About the linters in `src/radical/pilot/pmgr/launching/default.py`
> 
> ```
>   18   0    E0402   Attempted relative import beyond top-level package (relative-beyond-top-level)
> ```
> 
> This is wrong, the import is correct

I think it complains because there is no `__init__.py` file under the `radical` folder, where the pilot folder exists.

> 
> ```
>   941  27   W0212   Access to a protected member _get_resource_sandbox of a client class (protected-access)
>   942  27   W0212   Access to a protected member _get_session_sandbox of a client class (protected-access)
>   943  27   W0212   Access to a protected member _get_pilot_sandbox of a client class (protected-access)
>   944  27   W0212   Access to a protected member _get_client_sandbox of a client class (protected-access)
> ```
> 
> They are supposed to be protected members, as they are not part of the public session API. At the same time, the launcher needs that information. Python has no notion of a `friend` class or similar, so there is no reasonable way to fix this. The unreasonable way is to split the Session in a public and private part, which is not something I would favor just to fix these warnings :-/
> 

Since private variables are being used then it might make sense to silence pylint with `# pylint: disable=protected-access` before those lines.

> The example and test linters were supposed to be taken out of the Travis test, IIRC? We have an ongoing effort to fix both, they won't get done in this PR.

 PR #2144 should take care of that.

andre-merzky:
Uh,  this PR adds a `name` property to the description, and any string (and no string) is correct for SAGA.  And the string type is enforced by the description schema (that schema str specs are respected is tested by RU).  I don't really want to add a test for the SAGA JD creation right now....  I'll think of something...

> I think it complains because there is no __init__.py file under the radical folder, where the pilot folder exists

That's true, but then again, that's how namespaces work in Python 3.7+..

> Since private variables are being used then it might make sense to silence pylint with # pylint: disable=protected-access before those lines.

```
$ grep -re self._session._ src/ | wc -l
42
```

And there are several more patterns than 'self._session` I'm afraid... :-/  I kind of understand that test for private access makes sense - but friend access is simply not a language feature, so I am not sure how useful the check is...

> PR #2144 should take care of that.

Ack, and thanks!

iparask:
> > I think it complains because there is no **init**.py file under the radical folder, where the pilot folder exists
> 
> That's true, but then again, that's how namespaces work in Python 3.7+.

I did a quick search and found that the package is used on lines [469](https://github.com/radical-cybertools/radical.pilot/blob/feature/job_name/src/radical/pilot/pmgr/launching/default.py#L469) and [510](https://github.com/radical-cybertools/radical.pilot/blob/feature/job_name/src/radical/pilot/pmgr/launching/default.py#L510) to perform state checks, instead of using the `states` which are imported also in [line 19](https://github.com/radical-cybertools/radical.pilot/blob/feature/job_name/src/radical/pilot/pmgr/launching/default.py#L19).

As a result, the whole package is not needed. In my personal opinion cyclic imports are dangerous and should be avoided. In this case is easily avoidable.

mtitov:
hi @andre-merzky, I agree with @iparask about update `rp.` -> `rps.` (and about general approach).
will you make that changes? if there is no objections (or either me or Ioannis can do that)
p.s. need to push this PR forward

andre-merzky:
Oops - didn't meant to close.  Yes, I'll do that change.  Thanks.

PR #2153, opened 2020-05-25
Provide Docker recipe and instructions for testing.

eirrgang:
Provide a recipe and instructions for a self-contained testing environment. Based on the [mongo](https://hub.docker.com/_/mongo) image from Docker Hub (see also [Dockerfile](https://github.com/docker-library/mongo/blob/master/4.2/Dockerfile)), install `radical.pilot` in an image that will automatically configure and launch the required MongoDB instance.

A few seconds after the container is launched, the db is available to additional processes launched in the container.

**TO DO**
- [X] The database created by this Dockerfile is not yet sufficient for RP to be able to create Sessions.
- [x] Blocked by #2159
- [ ] The `examples/misc/task_overlay.py` script does not yet run in the resulting image. *see #2218 *


docker/Dockerfile
```diff
@@ -0,0 +1,80 @@
...
+        'radical.utils>=1.1'
+
+COPY --chown=rp:radical . /radical.pilot
+
+RUN (. ~rp/rp-venv/bin/activate && \
+    cd /radical.pilot && \
+    ~rp/rp-venv/bin/pip install .)
+
+# radical.saga logs `ModuleNotFountError`s if `redis` is not explicitly installed.
+RUN /home/rp/rp-venv/bin/python -m pip install redis
```
eirrgang, line 66: The exception appears in the log file, but I don't know whether it is a fatal error. If it is, then either the radical.sage package requirements should be updated, or a more prominent "feature unavailable" sort of error should be contrived.

eirrgang:
I thought the remaining issues might be related to the documented requirement of non-interactive ssh connections, so I made the "radical" container into a member of a service cluster and set up non-interactive ssh access to localhost for the "rp" user, but the `task_overlay.py` example still terminates with a SIGINT and does not produce the intended results.

docker/Dockerfile
andre-merzky: This has recently been addressed via radical-cybertools/radical.saga/pull/783

eirrgang:
Possibly, one thing missing would be a resource configuration specific to the Docker container. The resource definition inherited for "local.localhost" describes an environment that is not likely to be a good match for the docker configuration here. I don't know if this is what is keeping the current tests from starting the expected workers and agents, but it is worth looking in to.

Is there a standard place and name for a JSON file to override elements of the "local.localhost" resource description?

eirrgang:
Update: Troubleshooting with @andre-merzky established that the `docker-compose` method is not necessary at this point since the `local.localhost` resource can launch agents through local subprocess execution (it does not require non-interactive ssh), and the example (at least in this version) only requests a single node, single agent, and single master.

The hang that I have been experiencing can be avoided if the `master.wait(count=2)` is omitted from the task_overlay_master.py example script. @andre-merzky is investigating, but it appears that the Docker configuration in this branch has been confirmed to be sufficient for testing.

andre-merzky:
It failed for me, and I did not find time to debug further.  I know that some errors come from missing MPI installation, but those should not be fatal (just to spare you that specific rabbit hole in case you look into it).

mturilli:
Great, let me give it a go then.

eirrgang:
> I read the code and it looks good to me. Did you test it? If not, happy to give it a go before approving the PR.

As noted above, the task_overlay example does not work yet because of the unidentified bug in the `wait()` in the master script that @andre-merzky was looking into.

> It failed for me

Can you clarify or attach the terminal output?

I will merge `devel` in and try again. It has been a while since I checked both recipes.

eirrgang:
It looks like the `docker exec` to connect to the login container of the `docker-compose` version is not quite the way to go. Either I need to assign or discover its container name, or I need to just connect to it in a more conventional way. I assumed I would eventually use the no-password ssh login, but it looks like I should set that up sooner than later to have the services version in this PR. Alternatively, since the monolithic container seems to work fine, I can just remove that option for now and reintroduce it when it is necessary and complete.

eirrgang:
This PR has spanned a radical.pilot release and should probably be closed.

Should I open a new PR or is this not something that is wanted?

I don't know if the task_overlay problem was ever tracked. Is there something that should be cited or is there some other way to follow up on that?

andre-merzky:
We usually just merge current devel into the PR branch and use the PR, no matter the age of the PR.  Would that work for you?

For the task overlay: that still sees heavy development and is not considered stable, so I would suggest to leave that as is and to revisit the problems once the code stabilizes.

eirrgang:
> I would suggest to leave that as is and to revisit the problems once the code stabilizes.

will do

eirrgang:
> > I would suggest to leave that as is and to revisit the problems once the code stabilizes.
> 
> will do

Done. Docker recipe and Travis-CI config have been updated.

andre-merzky (approved):
This PR should wait for #2218 to get resolved, but can otherwise be merged.

.travis.yml
```diff
@@ -23,6 +23,27 @@ env:
...
+#      - echo TRAVIS_BRANCH is $TRAVIS_BRANCH
+#      - RPREF=$TRAVIS_BRANCH
+#      - (cd docker ; docker build -t rp-complete -f rp-complete.dockerfile --build-arg RPREF=${RPREF:-devel} .)
+    script:
+      - docker run --rm --name rp_test -d rp-complete
+      - docker ps
+      - sleep 5
+      # Consider testing more examples, based on problem areas.
+      - docker exec -ti -u rp rp_test bash -c "cd && . /home/rp/rp-venv/bin/activate && python radical.pilot/examples/00*"
+# Ref issue #2218
```
eirrgang, line 43: - [ ] @andre-merzky requests that we confirm resolution of 2218 before merging.

AymenFJA:
some updates here: #2218

PR #2154, opened 2020-05-26
Reconcile Session.close() docstring

eirrgang:
The default behavior of Session.close() changed in c8ea920178884b3394518ff0402b8b9dafdbe39d but the docstring was not updated.


mturilli:
Thank you for catching it!

PR #2156, opened 2020-05-28
Coverage report is coming

iparask:
This PR checks if travis actually creates a coverage report.


iparask:
@andre-merzky at last we have the bot. Things do not look promising, but at least we have the required values. If you agree I would like to merge this with devel.

PR #2165, opened 2020-06-14
Jobs handling at Lassen

mtitov:


src/radical/pilot/agent/launch_method/jsrun.py
```diff
@@ -84,7 +97,12 @@ def _create_resource_set_file(self, slots, uid, sandbox):
 
         """
 
-        rs_str = 'cpu_index_using: physical\n'
+        # if `cpu_index_using: physical` is set to run at Lassen@LLNL,
+        #  then it returns an error "error in ptssup_mkcltsock_afunix()"
+        if slots['nodes'][0]['name'].lower().startswith('lassen'):
```
andre-merzky, line 102: Should we check for the `lm_info['cvd_id_mode']` here instead of inspecting the hostname again, or am I missing something?

mtitov: that two methods are called at different times, that's why these changes are split...

mtitov:
I absolutely agree to move parameter `cvd_id_mode` into config (and looks like the value `logical` is applied to Summit only)

PR #2172, opened 2020-06-29
Examples fail without gpus key

mturilli:
Specify gpus number in examples' config file for ornl.summit*. Without it, examples fail.


andre-merzky:
Keeping this PR open for #2192

iparask:
What is the proposed change here? I do not see a diff at all.

PR #2175, opened 2020-07-01
Hotfix/lfs support

iparask:
@andre-merzky I did the changes as we discussed. This is the output from the debug log entries in the resource manager:
```
rp.session.js-17-185.jetstream-cloud.org.iparask.018444.0000.log:1593562103.267 : rp.session.js-17-185.jetstream-cloud.org.iparask.018444.0000 : 1193  : 140626738517824 : INFO     : radical.pilot version: 1.4.1-v1.4.1@hotfix-lfs_support
rp.session.js-17-185.jetstream-cloud.org.iparask.018444.0000.log:1593562103.933 : rp.session.js-17-185.jetstream-cloud.org.iparask.018444.0000 : 1193  : 140626738517824 : DEBUG    : lfs path environment value None
rp.session.js-17-185.jetstream-cloud.org.iparask.018444.0000.log:1593562103.933 : rp.session.js-17-185.jetstream-cloud.org.iparask.018444.0000 : 1193  : 140626738517824 : DEBUG    : RU lfs path None
```


src/radical/pilot/agent/resource_manager/slurm.py
```diff
@@ -78,7 +78,8 @@ def _configure(self):
         self.cores_per_node = self._cfg.get('cores_per_node', 0)
         self.gpus_per_node  = self._cfg.get('gpus_per_node',  0)  # FIXME GPU
         self.mem_per_node   = self._cfg.get('mem_per_node',   0)
-
+        self._log.debug('lfs path environment value %s', os.environ.get('$LOCAL', None))
```
mtitov, line 81: @iparask shouldn't it be `os.environ.get('LOCAL')`? (without `$`, which is not part of the name)

iparask:
Closing due because the fix should go on devel.

src/radical/pilot/agent/resource_manager/slurm.py
iparask: Fixed

iparask:
I resolve the bug I had in the debug messages. This is what I am getting now:
```
rp.session.js-17-185.jetstream-cloud.org.iparask.018449.0000.log:1594043083.610 : rp.session.js-17-185.jetstream-cloud.org.iparask.018449.0000 : 8207  : 140114699933504 : DEBUG    : lfs path environment value /local/9989050
rp.session.js-17-185.jetstream-cloud.org.iparask.018449.0000.log:1594043083.610 : rp.session.js-17-185.jetstream-cloud.org.iparask.018449.0000 : 8207  : 140114699933504 : DEBUG    : RU lfs path None
```

andre-merzky:
I checked the RU layer, and the env expansion seems to work as exepcted:

```sh
$ export FOO=BAR
$ python3 ./test.py 
BAR

$ unset FOO
$ python3 ./test.py 
None
```

The code is:
```python
#!/usr/bin/env python3

import radical.utils as ru

cfg = ru.Config(cfg={'foo': '${FOO}'})
print(ru.expand_env(cfg.foo))
```

andre-merzky:
@iparask : I added anotherr debug message - can you givce this a try in your setup?  Thanks!

iparask:
Hello @andre-merzky, sorry it took me this much. Here is the output:
```
rp.session.js-17-185.jetstream-cloud.org.iparask.018463.0001.log:1595255765.705 : rp.session.js-17-185.jetstream-cloud.org.iparask.018463.0001 : 9153  : 140509867358016 : DEBUG    : env lfs path /local/10177443
rp.session.js-17-185.jetstream-cloud.org.iparask.018463.0001.log:1595255765.705 : rp.session.js-17-185.jetstream-cloud.org.iparask.018463.0001 : 9153  : 140509867358016 : DEBUG    : cfg lfs path None
rp.session.js-17-185.jetstream-cloud.org.iparask.018463.0001.log:1595255765.705 : rp.session.js-17-185.jetstream-cloud.org.iparask.018463.0001 : 9153  : 140509867358016 : DEBUG    : exp lfs path None
```

andre-merzky:
So the problem is not the env expansion, but that `self._cfg.get('lfs_path_per_node'))` results in `None`.  Can you do a dump of `self._cfg` to see what's up?

iparask:
I just looked at `agent.0.cfg` and I see the following:
```
    "gpus_per_node": 2,
    "heartbeat": {
        "interval": 1.0,
        "timeout": 60.0
    },
    "lfs_path_per_node": null,
    "lfs_size_per_node": 3713368,
    "logdir": ".",
    "mode": "shared",
    "mpi_launch_method": "MPIRUN",
    "owner": "agent.0",
    "pid": "pilot.0000",
    "pilot_sandbox": "/pylon5/mc3bggp/paraskev/radical.pilot.sandbox/rp.session.js-17-185.jetstream-cloud.org.iparask.018466.0002/pilot.0000/",
```

Which means that the name is not passed correctly when the agent.0.cfg is constructed.

andre-merzky:
Can you give me a pointer to the client and pilot sandbox by any chance?  I am confused how that could happen...

iparask:
I think it is ready

src/radical/pilot/agent/scheduler/base.py
```diff
@@ -840,6 +840,10 @@ def _try_allocation(self, unit):
         self._change_slot_states(slots, rpc.BUSY)
         unit['slots'] = slots
 
+        env_dict = unit['description'].get('environment', {})
+        env_dict['NODE_LFS_PATH'] = slots['lfs_per_node']['path']
```
andre-merzky, line 844: We should prefix all variables we set with `RP_` - at least that is the scheme we do elsewhere, in order to minimize chances to collide with user or system environment.  Maybe `RP_LSF_PATH`?  (`node` and `local` (from LSF) are redundant anyway).

src/radical/pilot/agent/scheduler/base.py
```diff
@@ -840,6 +840,10 @@ def _try_allocation(self, unit):
         self._change_slot_states(slots, rpc.BUSY)
         unit['slots'] = slots
 
+        env_dict = unit['description'].get('environment', {})
```
andre-merzky, line 843: The task has always a description at this point, the UMGR should make sure of that.  If not, then this needs fixing.  So you can simplify the code to just setting the env).

src/radical/pilot/session.py
```diff
@@ -652,7 +652,7 @@ def get_resource_config(self, resource, schema=None):
             raise RuntimeError("Resource host '%s' unknown." % host)
 
         resource_cfg = copy.deepcopy(self._rcfgs[domain][host])
-
+        self._log.debug('base rcfg:', self._rcfgs[domain][host])
```
andre-merzky, line 655: Can we take this out again?

src/radical/pilot/agent/scheduler/base.py
iparask: I would agree, let's open a ticket for it and introduce backward compatibility for a bit. Then we can change it to.

src/radical/pilot/agent/scheduler/base.py
iparask: The unit does not always have an environment. I put this line to make sure that if there is an environment in the description to take it and if not to get an empty dictionary.

Do you think this is wrong?

src/radical/pilot/session.py
iparask: Yes

src/radical/pilot/agent/scheduler/base.py
andre-merzky: The default values will always be set, and that is an empty dictionary for the `environment` key:

```python
>>> import radical.pilot as rp
>>> cud = rp.ComputeUnitDescription({'executable': '/bin/true'})
>>> cud.environment
{}
```
We thus never need to check if a key was set by the application.

src/radical/pilot/agent/scheduler/base.py
andre-merzky: Makes sense - will you take care of the issue creation, please?

PR #2176, opened 2020-07-02
Hotfix/mpiexec hosts

andre-merzky:
This popped up on Bridges when using the default intel MPI module.  While the old version worked with OpenMPI's `mpiexec`, it is not what's prescribed for `mpiexec` arguments in the standard and thus was not portable.  Using `-host` and `-n` should fix this.


andre-merzky:
Oh, thanks for taking care of the tests!

Issue #1557, opened 2018-02-07
`SKIP_FAILED` is defined, but not implemented


andre-merzky:
it should go, or at least not be set by default.

vivek-bala:
I remember this old issue #1106. The idea behind it was to ignore data staging that fails and continue with unit execution. But the flag created for this was 'NONFATAL'. Maybe there is some overlap between the two? I am not sure what 'SKIP_FAILED' is meant to do.

andre-merzky:
Yeah, its likely related to that ticket - but as said, I don't see that flag being evaluated anywhere right now.  Is this still a valid use case?

vivek-bala:
I haven't use 'skip_failed', so not sure. I have used 'nonfatal' in the past and consider it useful, so would recommend keeping 'nonfatal'.

andre-merzky:
I don't see that flag evaluated anywhere either - I doubt it works... :/

iparask:
@andre-merzky, @vivek-bala Is this use case still valid?

vivek-bala:
I don't think so, no.

iparask:
I'm closing as not needed any more.

andre-merzky:
If it is not needed, the flag and its documentation should be removed...

PR #2178, opened 2020-07-07
Remove SKIP_FAILED

mturilli:
This should address #1557


andre-merzky:
@iparask :  Aaand here we have the example where the removal of a covered code line decreases test coverage ;-)

PR #2179, opened 2020-07-07
Add needed documentation

mturilli:
@andre-merzky , this assumes the availability of a command `radical-pilot-resources` or similar that returns the list of available resource tags alongside the resource FQDN. We used to have such a command but I did not find it.


andre-merzky:
I added that tool - will you have a look, please?

PR #2182, opened 2020-07-08
Specify the real memory required per node

mtitov:


mtitov:
@andre-merzky have couple question regarding this task:
- should we have an attribute name in pilot description as `memory` or `memory_per_node`? (second one is more explicit)
- currently SAGA layer allows to specify this attribute as `int`, should we change it to `str`? (so to allow such example as `1G`)

andre-merzky:
> * should we have an attribute name in pilot description as `memory` or `memory_per_node`? (second one is more explicit)

Yes, I think we should have `memory` as a requirement.  We don't have a notion of `nodes` in RP (yet), so I would refrain from using `memory_per_node` right now.

> * currently SAGA layer allows to specify this attribute as `int`, should we change it to `str`? (so to allow such example as `1G`)

For SAGA, we don't have that freedom - the API spec defines `TotalPhysicalMemory` as float (unit Megabytes), so that's what we do.  We usually try not to deviate from the spec unless the API is broken.

src/radical/pilot/pmgr/launching/default.py
```diff
@@ -884,6 +884,7 @@ def _prepare_pilot(self, resource, rcfg, pilot, expand):
         # pilot description and resource configuration
         number_cores    = pilot['description']['cores']
         number_gpus     = pilot['description']['gpus']
+        required_memory = pilot['description']['memory']  # per node
```
andre-merzky, line 887: I think this should not be per node, but total memory, to be in sync with the other attributes.  Also, the user is not even supposed to know in how many nodes the given number of cores and GPUs results.

src/radical/pilot/pmgr/launching/default.py
```diff
@@ -1281,6 +1282,7 @@ def _prepare_pilot(self, resource, rcfg, pilot, expand):
         jd.error                 = "bootstrap_0.err"
         jd.total_cpu_count       = number_cores
         jd.total_gpu_count       = number_gpus
+        jd.total_physical_memory = required_memory
```
andre-merzky, line 1285: SAGA's `total_physical_memory` on the other hand is supposed to define the global job memory.  If the above is changed to global memory too, then this should work - right now it should result in incorrect memory specs on the SAGA layer I think.

src/radical/pilot/pmgr/launching/default.py
mtitov: Thus we assume that user set the total amount memory? which later we suppose to divide by number of nodes? (that comment I've put for us, since later this parameter translates into `--mem`, which is "the real memory required per node")

src/radical/pilot/pmgr/launching/default.py
mtitov: In SAGA `total_physical_memory` translates into SLURM parameter `--mem` ([slurm](https://slurm.schedmd.com/sbatch.html)), which is real memory required per node

andre-merzky: Thanks for noting that - but then the SAGA Slurm adaptor needs fixing (see https://www.ogf.org/documents/GFD.90.pdf page 179).

mtitov: Thus we just set a total memory here (no more changes in RP, except removing that comment "per node", right?) and as the next step is to fix SAGA (translates total memory into memory per node for SLURM)?

andre-merzky: Yes, correct!  Please add a small comment on the new dscription key also, otherwise that looks good.

mtitov:
@andre-merzky it doesn't let to merge because of `pylint` - this is similar to PR #2146, is there some resolution with it?
(adding @iparask)

andre-merzky:
problem solved :-)  Thanks!

PR #2187, opened 2020-07-17
Add Profiler instance into Publisher and Subscriber (zmq.pubsub)

mtitov:


mtitov:
More comments (from the ticket): issue comes from not defined instance of `Profiler` for `Publisher` and `Subscriber` (which are from `radical.utils.zmq.pubsub`, e.g. [Publisher](https://github.com/radical-cybertools/radical.utils/blob/devel/src/radical/utils/zmq/pubsub.py#L194)). If `Profiler` instance is not provided then a new one is created - if path for the corresponding file is not set then it is taken from `radical.utils.config.DefaultConfig` (`'profile_dir': '${RADICAL_DEFAULT_PROFILE_DIR:$PWD}'`)

As an example, for the case of `radical-pilot-bridge`, the path for a new instance of `Profiler` is set as the current working directory (directory from where user's script is launched; from the config as `$PWD`), and corresponding files of `heartbeat` were created at the work dir.

PR #2194, opened 2020-07-29
No display ID/PASS from RADICAL_PILOT_DBURL

lee212:
It still prints hostname and db name in screen and a log file.


lee212:
This is related to EnTK: https://github.com/radical-cybertools/radical.entk/issues/470

PR #2197, opened 2020-08-03
Debug pilot/unit duration

mturilli:
Contains the set of debug duration for both pilot and unit entities. Happy to discuss the naming of each duration before merging. Names will be used in RA and will likely become those used in our discussions. Thus the need to be all relatively happy with them,.


src/radical/pilot/utils/prof_utils.py
```diff
@@ -199,6 +199,114 @@
...
+                                      {ru.EVENT: 've_activate_start', ru.STATE: None                       }  ], 
+        'p_agent_ve_activate'     : [ {ru.EVENT: 've_activate_start', ru.STATE: None                       },
+                                      {ru.EVENT: 've_activate_stop' , ru.STATE: None                       }  ],
+        'p_agent_install_init'    : [ {ru.EVENT: 've_activate_stop' , ru.STATE: None                       },
+                                      {ru.EVENT: 'rp_install_start' , ru.STATE: None                       }  ],
+        'p_agent_install'         : [ {ru.EVENT: 'rp_install_start' , ru.STATE: None                       },
+                                      {ru.EVENT: 'rp_install_stop'  , ru.STATE: None                       } ],
+        'p_agent_launching'       : [ {ru.EVENT: 'rp_install_stop'  , ru.STATE: None                       },
+                                      {ru.EVENT: 'state'            , ru.STATE: rps.PMGR_ACTIVE            }  ],
+        'p_agent_runtime'         : [ {ru.EVENT: 'state'            , ru.STATE: rps.PMGR_ACTIVE            },     # total pilot runtime
```
andre-merzky, line 238: We can add the timestamp where the pilot receives it's termination command - specifically for EnTK this may be relevant as that command arrives rather late sometimes.

src/radical/pilot/utils/prof_utils.py
```diff
@@ -199,6 +199,114 @@
...
+                                     [{ru.EVENT: 'state'            , ru.STATE: rps.DONE                   },
+                                      {ru.EVENT: 'state'            , ru.STATE: rps.CANCELED               },
+                                      {ru.EVENT: 'state'            , ru.STATE: rps.FAILED                 } ]]
+    }
+}
+
+# Set of default unit durations for RADICAL-Analytics. All the durations
+# are contiguos.
+UNIT_DURATIONS_DEBUG = {
+    'consume' : {
```
andre-merzky, line 248: Similar to the command on `p_*`, I would expand `u_*` to `unit_*`, or better even to `task_*`.

src/radical/pilot/utils/prof_utils.py
```diff
@@ -199,6 +199,114 @@
...
+                                           {ru.EVENT: 'exec_mkdir'      , ru.STATE: None                            }  ], 
+        'u_agent_execute_mkdir'        : [ {ru.EVENT: 'exec_mkdir'      , ru.STATE: None                            },
+                                           {ru.EVENT: 'exec_mkdir_done' , ru.STATE: None                            }  ],
+        'u_agent_execute_layer_start'  : [ {ru.EVENT: 'exec_mkdir_done' , ru.STATE: None                            },
+                                           {ru.EVENT: 'exec_start'      , ru.STATE: None                            }  ],
+        'u_agent_execute_layer'        : [ {ru.EVENT: 'exec_start'      , ru.STATE: None                            },
+                                          [{ru.EVENT: 'exec_ok'         , ru.STATE: None                            },
+                                           {ru.EVENT: 'exec_fail'       , ru.STATE: None                            } ]], # orte, ssh, mpi, ... 
+        'u_agent_cu_start'             : [ {ru.EVENT: 'cu_start'        , ru.STATE: None                            },
```
andre-merzky, line 278: the `_cu_` midfix (?) here refers to the unit shell script.  We may want to rename this (and likely should also rename the events), maybe to `_lm_` (for launch method, as those are somewhat launch method specific events), or `_script_` (as those are take in the unit shell script) - but the latter would be difficult to apply to function execution (`_lm_` would work there).

src/radical/pilot/utils/prof_utils.py
```diff
@@ -199,6 +199,114 @@
...
+        'u_agent_cu_pre_execute'       : [ {ru.EVENT: 'cu_pre_start'    , ru.STATE: None                            },
+                                           {ru.EVENT: 'cu_pre_stop'     , ru.STATE: None                            }  ],
+        'u_agent_cu_execute_start'     : [ {ru.EVENT: 'cu_pre_stop'     , ru.STATE: None                            },
+                                           {ru.EVENT: 'cu_exec_start'   , ru.STATE: None                            }  ],
+        'u_agent_cu_execute'           : [ {ru.EVENT: 'cu_exec_start'   , ru.STATE: None                            },
+                                           {ru.EVENT: 'cu_exec_stop'    , ru.STATE: None                            }  ],
+        'u_agent_cu_stop'              : [ {ru.EVENT: 'cu_exec_stop'    , ru.STATE: None                            },
+                                           {ru.EVENT: 'cu_stop'         , ru.STATE: None                            }  ],
+        'u_agent_cu_unschedule_start'  : [ {ru.EVENT: 'exec_stop'       , ru.STATE: None                            },
```
andre-merzky, line 290: The durations from here on should not have the `_cu_` midfix, they are not related to the unit LM script.

src/radical/pilot/utils/prof_utils.py
```diff
@@ -199,6 +199,114 @@
...
+        'u_agent_cu_push_to_umngr'     : [ {ru.EVENT: 'unschedule_stop' , ru.STATE: None                            },
+                                           {ru.EVENT: 'state'           , ru.STATE: rps.UMGR_STAGING_OUTPUT         }  ], # push/pull mongodb
+        'u_umngr_cu_destroy'           : [ {ru.EVENT: 'state'           , ru.STATE: rps.UMGR_STAGING_OUTPUT         },
+                                          [{ru.EVENT: 'state'           , ru.STATE: rps.DONE                        },
+                                           {ru.EVENT: 'state'           , ru.STATE: rps.CANCELED                    },
+                                           {ru.EVENT: 'state'           , ru.STATE: rps.FAILED                      } ]]
+    }
+} 
+
+
```
andre-merzky, line 309: I think I disagree with the unschedule parts.  My suggestion would be to create a continuous line without the `unschedule_*` events, and add `u_agent_unschedule` *additionally*, since the unschedule activity happens really concurent to the task's state progression (and the normal event timeline).  This also ensures consistentcy: the above can lead to negative durations on a fast unschedule for example.

src/radical/pilot/utils/prof_utils.py
mturilli: I derived the events you see from an actual run. Some events might have been missing (bootstrap_0_stop was missing). What would be the event pair? Would those be contiguous?

src/radical/pilot/utils/prof_utils.py
mturilli: I am indeed concern about length. I think it is mnemonically simple to associate p and u to pilot and unit. I would try like this and see whether it is a major issue. Would you be fine with that?

src/radical/pilot/utils/prof_utils.py
mturilli: 'lm' is probably more difficult to parse than 'u' or 'p' but done as 'cu' was redundant with 'u'.

src/radical/pilot/utils/prof_utils.py
mturilli: Done.

src/radical/pilot/utils/prof_utils.py
mturilli: Done.

src/radical/pilot/utils/prof_utils.py
andre-merzky: I don't like it, but also don't mind too much, no worries.

mturilli:
Writing 'manager' as 'mngr' follows the rule of deriving abbreviations by omitting vowels. 'mgr' is just wrong and, as such, should be changed in RTC too. I changed it anyway but I expect to get MANY brownie points for doing that, points that one day I will cash in ;)

src/radical/pilot/utils/prof_utils.py
andre-merzky: Have a look [here](https://github.com/radical-cybertools/radical.pilot/blob/devel/src/radical/pilot/utils/prof_utils.py#L36) where we define the current durations used for RA utilization: the `ignore` and `term` durations would be the ones to check.  Yes, they are continuous.

andre-merzky:
> Writing 'manager' as 'mngr' follows the rule of deriving abbreviations by omitting vowels. 'mgr' is just wrong and, as such, should be changed in RTC too. I changed it anyway but I expect to get MANY brownie points for doing that, points that one day I will cash in ;)

Yes, you earned those brownie points :-D  To my defense though, let me quote [webster](http://wordcentral.com/cgi-bin/student?mgr), and I relinguish my claims on the `pilot_` and `task_` prefixes :-D

andre-merzky:
BTW: I consider removing the `cu_cd_done` event: it was used to dig down on the file system scaling issues we had o BW, but nowadays the unit sandbox is mostly created in very remote parts of the code (and is now also done in bulk and asynchronously), and FS caching thus has no chance anymore of interferring with the `cd` call, and so the respective duration (`u_agent_lm_start`) is not very meaningful.  Do you have any concerns about that removal?  If so, I would not mind pushing this change into this PR?

andre-merzky:
PS.: the PR *de*creases code coverage of tests.

mturilli:
If we think there is no use for the cd event anymore I would be happy to remove it.

OK, OK, you get one point for giving up unit and pilot :) I would happily trade a -0.44% reduction for a 70% increase in the overall coverage due to adding more unit tests...

mtitov:
Andre already raised the point about using `task` instead of `unit`, I also would root for this, since (as I understand) there was a plan to start to use `task` some time. p.s. in terms of these changes `t_`

/my two cents/

mturilli:
Thanks @mtitov. I am for using 'task' too but I would like not to have two coexisting naming in RP. If we decide to move from cu to task, I would be happy to help and change it in all the code and documentation. Would you both agree?

andre-merzky:
> Thanks @mtitov. I am for using 'task' too but I would like not to have two coexisting naming in RP. If we decide to move from cu to task, I would be happy to help and change it in all the code and documentation. Would you both agree?

Yes!

mtitov:
@mturilli yeah, make sense (to change all together), agree!

mturilli:
https://travis-ci.com/github/radical-cybertools/radical.pilot/jobs/369185721 shows failures that for the most part seems reasonable to fix. Should I proceed? I think I can fix the following:
```
src/radical/pilot/utils/prof_utils.py
  445  20   W0612   Unused variable 'k' (unused-variable)
  608  25   W0613   Unused argument 'session' (unused-argument)
  615  11   W0612   Unused variable 'anodes' (unused-variable)
  615  19   W0612   Unused variable 'pnodes' (unused-variable)
  745  15   W0612   Unused variable 'anodes' (unused-variable)
  870  27   W0613   Unused argument 'session' (unused-argument)
  957  11   W0612   Unused variable 'anodes' (unused-variable)
  957  19   W0612   Unused variable 'pnodes' (unused-variable)
```
I think I would leave these to you:
```
  517  51   W0212   Access to a protected member _pilot_state_values of a client class (protected-access)
  518  51   W0212   Access to a protected member _pilot_state_inv_full of a client class (protected-access)
  520  51   W0212   Access to a protected member _unit_state_values of a client class (protected-access)
  521  51   W0212   Access to a protected member _unit_state_inv_full of a client class (protected-access)
src/radical/pilot/agent/executing/shell.py
  34  8   E0402   Attempted relative import beyond top-level package (relative-beyond-top-level)

src/radical/pilot/agent/executing/shell_fs.py
  34  8   E0402   Attempted relative import beyond top-level package (relative-beyond-top-level)

src/radical/pilot/agent/executing/sleep.py
  90  4   E0202   An attribute defined in pilot.agent.executing.sleep line 49 hides this method (method-hidden)
``

**Comment by mturilli** [comment]
Not sure I follow, sorry. I think having spurious variables or arguments is not a matter of nice code. I remember you thought that accessing protected members was in some cases the right thing to do so this is why I left that to you. I have been reading about E0402 and there seem to be valid reasons not to do it but I would leave it to you to judge whether our situation is special enough for us to do it.

**Comment by mturilli** [comment]
OK, all test passes now. I had a look at `prof_utils.py` to see whether I could spot an easy test to add to be a good citizen and increase test coverage but I am afraid I am still too inexperience to pull that out in a reasonable time. This is blocking RA release at the moment so I would ask whether we could go ahead and merge it anyway.

File: src/radical/pilot/utils/prof_utils.py (hunk 1 above)
Comment (mturilli) [reply src/radical/pilot/utils/prof_utils.py]: Apologies, I had missed this one. I pushed the termination events following, I think, your indications. How should we managed the 'tagging' of the events? I am not sure I understand why we would ignore an event, and do we know for certain that when we record `bootstrap_0_stop` the pilot has been already terminated? This is why we assume that `bootstrap_0_stop -> DONE/CANCELED/FAILED` does not consume resources?

Comment (andre-merzky) [reply src/radical/pilot/utils/prof_utils.py]: `ignore` here means that the respective duration is ignored when computing utilization.  `bootstrap_0_stop` is the last event written by the agent (or, more precisely, by the bootstrapper which started the agent) [before it finishes](https://github.com/radical-cybertools/radical.pilot/blob/devel/src/radical/pilot/agent/bootstrap_0.sh#L1928).  Well, it does some more things, like creating a tarball of the profiles - but at that point the profiles are closed and we consider the agent done.

**Comment by andre-merzky** [comment]
@mturilli : sorry, I wasn't clear: with the `awkward` bit I mean the `noqa` comments *on each line*.  I mean, the linting is supposed to make the code better to read, not to add code-unrelated comments all over the place, right?  I would probably prefer to disable the warnings for that file, or even globally if you (like me) think that this formatting is good and should be allowed.

**Comment by mturilli** [comment]
I do not like the `noqa` for each line either but I am happy to live with that as I am OK with using flake8. Those line break our coding formatting and, well, I paid the price for doing it. A possibly different approach would be:  could I omit `ru.STATE: None`, `ru.Event: STATE` and `ru.EVENT: 'cmd'` altogether?

File: src/radical/pilot/utils/prof_utils.py (hunk 1 above)
Comment (mturilli) [reply src/radical/pilot/utils/prof_utils.py]: Taggin like in a sort of annotation of groups of duration, i.e., ignore, consumed, etc.  For me, those are cumbersome as they are meaningful only in the contest of our current implementation of resource utilization. In all other contexts, I need to get rid of them. If I have to use different 'tags' also in the *DEBUG data structures, then RA users will not be able to directly use `rp.utils.*DEBUG`. I will have to go around it with a dedicated call that eliminate  one of the level of the dict of dict, or something like that.

Comment (mturilli) [reply src/radical/pilot/utils/prof_utils.py]: About pilot runtime, I guess bootstrap_0_start/stop is close enough? Do you have a better suggestion?

**Comment by andre-merzky** [comment]
> I do not like the `noqa` for each line either but I am happy to live with that as I am OK with using flake8.

I don't mind using flake8 either, but if we need to add comments to 90 lines, we are using it wrong.  I did not mean to turn flake8 off, but to exclude the specific tests for this specific file in a single comment on top, instead of adding those 90 comments.  And if you think that this formatting works in general, then we could turn off the *specific tests* globally...

File: src/radical/pilot/utils/prof_utils.py (hunk 1 above)
Comment (andre-merzky) [reply src/radical/pilot/utils/prof_utils.py]: Tagging: I see.  I would suggest to do that in a separate structure.  For example, the utilization could be defined as:
```py
pilot_utilization = {
    'provide' : {
        'total': ['p_agent_runtime']
    },
    'consume': {
        'boot'  : ['p_agent_ve_setup_init', 'p_agent_ve_setup', ...,]
        'launch': ['p_agent_launching'],
        'term'  : ['p_agent_terminate']
    },
    'agent': {
        'total': ['p_agent_runtime']
    }
}
```
So we would list all durations which contribute to a particular metric.  Would that make sense to you?

andre-merzky: > About pilot runtime, I guess bootstrap_0_start/stop is close enough? Do you have a better suggestion?

This one is exactly correct IMHO.

mturilli:
If those comments irk you, we can get with a file-wide exception I guess. Personally, I agree with the rules of not having spaces allover the places as most of the time they make difficult for me to read the code. Maybe it is my partial blindness or just my brain tricking me but I spent all the time being 'surprised' by the spaces. As ironic as it can be, those data structures are so baroque that unnatural spacing made them treatable for me when I was writing them. So I guess you are right: I am doing something wrong. From there, the question about: can I just omit ru.STATE: None, ru.Event: STATE and ru.EVENT: 'cmd' altogether? If so, I could simplify those things and avoid using all those spaces. Otherwise, I can follow your suggestion above and start to separate those monsters into independent data structures that, after all, would probably be the right thing to do?

andre-merzky:
> can I just omit ru.STATE: None, ru.Event: STATE and ru.EVENT: 'cmd' altogether?

No, alas not at the moment: the RA event selector will barf if `STATE` is not given.

mturilli:
Can we merge this? I need to get this into the RP release to to then release also RA.

andre-merzky:
Would #2202 wok for you?  Otherwise this is good to go, IMHO

mturilli:
Whatever works for you and help to move on.

mturilli:
I see your point about not waiting to fix the abnormal formatting. I am going to merge your other patch and probably add a method that simplifies the writing of the data structure. Thank you for the feedback, very much appreciated.

andre-merzky:
Thanks Matteo, and sorry about the back-and-forth...

src/radical/pilot/utils/prof_utils.py
mturilli: OK, I gave it a go and I come up with something similar to what you seemed to suggest. Please have a look and see whether it matches your expectations. To be usable, it would require extending the RU methods but I would do that later on as, at the moment, nobody uses the DEBUG durations for RU.

andre-merzky: Looks good to me!  I have some minor comments, but they do not hold up a merge.

src/radical/pilot/utils/prof_utils.py
```diff
@@ -23,182 +24,484 @@
...
+                                            ru.STATE: s.STATE_NAME},
+                                           {ru.EVENT: 'cmd',
+                                            ru.MSG: 'message_name'}]}
+    '''
+
+    ldurations = {}
+
+    for k,v in sdurations.items():
+
+        ldurations[k] = []
```
andre-merzky, line 250: You may want to use 
```
ldurations[k] = list()
```
more explicit, and easier to read to people not yet fluent in Python.  I mean, `[]` for list is pretty clear to most, but it is getting confusing when looking at tuples and sets for instance, so we elsewhere usually use the class names for clarity.

src/radical/pilot/utils/prof_utils.py
```diff
@@ -23,182 +24,484 @@
...
+        for ts in v:
+
+            if type(ts) is dict:
```
andre-merzky, line 253: You may want to change 
```py
if type(ts) is dict:
```
to 
```py
if isinstance(ts, dict):
```
That's Python's recommended way of type checking.  It also works for inherited types unlike the above - but that is not an issue in this specific case where you have explicit dict notation, of course.

src/radical/pilot/utils/prof_utils.py
```diff
@@ -23,182 +24,484 @@
...
+        lduration = {ru.EVENT: 'event_name', ru.STATE: None}
+        sduration = {'MSG': 'mesage_name'}
+        lduration = {ru.EVENT: 'cmd', ru.MSG: 'message_name'}
+    '''
+
+    assert(len(sduration.keys()) == 1), 'expand only one short form duration'
+    assert(list(sduration.keys())[0] == 'STATE' or
+           list(sduration.keys())[0] == 'EVENT' or
+           list(sduration.keys())[0] == 'MSG'), 'unknown timestamp type'
+
```
andre-merzky, line 300: I have one more (the review submit swallowed it):
if you want, you can write the above shorter as:
```py
assert(list(sduration.keys())[0] in ['STATE', 'EVENT', 'MSG']), 'unknown timestamp type'
```
although I would actually suggest to just return mismatches unchanged:
```
if len(sduration.keys() != 1:
    return sduration
if list(sduration.keys())[0] not in ['STATE', 'EVENT', 'MSG']:
    return sduration
```
The method then only works on the special entries and leves the others unchanged - for example, if you want to add `event: update_pushed, state: AGENT_EXECUTING` where the full qualification would be needed.

mturilli: Thanks. Done.

src/radical/pilot/utils/prof_utils.py
mturilli: Done.

src/radical/pilot/utils/prof_utils.py
mturilli: Done.

src/radical/pilot/utils/prof_utils.py
mturilli: I thought a bit more about this and I think it is too permissive. I still want to exit on invalid short-formed durations (use an unknown short-form key) and on invalid arbitrary durations with more than two types of timestamp, e.g., {event: update_pushed, state: AGENT_EXECUTING, event: update_pushed, state: AGENT_EXECUTING}

Issue #799, opened 2015-10-06
py.test behaviour with multiple fixtures


antonst:
Converting discussion on e-mail thread into a ticket:

Hi Andre,

I have pushed a small generic example to feature/testing. To run do:
cd tests/test_suites/remote_generic
py.test --junitxml result.xml test_suite_remote_generic.py

Your credentials are in https://github.com/radical-cybertools/radical.pilot/blob/feature/testing/tests/test_suites/pytest_credentials.json and are picked up. Hope this works.

There is a single test: https://github.com/radical-cybertools/radical.pilot/blob/feature/testing/tests/test_suites/remote_generic/test_suite_remote_generic.py#L157 which runs 4 bin/date tasks. There are two fixtures: one for gordon https://github.com/radical-cybertools/radical.pilot/blob/feature/testing/tests/test_suites/remote_generic/test_suite_remote_generic.py#L56 and one for comet https://github.com/radical-cybertools/radical.pilot/blob/feature/testing/tests/test_suites/remote_generic/test_suite_remote_generic.py#L108 Test uses session, pilot, pmgr and umgr from gordon fixture, but workload is executed on both gordon and comet. It should only be executed on gordon! Also I see pilot directories created for both sessions on both machines, e.g. 4 pilot directories in total, instead of two, but only two pilots are launched (which is expected). I have no explanation for this behaviour. If I would set-up doing some arithmetic this would work as expected.

Thanks,
Antons

AntonsT:
Hi Antons,

I just pushed a couple of changes.  In the pilot_launcher, pytest
stumbled over some unicode output -- I convert that into explicit
strings now.  For easier debugging, I replaces the cordon and comet
fixtures with two (different) local ones.  And finally I changed the
cfg file to parseable JSON.  After that, the tests seem to 'run' --
but there the results file seem to say that no test exists actually:

$ cat /home/merzky/saga/radical.pilot/tests/test_suites/remote_generic/result.xml
<?xml version="1.0" encoding="utf-8"?><testsuite errors="0"
failures="0" name="pytest" skips="0" tests="0"
time="8.099"></testsuite>

The local sandbox seem to indicate that two pilots get started, which
seems to match expectations.  Can you please try the current setup,
and let me know if / where that goes wrong for you?

Best, Andre.

AntonsT:
Hi Andre,

Thanks for your feedback.

>After that, the tests seem to 'run' --
>but there the results file seem to say that no test exists actually

[...]

Issue #806, opened 2015-10-07
gtod build fails on comet (sometimes)


marksantcroos:
```
$ cat bootstrap_1.err 
/usr/libexec/gcc/x86_64-redhat-linux/4.4.7/cc1: error while loading shared libraries: libmpfr.so.1: cannot open shared object file: No such file or directory
```

Will investigate.

marksantcroos:
This happened often enough that I submitted a ticket to XSEDE (#36660).

andre-merzky:
We could alleviate the problem by moving it to `$HOME/radical.pilot.sandbox/gtod/`, and only build if  it does not exist.  Once that worked, all following pilots will have it.

A ticket is the right thing to do of course...

marksantcroos:
> We could alleviate the problem by moving it to $HOME/radical.pilot.sandbox/gtod/, and only build if it does not exist. Once that worked, all following pilots will have it.

We would also have to verify that it still works then.

marksantcroos:
```
Mark

Just to update - we had an issue where mpfr was getting left out of the installs on some nodes (oddly enough not on all). In any case we have explicitly added it to the list of install on every node and its in place now.

This issue should be resolved now.

-Mahidhar
```

PR #2198, opened 2020-08-03
handling SAGA job description attribute SystemArchitecture (with SMT, CPUArchitecture, GPUArchitecture)

mtitov:
SMT (SAGA job description) can be set by using env variable
```
export RADICAL_SAGA_SMT=2
```
OR using resource config parameter
```
"saga_jd_supplement"          : {"smt": 2}
```

p.s. related SAGA PR [#799](https://github.com/radical-cybertools/radical.saga/pull/799)


mtitov:
With the latest updates of SAGA Job Description (proposed changes, see SAGA PR [#806](https://github.com/radical-cybertools/radical.saga/pull/806) for details) if `smt` is set using resource config then the format should be the following:
```
"saga_jd_supplement"          : {"system_architecture": {"smt": 2}}
```
p.s. with env variable `RADICAL_SAGA_SMT` no changes

mtitov:
New resource config parameter `"system_architecture": {"smt": <num>, "gpu": <name>, "cpu": <name>}` (this parameter is optional, as well as its dict keys)

mtitov:
@andre-merzky should I remove `RADICAL_SAGA_SMT` now? p.s. I thought it would be kept for quick changes (kinda right inside/outside of the code user can use it to set a different value, but I understand that user can use `Config` class as well)

PR #2199, opened 2020-08-05
Remove `munch` as a dependency and update RU and RS version requirement

mtitov:


andre-merzky:
@iparask : Giannis, do you have any idea what that 32.5% target is codecov keeps talking about?  Merges frequently get blocked because of this, or so it seems.  I thought that we require that coverage does not decrease, but would not require a specific value?

andre-merzky:
@iparask : ignore my earlier comment, please.  Anyway, this is another example wher code cleanup / remove leads to negative coverage.  That happened a couple of times lately and then always prompts admin intervention to get the PR merged.  I would prefer to remove that restriction.  @mturilli : any opinion?

PR #2201, opened 2020-08-07
move staging ops into separate worker

andre-merzky:
Oh boy, this went larger than expacted - but the rationale is this:
  - Both pmgr and pmgr launcher do staging on the pilot's behalf.  
  - Further, the launcher stages several individual files via different code paths.  
  - Further, the launcher had to open an addional job service to untar the pilot sandbox (for untar).

This PR
  - moves all staging ops into a separate worker
  - moves the untar op from the pmmgr launcher to the bootstrapper

The goal is to 
  - consolidate different code paths (migrating toward a generic data stager)
  - reduce the number of connections needed (hi Frontera).

We should also delegate the umgr staging ops to the new component - but that is not implemented as of yet.  Also, there is some code repetition in the pmgr and pmgr launching components which should be consolidated.  Finally, while the channel *toward* the stager is a queue as it should be, the backchannel is curently a pubsub: that's what we get from moving sync ops into an async component.


src/radical/pilot/compute_pilot.py
```diff
@@ -564,35 +579,48 @@ def cancel(self):
...
+        fetch files (default:`staging_output.tgz`) from the pilot sandbox
         '''
 
-        try:
-            psbox = self._session.get_fs_dir(self._pilot_sandbox)
-            psbox.copy('staging_output.tgz', self._client_sandbox)
+        sds = ru.as_list(sds)
+
+        if not sds:
+             sds = [{'source': 'client:///staging_output.tgz',
```
mtitov, line 611: should here be `source`: `pilot...` and `target`: `client...`?

src/radical/pilot/compute_pilot.py
```diff
@@ -564,35 +579,48 @@ def cancel(self):
...
+                     'target': 'pilot:///staging_output.tgz',
+                     'action': rp.TRANSFER}]
```
mtitov, line 613: `rps.TRANSFER`

src/radical/pilot/pmgr/launching/default.py
```diff
@@ -125,6 +132,10 @@ def _pmgr_control_cb(self, topic, msg):
         if cmd == 'pilot_staging_input_request':
 
             self._handle_pilot_input_staging(arg['pilot'], arg['sds'])
```
mtitov, line 134: this method was removed

src/radical/pilot/pmgr/launching/default.py
```diff
@@ -125,6 +132,10 @@ def _pmgr_control_cb(self, topic, msg):
...
+            self.publish(rpc.CONTROL_PUBSUB,
+                         {'cmd': 'pilot_staging_input_result',
```
mtitov, line 136: this command `pilot_staging_input_result` is not read anywhere else

src/radical/pilot/pmgr/launching/default.py
```diff
@@ -125,6 +132,10 @@ def _pmgr_control_cb(self, topic, msg):
...
+                          'arg': {'pilot': pilot,
```
mtitov, line 137: `arg['pilot'], arg['sds']` ?

src/radical/pilot/pmgr/launching/default.py
```diff
@@ -1258,22 +1118,26 @@ def _prepare_pilot(self, resource, rcfg, pilot, expand):
...
+                    ret['fts'].append({'src': cc_path,
+                                       'tgt': '%s/%s' % (session_sandbox, cc_name),
+                                       'rem': False})
 
                 self._sandboxes[resource] = True
 
+        # always stage the bootstrapper for each pilot, but *not* in the tarball
+        bootstrapper_path = os.path.abspath("%s/agent/bootstrap_0.sh"
+                                           % self._root_dir)
+        tgt = '%s/%s/bootstrap_0.sh' % (session_sandbox, pid)
```
mtitov, line 1130: `tgt = '%s/bootstrap_0.sh' % pilot_sandbox`?

src/radical/pilot/pmgr/launching/default.py
```diff
@@ -1336,5 +1199,134 @@ def _prepare_pilot(self, resource, rcfg, pilot, expand):
...
+                   'pilot'   : pilot_sandbox,
+                   'resource': resource_sandbox}
+
+        sds = ru.as_list(sds)
+
+        for sd in sds:
+            sd['prof_id'] = pilot['uid']
+
+        for sd in sds:
+            sd['source'] = str(complete_url(sd['source'], self._rem_ctx, self._log))
```
mtitov, line 1266:
```
sd['source'] = str(complete_url(sd['source'], loc_ctx, self._log))
sd['target'] = str(complete_url(sd['target'], rem_ctx, self._log))
```

may be change variables `loc_ctx` and `rem_ctx` into `src_ctx` and `tgt_ctx`? (wherever it is used, for less confusion)

src/radical/pilot/worker/stager.py
```diff
@@ -0,0 +1,131 @@
...
+            tmp.path = '/'
+            key      = str(tmp)
+
+            with self._cache_lock:
+                if key in self._saga_fs_cache:
+                    fs = self._saga_fs_cache[key]
+
+                else:
+                    fs = rsfs.Directory(key, session=self._session)
+                    self._saga_fs_cache[key] = fs
```
mtitov, line 121: from the previous PR :)
```python
                fs_url      = rs.Url(tgt)
                fs_url.path = '/'
                key         = str(fs_url)

                with self._cache_lock:
                    if key in self._saga_fs_cache:
                        fs = self._saga_fs_cache[key]
                    else:
                       fs = rsfs.Directory(fs_url, session=self._session)
                        self._saga_fs_cache[key] = fs
```

src/radical/pilot/worker/stager.py
```diff
@@ -0,0 +1,131 @@
...
+
+            fs.copy(src, tgt, flags=flags)
+
+            sd['state'] = rps.DONE
+
+            self._prof.prof('staging_in_stop', uid=prof_id, msg=uid)
```
mtitov, line 127: `'staging_in_stop'` or `'staging_stop'`? (since method's name is `_handle_staging`)

src/radical/pilot/compute_pilot.py
andre-merzky: Indeed - thanks!

src/radical/pilot/compute_pilot.py
andre-merzky: Yep.

src/radical/pilot/pmgr/launching/default.py
andre-merzky: thanks, fixed.

src/radical/pilot/pmgr/launching/default.py
andre-merzky: also fixed.

src/radical/pilot/pmgr/launching/default.py
andre-merzky: That code is gone now.

src/radical/pilot/pmgr/launching/default.py
andre-merzky: Changed, thanks

src/radical/pilot/pmgr/launching/default.py
andre-merzky: This will remain confusing: on input staging, the src is local and tgt is remote; on output staging the src is remote and tgt is local...

src/radical/pilot/worker/stager.py
andre-merzky: Hmm, isn't that the same (apart from the name `tmp <-> fs_url`) and using the URL type instead of string on the DIrectory ctor (saga takes both)?

src/radical/pilot/worker/stager.py
andre-merzky: yes, indeed - fixed.

andre-merzky:
@mtitov : as always, thanks for digging through this!

src/radical/pilot/pmgr/launching/default.py
mtitov: agree with explanation, BUT in the code, for example with output staging, src (as remote) is set as `loc_ctx`, that what I mean with confusion, kinda better just set direction when these variables are defined (not when they are used later)

src/radical/pilot/pmgr/launching/default.py
```diff
@@ -1258,23 +1104,25 @@ def _prepare_pilot(self, resource, rcfg, pilot, expand):
...
+                    ret['fts'].append({'src': cc_path,
+                                       'tgt': '%s/%s' % (session_sandbox, cc_name),
+                                       'rem': False})
 
                 self._sandboxes[resource] = True
 
+        # always stage the bootstrapper for each pilot, but *not* in the tarball
+        bootstrapper_path = os.path.abspath("%s/agent/bootstrap_0.sh"
+                                           % self._root_dir)
+        bootstrap_tgt = '%s/bootstrap_0.sh' % (pilot_sandbox, pid)
```
mtitov, line 1116: `pid` need to be removed

src/radical/pilot/pmgr/launching/default.py
mtitov: BTW here are local variables (not class/instance related), thus no `self._` and should be switched, and for local variable could be applied what I propose

src/radical/pilot/worker/stager.py
mtitov: yeah, I though the key/important change is to use URL type instead of string for `rsfs.Directory`

andre-merzky: Ah - the `rsfs.Directory` ctor  won't care.  But ack, I'll make it consistent.

src/radical/pilot/pmgr/launching/default.py
andre-merzky: Thanks, done.

src/radical/pilot/pmgr/launching/default.py
andre-merzky: ack on the `self` (fixed) - but the `remote` reference` to the `source` part is correct AFAICS - I don't see `loc_ctx` being used for the source?

src/radical/pilot/pmgr/launching/default.py
```diff
@@ -1269,8 +1268,8 @@ def _stage_out(self, pilot, sds):
             sd['prof_id'] = pilot['uid']
 
         for sd in sds:
-            sd['source'] = str(complete_url(sd['source'], self._rem_ctx, self._log))
-            sd['target'] = str(complete_url(sd['target'], self._loc_ctx, self._log))
+            sd['source'] = str(complete_url(sd['source'], rem_ctx, self._log))
+            sd['target'] = str(complete_url(sd['target'], loc_ctx, self._log))
```
mtitov, line 1272: may be I'm missing something, just what I see: `source` uses info from `rem_ctx`, which has `'pwd'     : client_sandbox`, and `target` uses info from `loc_ctx`, which has `'pwd'     : pilot_sandbox`. Thus it is a transfer from client to pilot, which looks like stage in, right? but should be switched for stage out?

andre-merzky: Yeah, that sounds wrong, thanks!

andre-merzky: This should be fixed now - would you care to re-review, please @mtitov ?  Thanks!

src/radical/pilot/pilot_manager.py
```diff
@@ -452,27 +452,30 @@ def _pilot_staging_input(self, pilot, directives):
...
+            sd_states = [sd['state'] for sd
+                                     in  self._active_sds.values()
+                                     if  sd['uid'] in uids]
+        # push them out
+        self._stager_queue.put(sds)
+
+        # wait for their completion
         with self._sds_lock:
             sd_states = [sd['state'] for sd in self._active_sds.values()
                                             if sd['uid'] in uids]
```
mtitov, line 481: Is these 3 lines are needed? they kinda duplicate L472-474 even it regenerates to get updated states (it is similar to how this is organized in method `_pilot_staging_input`)

src/radical/pilot/utils/component.py
```diff
@@ -758,8 +763,9 @@ def register_output(self, states, output=None):
         sure that the thing is in fact in a final state.
         '''
 
-        if not isinstance(states, list):
-            states = [states]
+        states = ru.as_list(states)
+        if not states:
+            states = None
```
mtitov, line 768: should it be `states = [None]`? (similarly as for `register_input`) and also with the next line it is expected `states` to be a `list`

src/radical/pilot/utils/component.py
```diff
@@ -791,10 +821,12 @@ def unregister_output(self, states):
         this removes any outputs registerd for the given states.
         '''
 
-        if not isinstance(states, list):
-            states = [states]
+        states = ru.as_list(states)
+        if not states:
+            states = None
```
mtitov, line 826: same here, `states = [None]`?

mtitov:
besides these small corrections above and that `test_launcher` should be updated (just need to add `uid` value: `component._uid = '...'`) everything else looks good to me

src/radical/pilot/pilot_manager.py
andre-merzky: You are right, removed one set of those lines.

src/radical/pilot/utils/component.py
andre-merzky: Ack, fixed.

src/radical/pilot/utils/component.py
andre-merzky: same, thanks

src/radical/pilot/utils/component.py
mtitov: btw, but in `register_input` method it says `worker handles stateless entities` (and there allowed `if not states: states = [None]`). Is it not applied for `output`? Just want to resolve this and we can push this PR

andre-merzky: ack, changed.

src/radical/pilot/utils/component.py
andre-merzky: same

mtitov (approved):

andre-merzky:
not idea how to silence flake8 on `__init__.py` - anybody?

mtitov:
@andre-merzky found 2 solutions:
1. `from .my_class import MyClass  # noqa: F401` (looks like explicitly set the id)
2. use `__all__ = ['Stager', 'Update']` in `worker/__init__.py` (this looks more pythonic)

andre-merzky:
> @andre-merzky found 2 solutions:
> 
>     1. `from .my_class import MyClass  # noqa: F401` (looks like explicitly set the id)
> 
>     2. use `__all__ = ['Stager', 'Update']` in `worker/__init__.py` (this looks more pythonic)

Thanks for looking around, @mtitov !  But urgh, neither works well for large files like `radical/pilot/__init_.py` :-/   Oh well.  We don't touch init files very often, so lets ignore it?

mtitov:
@andre-merzky but if to have it (try option 2) for `worker/__init__.py` for now, since currently, flake8 points to that file

andre-merzky:
> @andre-merzky but if to have it (try option 2) for `worker/__init__.py` for now, since currently, flake8 points to that file

@mtitov : I can see that this would work - but I would like to avoid the route to use one form of code in one place, and another form in another, just to make the linter happy.  Linter is supposed to _**improve**_ code coherence after all...

We don't have any code in `__init__.py` in general (apart from `ru.get_version(...)` in the top level files).  If we agree to keep it that way, it may also fairly safe to exclude those files from linting?

PR #2204, opened 2020-08-13
allow `virtenv_mode=local` to reuse client ve

andre-merzky:
This speeds up pilot starttime during debug


mtitov:
Hi @andre-merzky, I guess I would need more context, sorry. My base tests showed that it would be valid for conda env only, but I might be missing something.
```
### no any virtualenv
$ python3
>>> import sys
>>> sys.base_prefix
'/usr/local/Cellar/python@3.8/3.8.5/Frameworks/Python.framework/Versions/3.8'
```
```
### with py virtualenv
$ source ./Projects/_venv/rp/bin/activate
(rp) $ python
>>> import sys
>>> sys.base_prefix
'/usr/local/Cellar/python@3.8/3.8.5/Frameworks/Python.framework/Versions/3.8'
```
as a note: py virtualenv makes a link to system's python installation
```
### with conda env
$ conda activate rp-dev
(rp-dev) $ python
>>> import sys
>>> sys.base_prefix
'/Users/matitov/Projects/_miniconda/envs/rp-dev'
```

And later, after parameter `virtenv` is set, it will be used in [bootstrap_0.sh](https://github.com/radical-cybertools/radical.pilot/blob/497dd8e6f62cb5b1e647ce7964ae6df28cc2d4bb/src/radical/pilot/agent/bootstrap_0.sh#L844). 
\* If it is for conda only, then need to add `virtenv_dist == 'anaconda'`

andre-merzky:
Wow - I did not expect sys.base_prefix to be different in virtualenv - what's the point then?  Thanks for catching that, I'll find a different way to inspect that...

mtitov:
this is more about assigning name of virtual env automatically, may be to use env variables for that? (and assign `virtenv_dist` based on which env variable exists and active)
- py virtenv -> `VIRTUAL_ENV` (e.g., will give `/Users/matitov/Projects/_venv/rp`)
- conda env -> `CONDA_PREFIX` (e.g., will give `/Users/matitov/Projects/_miniconda/envs/stack`)

and later bootstrap would know how to activate them

src/radical/pilot/pmgr/launching/default.py
```diff
@@ -1119,6 +1121,23 @@ def _prepare_pilot(self, resource, rcfg, pilot, expand):
             if virtenv_mode != 'private':
                 cleanup = cleanup.replace('v', '')
 
+        # don't stage sdists on 'installed', 'local'
+        if virtenv_mode in ['installed', 'local']: stage_sdists = False
```
mtitov, line 1125: `virtenv_mode == 'local' or rp_version == 'installed'`? there is no value `'installed'` for `virtenv_mode`

src/radical/pilot/pmgr/launching/default.py
```diff
@@ -1119,6 +1121,23 @@ def _prepare_pilot(self, resource, rcfg, pilot, expand):
...
+        else                                     : stage_sdists = True
+
+        # use local VE ?
+        if virtenv_mode == 'local':
+            if 'VIRTUAL_ENV' in os.environ:
```
mtitov, line 1130: should we check that actual value is not empty? `if os.environ.get('VIRTUAL_ENV')`
no sure that it is really necessary, but kinda worry more for `'CONDA_PREFIX'`

src/radical/pilot/pmgr/launching/default.py
```diff
@@ -5,6 +5,7 @@
 
 
 import os
+import sys
```
mtitov, line 8: with latest updates this is not needed

src/radical/pilot/pmgr/launching/default.py
```diff
@@ -1119,6 +1121,23 @@ def _prepare_pilot(self, resource, rcfg, pilot, expand):
...
                 cleanup = cleanup.replace('v', '')
 
+        # don't stage sdists on 'installed', 'local'
+        if virtenv_mode in ['installed', 'local']: stage_sdists = False
+        else                                     : stage_sdists = True
+
+        # use local VE ?
+        if virtenv_mode == 'local':
+            if 'VIRTUAL_ENV' in os.environ:
+                virtenv_dist = 'default'
```
mtitov, line 1131: `python_dist = 'default'`

src/radical/pilot/pmgr/launching/default.py
```diff
@@ -1119,6 +1121,23 @@ def _prepare_pilot(self, resource, rcfg, pilot, expand):
...
+                virtenv      = os.environ['VIRTUAL_ENV']
+            elif 'CONDA_PREFIX' in os.environ:
+                virtenv_dist = 'anaconda'
```
mtitov, line 1134: `python_dist = 'anaconda'`

src/radical/pilot/pmgr/launching/default.py
```diff
@@ -1119,6 +1121,23 @@ def _prepare_pilot(self, resource, rcfg, pilot, expand):
             if virtenv_mode != 'private':
                 cleanup = cleanup.replace('v', '')
 
+        # don't stage sdists on 'installed', 'local'
+        if virtenv_mode in ['installed', 'local']: stage_sdists = False
+        else                                     : stage_sdists = True
+
+        # use local VE ?
+        if virtenv_mode == 'local':
```
mtitov, line 1129: we can reassign it to `'use'` value:
```python
if virtenv_mode == 'local':
    virtenv_mode = 'use'
    ... 
    # check env var, set virtenv and python_dist
```
and then we don't need new value for `virtenv_mode` in `bootstrap_0.sh`

src/radical/pilot/pmgr/launching/default.py
```diff
@@ -1235,32 +1257,37 @@ def _prepare_pilot(self, resource, rcfg, pilot, expand):
...
 
                 # Some machines cannot run pip due to outdated CA certs.
                 # For those, we also stage an updated certificate bundle
                 # TODO: use booleans all the way?
                 if stage_cacerts:
 
-                    cc_name = 'cacert.pem.gz'
-                    cc_path = os.path.abspath("%s/agent/%s" % (self._root_dir, cc_name))
-                    self._log.debug("use CAs %s", cc_path)
+                    certs = 'cacert.pem.gz'
```
mtitov, line 1285: new variable name should be updated in L1367 as well

src/radical/pilot/pmgr/launching/default.py
mtitov: p.s. I see that in one of my previous comments I made this error... (`virtenv_dist is 'default' or 'system'` is used only if `python_dist == 'default'` and `python_dist is 'default' or 'anaconda'`)

src/radical/pilot/pmgr/launching/default.py
andre-merzky: Thanks, good catch!

src/radical/pilot/pmgr/launching/default.py
andre-merzky: Yeah, I had that at some point, but worried that it is too obfuscated when reading  / debugging the bootstrapper, and thus rather would leave it explicit.

src/radical/pilot/pmgr/launching/default.py
andre-merzky: Ack, done.

src/radical/pilot/pmgr/launching/default.py
andre-merzky: oops - fixed!

src/radical/pilot/pmgr/launching/default.py
andre-merzky: Thanks!

andre-merzky:
Thanks for your feedback, @mtitov - that caught quite a number of things.  It seems to work now for conda and VE - could you have another look?  I would like to get this one out as well...

mtitov:
@andre-merzky have a follow up question: when `virtenv_mode == 'local'` do we assume that we also use RCT stack from this local virtual env? If yes, then should we add [here](https://github.com/radical-cybertools/radical.pilot/blob/8cfe4c21a35678e1646bdd7bd072e0e9ba29a947/src/radical/pilot/pmgr/launching/default.py#L1119) `rp_version = 'installed'` (for both `python_dist = 'default|anaconda'`)

and as part of cleanup - `rp_version` doesn't have value `'debug'`, right? ([link](https://github.com/radical-cybertools/radical.pilot/blob/8cfe4c21a35678e1646bdd7bd072e0e9ba29a947/src/radical/pilot/pmgr/launching/default.py#L1078))

src/radical/pilot/pmgr/launching/default.py
mtitov: don't mind both, but I would treat `'local'` value as a flag parameter to set all others including itself (`python_dist`, `virtenv_mode`, `virtenv`, `rp_version`)

andre-merzky:
> @andre-merzky have a follow up question: when virtenv_mode == 'local' do we assume that we also use RCT stack from this local virtual env? If yes, then should we add here rp_version = 'installed' (for both python_dist = 'default|anaconda')

Hmm, fair - in practice that is likely the most common case.  But I don't think we should enforce it, other options still work.  Well, they should - let me test and report back! :-) 

> and as part of cleanup - rp_version doesn't have value 'debug', right? (link)

Ah, I tried to clean out the `debug` path - thanks for noting that one, I'll fix it.

andre-merzky:
Yeah, that still works - I just used a local VE with RP 1.5.2 with an `rp_version = @1.4.0`, and that had the pilot then running under 1.4.0.  I would like to keep that functionality...

mtitov:
@andre-merzky one more suggestion and I'm done :) for `rp_version`
- remove default value ([here](https://github.com/radical-cybertools/radical.pilot/blob/c7a07ef26e097ecaeddceadcce658569efaa3843/src/radical/pilot/pmgr/launching/default.py#L913)) and to have `rp_version              = rcfg.get('rp_version')`
- put the following code [here](https://github.com/radical-cybertools/radical.pilot/blob/c7a07ef26e097ecaeddceadcce658569efaa3843/src/radical/pilot/pmgr/launching/default.py#L1076)
```python
        if rp_version is None:
            if virtenv_mode == 'local':
                rp_version = 'installed'
            else:
                rp_version = DEFAULT_RP_VERSION
```
Thus, if we want to use everything from the local virtenv, then `"virtenv_mode": "local"` would be enough, otherwise corresponding `rp_version` can be set.

andre-merzky:
Yeah, that's fair, I can live with that.  Thanks :-)

docs/architecture/concepts/ve_separation/runme.sh
```diff
@@ -0,0 +1,53 @@
+#!/bin/sh
+
+CONDA_ROOT='/home/merzky/.miniconda3/'
```
mtitov, line 3: should here be kept `CONDA_ROOT="$HOME/.miniconda3/"`?

docs/architecture/concepts/ve_separation/runme.sh
```diff
@@ -0,0 +1,53 @@
...
+# prepare a module libdir
+os_path=$(python -c 'import os; print(os.__file__)')
+cp $os_path modules/my_python_dir/lib/ | progress 'create module'
+
+
+# create a conda env for another test workload
+# run in subshell to maintain the shell env
+(
+    . $CONDA_ROOT/etc/profile.d/conda.sh
+    conda create -y -n conda_test | progress 'create conda env'
```
mtitov, line 33: `conda create -y -n conda_test python=3.7` (need to add python installation)

mtitov (approved):
LGTM!

docs/architecture/concepts/ve_separation/runme.sh
mtitov: and in `test_conda.sh`? (kinda to have a "universal" test without changes on user side)

PR #2205, opened 2020-08-17
Arbitrary udurations for consumed resources

mturilli:
This enable defining arbitrary units durations for `get_consumed_resources`. This allows to correctly plot resource utilization when some event timestamps are not available. It is backward compatible with the previous API.


andre-merzky (approved):
LGTM!

PR #2207, opened 2020-08-19
Test/worker

andre-merzky:
This PR adds a test for the task overlay worker which can run that RP component in isolation.  That should also provide a blueprint of running other RP components in isolation.


andre-merzky:
This should be ready to merge, but needs #2204 to be merged first.

Issue #808, opened 2015-10-07
relative path in pilot.stage_in()


andre-merzky:
I use:

```
          pilot.stage_in({'source': 'input.dat',
                          'target': 'staging:///input.dat',
                          'action': rp.TRANSFER})
```

and would expect that `input.dat` is interpreted relative to the application dir.  It is, however, interpreted relative to the pilot sandbox (it seems), as I get this exception:

```
Exception: copy (file://localhost/home/merzky/radical.pilot.sandbox/rp.session.cameo.merzky.016715.0064-pilot.0000/staging_area/input.dat
  -> file://localhost/home/merzky/radical.pilot.sandbox/rp.session.cameo.merzky.016715.0064-pilot.0000/staging_area/input.dat) 
  failed (1): (out: cp: cannot stat 
    ‘/home/merzky/radical.pilot.sandbox/rp.session.cameo.merzky.016715.0064-pilot.0000/staging_area/input.dat’: 
    No such file or directory
) (err: None) 
(/home/merzky/saga/radical.pilot.docs_am/ve/local/lib/python2.7/site-packages/saga/adaptors/shell/shell_file.py +527 (copy)  :  % (src, tgt, ret, out, err)))

```

andre-merzky:
see also #1192

andre-merzky:
This is addressed  by now, we support different schemas to point to client, unit, resource and / or agent sandboxes.

PR #2209, opened 2020-08-19
Update for Summit's resource config (added option `gpumps` explicitly)

mtitov:
related to SAGA PR [#808](https://github.com/radical-cybertools/radical.saga/pull/808)


andre-merzky (approved):
:+1:

PR #2213, opened 2020-08-27
Hotfix/prte profiling

mturilli:
The previous PR broke two unit tests. I noticed that after approving it. Here the fix.


andre-merzky (approved):
LGTM

Issue #2152, opened 2020-05-22
Tagging is not working any more


iparask:
Tags are defined in the Unit description as a dictionary (see [here](https://github.com/radical-cybertools/radical.pilot/blob/devel/src/radical/pilot/compute_unit_description.py#L41) and [here](https://github.com/radical-cybertools/radical.pilot/blob/devel/src/radical/pilot/compute_unit_description.py#L385)), while the scheduler checks for `tag` (see [here](https://github.com/radical-cybertools/radical.pilot/blob/devel/src/radical/pilot/agent/scheduler/continuous.py#L364)).

First the check in the scheduler is different from the unit description key and second, there is no example in the documentation that shows how a tag can be a dictionary. If I remember correctly it was a string. This in turn affects EnTK's tagging capabilities

andre-merzky:
closed as fixed.

PR #2217, opened 2020-09-01
Added "tag" attribute to ComputeUnitDescription

mtitov:
This is a temporary solution until `tags` attribute in `ComputeUnitDescription` will be reworked. Currently it will bring possibility to use `tag` for `Continuous` scheduler ([L468](https://github.com/radical-cybertools/radical.pilot/blob/fix/unit_description_tag/src/radical/pilot/agent/scheduler/continuous.py#L468))

Related ticket is #2152 and other related branch for tags update (co-location) is [fix/tags](https://github.com/radical-cybertools/radical.pilot/tree/fix/tags)

@iparask and @lee212 can you please check EnTK with this branch and confirm that this update will not brake it.


andre-merzky (approved):
Thanks!

PR #2222, opened 2020-09-08
Fix python stack on Summit

mturilli:
This should fix the Agent's Python stack on Summit. Tested only with JSRUN but opened because of the release.


andre-merzky (approved):
Thanks!

PR #2224, opened 2020-09-10
Launch method PRTE v2

mtitov:


andre-merzky (approved):
LGTM, thanks!  

If you don't mind, could you please add a test similar to [prte](https://github.com/radical-cybertools/radical.pilot/blob/devel/tests/test_lm/test_unit/test_prte.py), and some test cases?  That can go into a separate PR though.

mtitov:
@andre-merzky rushed a little bit and added tests here, would that still work?

andre-merzky:
Yes, thanks! :-)

Issue #2114, opened 2020-04-09
example cleanup


andre-merzky:
@aydinsaribudak and @AymenFJA began to clean up the RP examples - this ticket is to discuss those changes where needed.

This work lives in the `fix/travis` branch.

andre-merzky:
From @aydinsaribudak :
```
/misc/hello_synapse.py (DONE)
1. declaration of cud.cores is removed since it is not in unitmanager schema.
session id: rp.session.js-17-94.jetstream-cloud.org.aydins.018359.0002
session id: rp.session.js-17-94.jetstream-cloud.org.aydins.018360.0009

/misc/ordered_pipelines.py (DONE) 
1. config = ru.read_json('%s/../config.json' %pwd) is the update.
This is to read the config file successfully.
session id: rp.session.js-17-94.jetstream-cloud.org.aydins.018359.0009
session id: rp.session.js-17-94.jetstream-cloud.org.aydins.018360.0010

/misc/profile_analysis.py is removed.
1. This is because radical.pilot.utils is missing these attributes: 'prof2frame' and 'combine_profiles'


/misc/rp_app_comm.py (DONE)
session id: rp.session.js-17-94.jetstream-cloud.org.aydins.018360.0003
session id: rp.session.js-17-94.jetstream-cloud.org.aydins.018360.0013

/misc/rp_app_master.py 
/misc/rp_app_worker.py are removed because
following are not available in os.environ
'RP_UNIT_ID','RP_WORK_QUEUE_IN', 'RP_RESULT_QUEUE_OUT'
For example, we got error with this line: uid = os.environ['RP_UNIT_ID']

/misc/task_overlay.py
/misc/task_overlay_master.py
/misc/task_overlay_worker.py
1. resource is fixed.
2. import statement is fixed.
However, it requires a second argument (i.e. worker) when I ran the script.
It may work with correct arguments.
session id: rp.session.js-17-94.jetstream-cloud.org.aydins.018360.0060

/misc/wl_shape_02.py (DONE)
1. pilot.stage_in source path is fixed, and an empty gromacs folder is created 
(maybe this folder should contain some specific scripts similar to those saved 
under ../data/gromacs_mdrun_0/ folder)
session id: rp.session.js-17-94.jetstream-cloud.org.aydins.018360.0025


/misc/dynamic_ec2_pilot.py is removed because in order to run this script
following environment variables should be set:
[...]

PR #2226, opened 2020-09-11
Fix/travis

AymenFJA:
This PR is based on this ticket https://github.com/radical-cybertools/radical.pilot/issues/2114#issuecomment-665137207 and the purpose of it is to `update/clean` the examples under `radical.pilot` repository.


iparask (approved):
This PR seems okay. Pylint fails because it tries to lint removed files.

PR #2227, opened 2020-09-15
Feature/prep env

andre-merzky:
Prepare task specific env settings:
  - add `prepare_env` key to pilot description
  - add `prepare_env` method to pilot (the above key internally just triggers that method)
  - up calling the above, the parameter is passed as `pilot_command` (via MongoDB) to the agent
  - when receiving said `pilot_command` (`prep_env`), the pilot will call `radical-pilot-create-static-ve` with some parameters to create the respective named environment
  - add an `named_env` parameter to the task description
  - if said parameter is present, the `popen` executer will activate the respective env in the task shell script

Missing
  - support for conda
  - ensure that env activation survives the launch method invocation
  - proper env isolation (the pilot's `deactivate` is called, but that needs verification.  Do we need to call it before `prep_env`?)
  - should task pre_exec's happen before or after env activation?


andre-merzky:
@mtitov : ping

src/radical/pilot/agent/executing/popen.py
```diff
@@ -135,12 +135,26 @@ def work(self, units):
...
+
+            # ensure that the named env exists
+            env = descr.get('named_env')
+            if env:
+                if not os.path.isdir('%s/%s' % (self._pwd, env)):
+                    raise ValueError('invalid named env %s for task %s' 
+                                    % (env, cu['uid']))
+                pre = ru.as_list(descr.get('pre_exec'))
+                pre.insert(0, '. %s/%s/bin/activate' % (self._pwd, env))
+                pre.insert(0, '. %s/deactivate'      % (self._pwd))
```
mtitov, line 148: do we create an executable file `deactivate`? otherwise it should be `pre.insert(0, 'deactivate')`

p.s. for `conda` env it would be `pre.insert(0, 'conda %s/%s/bin/deactivate' % (self._pwd, env))`, but we keep it for later

src/radical/pilot/task_overlay/worker.py
```diff
@@ -217,6 +217,29 @@ def _call(self, data):
         return out, err, ret
 
 
+    # --------------------------------------------------------------------------
+    #
+    def _mpi(self, data):
```
mtitov, line 222: there is an inconsistency in this method (with loaded modules and used corresponding methods), and I wasn't sure what's inside variable `data` and it should be input for `rpa.LaunchMethod.create(...`?

p.s. may be move this method to another PR?

mtitov:
@andre-merzky sorry for delay with this review, the part with `virtualenv` looks good to me, but I'd like to discuss some of it parts - e.g., having `virtualenv` inside `conda` env (?)

src/radical/pilot/agent/executing/popen.py
andre-merzky: The bootstrapper does indeed create a `deactivate` script [here](https://github.com/radical-cybertools/radical.pilot/blob/devel/src/radical/pilot/agent/bootstrap_0.sh#L22).  That is a bit simplistic at the moment and will get some more sophistication eventually - but this is what is being sourced here.

src/radical/pilot/task_overlay/worker.py
andre-merzky: Hmm, how did that get in here indeed - that's unrelated I think...  Let me check what happened, will move it out.

andre-merzky: An update from devel solved this :-)

mtitov (approved):
LGTM

Issue #2221, opened 2020-09-08
RADICAL-Pilot shell executor cu['description']['mpi'] need to be updated.


AymenFJA:
This issue was raised while running a unit test for the `_handle_unit` in `shell.py`:

```
1599593324.807 : dummy                : 1198  : 140193279575872 : ERROR    : error running CU
Traceback (most recent call last):
File "/home/aymen/ve/RCT/lib/python3.8/site-packages/radical/pilot/agent/executing/shell.py", line 240, in _handle_unit
if cu['description']['mpi']:
KeyError: 'mpi'
```

Reason: shell `cu['description']['mpi']` is old and need to be updated to `cu['description']['cpu_process_type'] = 'MPI'`.
Fix: I will provide a pull request soon.
Test: [test_hand_unit](https://github.com/radical-cybertools/radical.pilot/blob/016eda617efab6099e9bb8178440dd609b040f49/tests/test_executing/test_unit/test_shell.py#L30)

PR #2231, opened 2020-10-01
Fix cu description for shell executor based on issue #2221

AymenFJA:
This PR is based on issue #2221. The PR is fixing the `CU description`  in the `Shell executor`.


iparask (approved):
Seems good and the test is in another branch

PR #2233, opened 2020-10-02
Fixed agent config definition (in PilotManager)

mtitov:
Paths should be consistent, thus `ru.Config` should be responsible to set right ones and pick configs from there


andre-merzky (approved):
Thanks for cleaning that up!  I don't mind merging that already if it works as expected.  Can you run a small test with overloading some setting in `~/.radical/pilot/configs` to make sure?  Thanks!

mtitov:
@andre-merzky yes, confirm that! Just made several tests at Frontera, works the same as for resource configs (to pick new user's config or to update an existing)

PR #2242, opened 2020-10-22
Feature/test agent executing

AymenFJA:
Merging the `executors unit test` with `Devel.`


andre-merzky:
Alas, the unit tests fail on Travis - @AymenFJA : can you have a look at that failure, please?

AymenFJA:
> Alas, the unit tests fail on Travis - @AymenFJA : can you have a look at that failure, please?

Yes, I will check/fix and report.

iparask:
Apparently, purest does not like test with the same filename. We had a `test_base.py` for the agent executing component and scheduler. I renamed those files and not it seems to work.

iparask (approved):
I am okay with this PR.

andre-merzky:
@iparask: that naming problem is awkward, thanks for finding it!  I am a bit worried that we stumble over that one again and again - could we follow the advice from [here](https://docs.pytest.org/en/stable/goodpractices.html#tests-outside-application-code) (see second `NOTE` in that section) maybe and use `--import-mode=importlib` - what do you think?

iparask:
Yes, let me check it and fix it properly

iparask:
I followed the current instructions:
```
Note that this scheme has a drawback if you are using prepend import mode (which is the default): your test files must have unique names, because pytest will import them as top-level modules since there are no packages to derive a full package name from. In other words, the test files in the example above will be imported as test_app and test_view top-level modules by adding tests/ to sys.path.

If you need to have test modules with the same name, you might add __init__.py files to your tests folder and subfolders, changing them to packages:
```
This is because the `--import-mode=importlib` mode requires refactoring the tests. I would like to avoid that as much as possible.

iparask:
@andre-merzky, if you are ok with this PR let me know to merge it and move forward with RP's testing.

andre-merzky:
> I followed these instructions:
> 
> ```
> Note that this scheme has a drawback if you are using prepend import mode (which is the default): 
> your test files must have unique names, because pytest will import them as top-level modules since 
> there are no packages to derive a full package name from. In other words, the test files in the example 
> above will be imported as test_app and test_view top-level modules by adding tests/ to sys.path.
> 
> If you need to have test modules with the same name, you might add __init__.py files to your tests 
> folder and subfolders, changing them to packages:
> ```
> 
> This is because the `--import-mode=importlib` mode requires refactoring the tests. I would like to avoid that as much as possible.

Ah, I looked at the same section but wasn't aware that `importlib` would imply code changes, but yeah, when testing I saw that relative imports  won't work anymore.  What a fucking mess :-/  I really don't look forward to tracing file names manually...  The only other option I can think of to use individual pytest invokations per subdir - which might have some granularity advantages, too.  If you don't like that either I won't hold up that PR...

iparask:
> Ah, I looked at the same section but wasn't aware that `importlib` would imply code changes, but yeah, when testing I saw that relative imports won't work anymore. 

Yeah...

>The only other option I can think of to use individual pytest invokations per subdir - which might have some granularity advantages, too. If you don't like that either I won't hold up that PR...

I think that would be too much. Making sure there is a `__init__.py` per folder similar to a package seems to solve the issue for now. I would merge this one and dig a bit to see what we can do to solve this permanently. I will probably suggest a change in the testing guidelines :)

Thank you @andre-merzky

PR #2245, opened 2020-11-02
Hotfix/demunch

andre-merzky:
ensure that entities are demunch'ed before pushing to MongoDB, which will otherwise barf in some cases.


src/radical/pilot/db/database.py
```diff
@@ -73,7 +73,7 @@ def __init__(self, sid, dburl, cfg, log, connect=True):
             self._c.insert({'type'      : 'session',
                             '_id'       : sid,
                             'uid'       : sid,
-                            'cfg'       : copy.deepcopy(cfg),
+                            'cfg'       : ru.demunch(cfg.as_dict()),
```
mtitov, line 76: I would propose to keep one operation (not to have 2 cycles of demunching): `ru.demunch(cfg)`

andre-merzky: Yes, that makes sense of course -  but then we can also use `as_dict()` I guess as that is more informative in this code when the user is not aware of the munch mechanism in the first place.  Thanks, will fix.

mtitov: just we always have to be sure that `cfg` is `Config` object, that's why I thought to use `ru.demunch`, which is also aligned with other places it applied. But both works

mtitov (approved):
LGTM

src/radical/pilot/db/database.py
andre-merzky: fair point :-)

PR #2250, opened 2020-11-07
support for custom task and pilot IDs

andre-merzky:
This addresses part `A` from radical-cybertools/radical.entk/issues/500


src/radical/pilot/compute_pilot.py
```diff
@@ -75,6 +73,14 @@ def __init__(self, pmgr, descr):
         # pilot failures can trigger app termination
         self._exit_on_error = self._descr.get('exit_on_error')
 
+        # ensure uid is unique
+        if self._uid:
+            if not self._umgr.check_uid(self._uid):
```
iparask, line 78: Does the pilot have a `_umgr` object? Pylint objects and I think it is correct:
```
78   19   E1101   Instance of 'ComputePilot' has no '_umgr' member; maybe '_pmgr'? (no-member)
```

iparask: Also, why do we care about the Unit manager to verify the pilot's ID and not the pilot manager?

iparask (changes requested):
There is a comment that I think it is a bug.

I would also suggest to include some tests with this PR. It make a significant change and it would be nice to have a small test. Also `pylint` objects for a lot of things.

src/radical/pilot/compute_pilot.py
andre-merzky: Yeah, that's a cut'n'paste error, should be the pmgr of course, thanks!

andre-merzky:
@iparask @mturilli : out of the 37 linter complaints, 36 were W0212 (private access) which I cannot sensibly fix, and one was useful.  I almost missed that one because of the W0212 noise.  I would like to raise that issue again to silence that one globally as to make the other messages more useful.

iparask:
I do not have a strong opinion on `protected_access`. I do not like it, but since it is there and cannot be fixed easily, let's mute it. However this PR does not have a test still. Can you add them?

andre-merzky:
Yes, that's still on my table :)

iparask (changes requested):
Can we add tests?

andre-merzky:
@iparask : can I ask for your help?  I added unit tests for the task and pilot UID settings.  For the task it works ok, for the pilot I am stumbling over a failing mock:

```py
        self._pilot_jsurl, self._pilot_jshop \
                               = self._session._get_jsurl           (pilot)
```

The session is mocked, but that mocked method only returns a single `Mock` instance, so the unpack fails.  How would I change the session mock to achieve that, any idea?  Thanks!

iparask:
> The session is mocked, but that mocked method only returns a single `Mock` instance, so the unpack fails. How would I change the session mock to achieve that, any idea? Thanks!

Yes! When you mock an object, its instances are mocked. Any method or attribute that the object may have does not exist and cannot be called. What I do is to create a mock object and then mock the methods which are called. Let me do them and you will see how it goes.

andre-merzky:
> What I do is to create a mock object and then mock the methods which are called. Let me do them and you will see how it goes.

Thanks!

tests/test_pilot/test_pilot.py
```diff
@@ -0,0 +1,61 @@
...
+        pmgr._prof    = mock.Mock()
+        pmgr._session = mock.Mock()
+        pmgr._session.uid = str(time.time())  # restart uid counter
+        sandbox_url = mock.Mock()
+        sandbox_url.path = './'
+        pmgr._session._get_jsurl = mock.Mock(return_value=('ssh', sandbox_url))
+        pmgr._session._get_resource_sandbox = mock.Mock(return_value=sandbox_url)
+        pmgr._session._get_session_sandbox = mock.Mock(return_value=sandbox_url)
+        pmgr._session._get_pilot_sandbox = mock.Mock(return_value=sandbox_url)
+        pmgr._session._get_client_sandbox = mock.Mock(return_value=sandbox_url)
```
iparask, line 31: When the pilot is initialized, it calls these methods from the session. Since the session is mocked, the methods are not defined and cannot be called. As a result, the test will fail for every method call with an error from `mock`.

Mocking the method calls will allow the test to not fail, but also allows us to provide different type of `return_values`, via a `side_effect` based on what we want to test.

PR #2254, opened 2020-11-16
PRRTE multi-DVM

mtitov:


mtitov:
@andre-merzky should I consider updates from `project/covid-19` branch here (particularly changes in `bootstrap_0.sh` related to conda env)? and should `agent_default_sa.json` be here? (it was in configs before, but not now)

src/radical/pilot/agent/scheduler/continuous.py
```diff
@@ -381,15 +381,32 @@ def schedule_unit(self, unit):
         is_last  = False
         tag      = cud.get('tag')
 
+        # - PRRTE related - start -
+        dvm_hosts_list = self._rm_lm_info.get('dvm_hosts') or []
+        # use tag as a DVM ID, so task will be allocated to nodes that are
+        # handled by a corresponding DVM (only applied with `dvm_hosts_list`)
+        # FIXME: new coming attribute `tags` would consider `dvm_id`
+        if dvm_hosts_list and tag is not None and tag not in self._tag_history:
```
andre-merzky, line 389: Why is the tag converted to int only when it's not in the tag history?  Shouldn't it *always* be converted?  I also would suggest to move the tag_history entry to lower, so that we have a single place where the history is changed. (see comment there). 

I would suggest for here:
```py
if dvm_host_list and tag:
    try:
        tag = int(tag)
        if len(dvm_hosts_list) <= tag:
                raise ValueError('dvm_id (%s) out of range' % tag)
    except (TypeError, ValueError):
        tag = None
````

File: src/radical/pilot/agent/scheduler/continuous.py [hunk 2]
```diff
@@ -381,15 +381,32 @@ def schedule_unit(self, unit):
...
         alc_slots = list()
         rem_slots = req_slots
 
         # start the search
         for node in self._iterate_nodes():
 
-            node_uid  = node['uid']
-          # node_name = node['name']
+          # node_uid  = node['uid']
+            node_name = node['name']
```
andre-merzky, line 409: Oh, why was that switched from `node_uid` to `node_name`?  Specifically for localhost and debug modes, node names are not necessarily unique, but the UIDs are.  Am I missing something?

src/radical/pilot/agent/scheduler/continuous.py
```diff
@@ -451,6 +483,11 @@ def schedule_unit(self, unit):
                 # try next node
                 continue
 
+            # - PRRTE related - start -
+            if node_dvm_id is not None and unit_dvm_id is None:
```
andre-merzky, line 487: Since IDs never can be empty, this (and elsewhere) could be:
```py
if node_dvm_id and not unit_dvm_id:
```
?

andre-merzky (approved):
Who is always complaining about late reviews?  Me.
Who is always late with reviews?  Also me.
Go figure :-/

Apart from minor nitpicks, this looks all good to me.  We'll have to touch the code soon on merging `fix/tags`, but that can't be avoided.

src/radical/pilot/agent/scheduler/continuous.py
```diff
@@ -479,11 +516,10 @@ def schedule_unit(self, unit):
                  'lm_info'       : self._rm_lm_info,
                 }
 
-
         # allocation worked!  If the unit was tagged, store the node IDs for
         # this tag, so that later units can reuse that information
-        if tag:
-            self._tag_history[tag] = [node['uid'] for node in slots['nodes']]
+        if not dvm_hosts_list and tag is not None:
+            self._tag_history[tag] = [node['name'] for node in slots['nodes']]
```
andre-merzky, line 522: I would suggest to cover both cases here:
```py
if tag:
    if dvm_host_list:
        self._tag_history[tag] = dvm_hosts_list[tag]
    else: 
        self._tag_history[tag] = [node['name'] for node in slots['nodes']]
````

File: src/radical/pilot/agent/scheduler/continuous.py [hunk 5]
```diff
@@ -381,15 +381,32 @@ def schedule_unit(self, unit):
         is_last  = False
         tag      = cud.get('tag')
 
+        # - PRRTE related - start -
+        dvm_hosts_list = self._rm_lm_info.get('dvm_hosts') or []
```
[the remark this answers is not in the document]
mtitov: this one as a small "protection" for an incorrect empty value, in case value for `self._rm_lm_info['dvm_hosts']` could be as `None` or `''`. Kinda I like to be sure that a certain key is in checked dictionary and the corresponding value is a non-empty, otherwise I use default value on my own (empty list in this case)

src/radical/pilot/agent/launch_method/prte.py
```diff
@@ -27,173 +32,198 @@ def __init__(self, name, cfg, session):
...
+
+            dvm_uri_list.append(dvm_uri)
+
+        lm_info = {'dvm_uri'     : dvm_uri_list,
+                   'dvm_hosts'   : dvm_hosts_list,
                    'version_info': prte_info,
-                   'cvd_id_mode' : 'physical'
-                  }
+                   'cvd_id_mode' : 'physical'}
+        time.sleep(10)
```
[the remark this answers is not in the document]
mtitov: > allow the DVM to stabilize

yes, will add it

src/radical/pilot/agent/scheduler/continuous.py
mtitov: tricky, considering that the value could be `0`, that's why I forced to check everywhere that `None` is a not-set value, otherwise value is set

andre-merzky: Ah, right  - you use integers...  Sorry, I hade `xyz,000` type IDs in mind which we use otherwise...

src/radical/pilot/agent/scheduler/continuous.py
mtitov: > Why is the tag converted to int only when it's not in the tag history? Shouldn't it always be converted?

oh, absolutely valid, somehow missed that, yes, should be converted before being checked

and couple responses on other points:
- if `tag` is set and we have non-empty `dvm_host_list`, which means that DVMs will be set, then `tag` is considered not as a label for host(s) but as an ID for DVM. And tag should be added to the `self._tag_history` before going through the list of available hosts. I would think that after it will migrate to `tags` it will be with a valid name `dvm_id`;
- speaking about your example: I did separate tag/dvm_id value as if it is non-numeric then we just ignore it, but if it is a numeric value then it should be within a range of DVM IDs, thus if this `len(dvm_hosts_list) <= tag` is valid then exception should be raised

src/radical/pilot/agent/scheduler/continuous.py
mtitov: > Specifically for localhost and debug modes

yeah, I didn't think about that... then both values should be used - one for DVM and another for other cases. `node_name` is needed for DVM case, since `dvm_host_list` contains a list of host names

src/radical/pilot/agent/scheduler/continuous.py
mtitov: here we deal with `tag` case that is not related to DVM, that's why cannot combine both of them here. When they all will be in `tags` this wouldn't be a case

src/radical/pilot/agent/scheduler/continuous.py
andre-merzky: The tags PR is close to being merged, too, so that all will soon cleanup anyway.  Thanks for clarifying! :-)

src/radical/pilot/agent/scheduler/continuous.py
andre-merzky: Wouldn't it make sense to then use `node_id` in the host list, and to look up the `node_name` as needed?

src/radical/pilot/agent/scheduler/continuous.py
andre-merzky: ack

src/radical/pilot/agent/scheduler/continuous.py
mtitov: I'll update `dvm_host_list` then, so I'll keep `node_id`s there. Just at the beginning I've started to store host names there (names we put into `prrte.hosts` file), but since RP handles that then to keep ids is valid as well

mtitov: updated

andre-merzky:
@mtitov : once you are through with the comments, I think this is good to merge.  Thanks a bunch for your patience!

PR #2255, opened 2020-11-16
Empty env var NODE_LFS_PATH

mtitov:
If task description has `None` value in `path` for `lfs_per_node` (e.g., `unit.000000.sl` has the following `'lfs_per_node': {'path': None, 'size': 0},`) and then exec script has the following line `export "NODE_LFS_PATH=None"`. Thus `NODE_LFS_PATH` should be set only if there is a path

### DIFF
--- a/tests/test_scheduler/test_unit/test_base.py
+++ b/tests/test_scheduler/test_unit/test_base.py
@@ -78,12 +78,8 @@ class TestBase(TestCase):
             self.assertEqual(unit['slots'], result['slots'])
 
             # test environment variable(s)
-            if 'NODE_LFS_PATH' not in result['description']['environment']:
-                self.assertNotIn('NODE_LFS_PATH',
-                                 unit['description']['environment'])
-            else:
-                self.assertEqual(unit['description']['environment'],
-                                 result['description']['environment'])
+            self.assertEqual(unit['description']['environment'],
+                             result['description']['environment'])
 
 
     # ------------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


tests/test_scheduler/test_unit/test_base.py
```diff
@@ -76,18 +65,25 @@ def test_try_allocation(self, mocked_init, mocked_schedule_unit, mocked_handle_c
...
+            unit = input_data['unit']
+            component._try_allocation(unit=unit)
+
+            # test unit's slots
+            self.assertEqual(unit['slots'], result['slots'])
+
+            # test environment variable(s)
+            if 'NODE_LFS_PATH' not in result['description']['environment']:
+                self.assertNotIn('NODE_LFS_PATH',
+                                 unit['description']['environment'])
```
iparask, line 83: I think that the following `assertEqual` will fail if `NODE_LFS_PATH` is not in the result and is in the unit's environment. Am I missing something?

Having said that I am perfectly fine with leaving it.

iparask (approved):

tests/test_scheduler/test_unit/test_base.py
mtitov: that's correct, it should, since in `result` is our expectations, which should reflect the state of `unit`. There are 2 test cases, in one of them this env variable is expected to be set and in another one this env variable should not be in unit's `environment`

iparask:
@mtitov, do you mind syncing this PR with devel? Let's see how the codecov diff changes. I think it will be positive

tests/test_scheduler/test_unit/test_base.py
iparask: If that is the case, does removing it make sense to you?

mtitov: sorry, can you explain more what should be removed?

iparask: I mean the `if... else` statement. The check in line 85 should be more than enough, right?

mtitov: yeah, you are right, the next one `assertEqual` duplicates this one, I thought to emphasize this env variable check, but the following assert would be enough. Will update it accordingly

PR #2260, opened 2020-11-20
Introduce safe branches

iparask:
This PR introduces what is called safelist. The safe list is a set of branches that travis will build as soon as there is a push. This will help  us have the correct coverage diff from codecov.

The change was triggered by PR #2255. The PR introduces a test for the change it makes, but codecov still shows negative delta in coverage, where it should be 0 change or minimal positive change. 

Thank you!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,11 @@ dist:     xenial
 git:
   depth: 1
 
+# safelist
+branches:
+  only:
+  - devel
+
 env:
   global:
     - CODECOV_TOKEN="c7e0d993-d31a-4d39-8ffa-07c30ea73e48"


iparask:
Same in this PR. I touched travis.yml and there is a negative codecov without reason

andre-merzky:
I am glad that you are getting annoyed by that, too :-)  What is the effect of that save list, and do you understand why #2255 results in negative coverage change?

iparask:
> I am glad that you are getting annoyed by that, too :-) 

I am annoyed because it shows wrong information. 

> What is the effect of that save list, and 

The safe list allows us to define branches that will be built every time there is a push in them and not any other, after we enable `build pushed branches` ([here](https://docs.travis-ci.com/user/web-ui/#build-pushed-branches)) setting on travis. As a result, with the safgelist and the setting enabled, every time there is a push in devel it will be built updating codecov also.

>... do you understand why #2255 results in negative coverage change?
The devel branch is not build on Travis every time a push is performed into it. There is not even a cron job to build every now and then.  As a result a pull request is compared to the last time that the devel branch was built on Travis and codecov was updated. I also similar behavior on EnTK where a PR made a small impact in coverage and Codecov reported something bigger.

Does this help?

mtitov (approved):

Issue #43, opened 2014-01-24
Agent Support for Cray (Archer)


oleweidner:
We need to implement proper single-core task execution support for Cray systems, specifically Archer.

PR #2263, opened 2020-11-24
Feature/blocked cores

andre-merzky:
This PR adds a `blocked_cores` key to the resource config to exclude cores from the scheduler's visibility.


src/radical/pilot/agent/scheduler/continuous.py
```diff
@@ -129,6 +129,27 @@ def _configure(self):
...
+                for idx in blocked_cores:
+                    assert(len(node_entry['cores']) > idx)
+                    node_entry['cores'][idx] = rpc.DOWN
+                    self._rm_cores_per_node -= 1
+
+            blocked_gpus = self._cfg.resource_cfg.blocked_gpus
+            if blocked_gpus:
+
+                self._log.info('blocked gpus : %s' % block_gpus)
+                for idx in block_gpus:
```
mtitov, line 147: small typo (x2): `block_gpus` -> `blocked_gpus`

mtitov:
I like this new introduction! BTW should we extend it to the Summit case as well? (so in resource config, id of a physical core will be set, and in scheduler the code with `smt` will be reused)

src/radical/pilot/agent/scheduler/continuous.py
andre-merzky: Duh!  Fixed.

andre-merzky:
> I like this new introduction! BTW should we extend it to the Summit case as well? (so in resource config, id of a physical core will be set, and in scheduler the code with `smt` will be reused)
> 
> **edit** for Summit it also says that 1 core at **each** socket is reserved and not for allocation through `jsrun` ([user guide](https://docs.olcf.ornl.gov/systems/summit_user_guide.html#compute-node-description))

Yes, I think that would make sense.  Would you mind adding that?  Thanks!

andre-merzky:
@iparask : any suggestion how we should handle this test error:
```py
            blocked_cores = self._cfg.resource_cfg.blocked_cores
            if blocked_cores:
                self._log.info('blocked cores: %s' % blocked_cores)
>               for idx in blocked_cores:
E               TypeError: 'Mock' object is not iterable
```
I think self._cfg. is mocked, so where do we get the blocked cores from?

mtitov:
I have something to discuss about cores (including blocked ones) on Summit:

- each node has 2 sockets with 22 physical cores each
- 1 core on each socket is not available for allocation through jsrun
   - Summit provides 42 cores (set through env variable LSB_DJOB_HOSTFILE)
   - core ids: 0-20,22-42 (thus `#21` and `#43` are removed)

Thus, we deal with all available cores per node, but their sequence of ids is "broken" in the middle.
Should we not just read information about cores from Summit, but set cores_per_node on our side as well (what we currently don't do)? Then resource config would have 44 cores per node and cores `#21` and `#43` blocked?

andre-merzky:
> I have something to discuss about cores (including blocked ones) on Summit:
> 
>     * each node has 2 sockets with 22 physical cores each
> 
>     * 1 core on each socket is not available for allocation through jsrun
>       
>       * Summit provides 42 cores (set through env variable LSB_DJOB_HOSTFILE)
>       * core ids: 0-20,22-42 (thus `#21` and `#43` are removed)
> 
> 
> Thus, we deal with all available cores per node, but their sequence of ids is "broken" in the middle.
> Should we not just read information about cores from Summit, but set cores_per_node on our side as well (what we currently don't do)? Then resource config would have 44 cores per node and cores `#21` and `#43` blocked?

Well, that sounds reasonable, but I'd rather not fix what isn't broken.  In other words, I am not sure of the side effects of that change.  Like, if we set cpn, then the pilot launcher will pick it up and use it to talk to SAGA.  Will SAGA still be able to get nodes with that cpn value even if LSF knows that that many cores are not usable?  And there are maybe other places where cpn is used - we would need to check those first.

mtitov:
@andre-merzky thank you! yeah, agree (all that should be checked). Then I'll create a separate ticket for this (will move our comments there) and the current PR we'll keep as it is. Do you want me to check tests or you are on them?

andre-merzky:
Ack and thanks on the ticket.

> Do you want me to check tests or you are on them?

If you find the time to do them, that would be great - otherwise np, it's on my table (somewhere) :-)

mtitov (approved):

PR #2265, opened 2020-11-24
Feature/env isolation

andre-merzky:
This is an early part of the env isolation effort, but can be already merged as it contains mostly documentation and self-contained cleanup code.


docs/architecture/env_isolation/README.md
```diff
@@ -0,0 +1,48 @@
...
+       - can be cached: identical command sets are run only once, env is then
+         reused for other tasks
+     - `pre_exec_cmd`: any action needed to run the task
+       - runs on the compute node
+       - will block all task resources apart from those assigned to rank 0
+       - runs in the task env (but not in the MPI env)
+     - backward compatibility: `pre_exec` is interpreted as `pre_exec_env`
+   - each LM has it's individual env settings (module load etc)
+     - any number of LMs can be configured for an agent
+     - TODO: clarify how a LM is selected for a task
```
iparask, line 22: Will this replace the CU env parameter?

iparask (commented):
This is a big PR. Can you add a way to run it and see everything working?

I left some questions and suggestions.

src/radical/pilot/configs/resource_xsede.json
```diff
@@ -231,16 +231,18 @@
...
         },
+        "cores_per_node"              : 68,
         "default_queue"               : "normal",
         "resource_manager"            : "SLURM",
         "agent_scheduler"             : "CONTINUOUS",
         "agent_spawner"               : "POPEN",
         "agent_launch_method"         : "IBRUN",
         "task_launch_method"          : "IBRUN",
         "mpi_launch_method"           : "IBRUN",
+      # "blocked_cores"               : [64, 65, 66, 67],
```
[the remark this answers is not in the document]
andre-merzky: This crept in from another PR (#2263), and was an attempt to support gromacs placement by blocking cores.

docs/architecture/env_isolation/00_env_isolation_utils.py
```diff
@@ -0,0 +1,202 @@
...
+    # defined in the current env
+    if  remove is None:
+        remove = os.environ
+
+    # empty pre_exec_env settings are ok - just ensure correct type
+    pre_exec_env = ru.as_list(pre_exec_env)
+
+    # cache lookup
+    cache_key = str(base) + str(remove) + str(pre_exec_env)
+    if cache_key in _env_cache:
```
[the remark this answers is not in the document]
andre-merzky: Basically yes.  We could do a checksum on that string to make it shorter - but that is exactly what the dict hashing algorithm does anyway...  We could replace this with an UID or name - but the effect would be the same.

docs/architecture/env_isolation/README.md
andre-merzky: No, it will augment it.  The goal is to avoid repeated `conda activate` calls for many small tasks at scale which kneels the file system, and also to make sure that the pilot env and modules do not creep into the task environments.

iparask: Ah, I see. I would love an example, but do not let that keep the PR.

iparask (approved):

PR #2266, opened 2020-11-24
Feature/raptor mpi

andre-merzky:
add stub for mpi executor in raptor.  This code is not yet active, but I want to merge to reduce number of branches.

### DIFF
--- a/src/radical/pilot/task_overlay/worker.py
+++ b/src/radical/pilot/task_overlay/worker.py
@@ -219,6 +219,29 @@ class Worker(rpu.Component):
 
     # --------------------------------------------------------------------------
     #
+    def _mpi(self, data):
+        '''
+        launch an rp.MPI_LAUNCH_METHOD for the task
+        '''
+
+        try:
+            cmd = rp.agent.launch_method.construct_command(data,
+                    executable=self.exe, args=data['func'])
+            out = rp.sh_callout(cmd)
+            err = None
+            ret = 0
+
+        except Exception as e:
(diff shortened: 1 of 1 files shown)


src/radical/pilot/task_overlay/worker.py
```diff
@@ -217,6 +217,29 @@ def _call(self, data):
...
 
+    # --------------------------------------------------------------------------
+    #
+    def _mpi(self, data):
+        '''
+        launch an rp.MPI_LAUNCH_METHOD for the task
+        '''
+
+        try:
+            cmd = rp.agent.launch_method.construct_command(data,
```
iparask, line 228: Instead of using the whole package, why not just import what is needed?

Based on the linters, `rp` does not exist, creating a bug.

iparask (changes requested):
This is what we were discussing, right? 

I would like to see a test for this or at least why it cannot be tested. Also, the linters are showing an important bug.

Thank you

src/radical/pilot/task_overlay/worker.py
andre-merzky: > I would like to see a test for this or at least why it cannot be tested. Also, the linters are showing an important bug.

It can't be tested as it does not work yet.  I converted the PR into a comment for now.  This is really just about closing that branch for the time being as we won't have development on this for a while...

iparask: Okay, then let's move forward.

iparask (approved):

PR #2267, opened 2020-11-24
Fix/mpiexec mpt

andre-merzky:
This branches fixes the `mpiexec_mpt` behavior for Cheyenne (which is the only machine which uses that launcher).  It has been confirmed to work by the user requesting the fix.


src/radical/pilot/agent/launch_method/mpiexec.py
```diff
@@ -97,24 +120,38 @@ def construct_command(self, cu, launch_script_hop):
...
+                host_string += '-host '
+                host_string += '%s -n %s ' % (','.join([node] * nslots), nslots)
+        else:
+            slots = 0
+            hosts = list()
+            for node,nslots in list(host_slots.items()):
+                slots +=          nslots
+                hosts += [node] * nslots
+            host_string += ','.join(hosts)
+            host_string += ' -n 1'
```
mtitov, line 142: 1) Is variable `slots` needed here? (also variable `slots` already exists, even it is not used later, and this operation reassigns it);
2) `host_string` here doesn't need `-host`, right? (didn't find about it in the documentation..)

mtitov:
I've added a couple of tests and besides the comments above LGTM

src/radical/pilot/agent/launch_method/mpiexec.py
andre-merzky: 1: yeah, you are right - removed
2: correct, the MPT version doesn't like `-host` ...

andre-merzky:
Thanks for adding the test cases!

PR #2269, opened 2020-11-24
Fix/sa init

andre-merzky:
This branch has mostly been merged in a `hotfix`, but a small patch remained unmerged.  That introduces an `output` method on the component base class which, as the name suggests, allows a component to push something on an output queue.  That was always possible - the method just formalizes that.


iparask (changes requested):
I think a small test for `output` will be welcome!

andre-merzky:
> I think a small test for `output` will be welcome!

Do you have an suggestion on what that test would look like?  If you have the time, would you mind providing one?

iparask:
> > I think a small test for `output` will be welcome!
> 
> Do you have a suggestion on what that test would look like? If you have the time, would you mind providing one?

I just pushed a small test. It misses a positive one, where something is put in the `_outputs` dictionary because I do not know the data type of its values. Does it help?

andre-merzky:
> I just pushed a small test. It misses a positive one, where something is put in the `_outputs` dictionary because I do not know the data type of its values. Does it help?

Thanks, that works for me.  I added some test cases.

tests/test_components/test_component.py
```diff
@@ -0,0 +1,47 @@
...
+
+        component._outputs = {'test_state':[]}
+
+        with self.assertRaises(ValueError):
+            component.output(things='test_thing')
+
+        with self.assertRaises(ValueError):
+            component.output(things='test_thing', state='test_state2')
+
+        component.output(things=None,  state='test_state')
```
iparask, line 42: How do you expect `component._output` to look like after executing this line? After every call, an assert can be evaluated.

For example:
```
expected_output = ...
self.assertEqual(component._output, expected_output)
```

andre-merzky: `_outputs` is a list of ZMQ endpoints.  The endpoints are stateless and don't change after the send.  We can't really check the entities unless we actually create ZMQ endpoints, so I think this is ok as is?

iparask: It makes sense.

iparask:
For some reason, all the tests fail after the cleanup. Let me fix that.

iparask (approved):
I think this PR is ready!

andre-merzky:
What is the effect of enabling protected access (and others) at the end of the files?  I thought the scope for those instructions is limited by the file anyway?

mturilli:
Can we merge this?

andre-merzky:
> Can we merge this?

yes :-)

PR #2271, opened 2020-11-24
Raptor fixes (gb)

andre-merzky:
scattered fixes to raptor


src/radical/pilot/task_overlay/master.py
```diff
@@ -227,23 +229,31 @@ def wait(self, count=None, uids=None):
         if count:
             self._log.debug('wait for %d workers', count)
             while True:
+                stats = {
```
mtitov, line 232: just a stylistic comment - should we call it `statuses`? `stats` is referred to as `statistics`

src/radical/pilot/task_overlay/worker.py
```diff
@@ -467,10 +500,18 @@ def _dispatch(self, task):
 
         # ----------------------------------------------------------------------
         def _dispatch_thread(tlock):
+            # FIXME: do we still need this thread?
+
+            os.environ['RP_TASK_CORES'] = ','.join(str(i) for i in task['resources']['cores'])
+            os.environ['RP_TASK_GPUS']  = ','.join(str(i) for i in task['resources']['gpus'])
```
mtitov, line 506: with these two env variables, it is not clear why they are introduced (one of them is used inside in this method as well and nowhere else)

mtitov:
I also would like to propose to update inputs for method `submit` in `rp.task_overlay.master.Master` ([link](https://github.com/radical-cybertools/radical.pilot/blob/4b98e23e8e821385707c85d87355299f0cbf4238/src/radical/pilot/task_overlay/master.py#L161)): 
to have `def submit(self, descr)` and all other parameters are supposed to be part of `descr`
```python
...
        descr['cpu_processes']    = count
        descr['cpu_threads']      = cores
        descr['gpu_processses']   = gpus
        master.submit(descr=descr)
...
```

iparask:
Can we add a test?

src/radical/pilot/task_overlay/master.py
andre-merzky: Ha, the `stats` is about gathering crude statistics: how many workers found in what status :-)

src/radical/pilot/task_overlay/worker.py
andre-merzky: Yeah, we probably can remove them.  I used them for testing to ensure that the tasks land on different cores, but otherwise they are likely not useful.

andre-merzky:
> Can we add a test?

This is a fair request, but not easy to answer.  The PR touches small pieces of code in Raptor all over the place -- but raptor has no test coverage whatsoever.  So yeah, maybe we should start just that in this PR, but it will take a bit I assume...

src/radical/pilot/task_overlay/master.py
mtitov: ah... right!

andre-merzky:
@mtitov : sorry for the late reply on the `submit` signature.  If you don't mind I'd rather leave this for now - that signature (and in fact that method) will change soon anyway when we integrate raptor into the executor component...

mtitov:
with the latest changes raptor's example (`radical.pilot/examples/misc/task_overlay_*`) has some misses:
- `task_overlay_master.py`: `master.run` -> `master.start` ([link](https://github.com/radical-cybertools/radical.pilot/blob/project/gb/examples/misc/task_overlay_master.py#L145))
- `task_overlay_worker.py`: `self.register_call` -> `self.register_mode` ([link](https://github.com/radical-cybertools/radical.pilot/blob/project/gb/examples/misc/task_overlay_worker.py#L25))

PR #2274, opened 2020-11-27
Pilot RPC

andre-merzky:
This PR implements a pilot RPC call, like this:

```py
        print(pilot.rpc(rpc='hello', args=['world']))
```

which prints:

```sh
hello world
```

The main purpose is to support a use case in the ScaleMS project, where a request to prepare a specific task environment is to be sent to a pilot, and the application needs confirmation that the request succeeded (the existing `prepare_env` call on the pilot does not provide that feedback - it would only fail any later submitted tasks which happen to request an invalid env).

The PR also removes the special pilot staging directory `staging_area`, and instead stages to the pilot sandbox directly.  That change crept into this branch prematurely, and will arrive in devel via #2201 -- sorry for that.

@mtitov , @iparask : I would appreciate help with writing tests for this.  The individual methods are fairly simple - but the whole systems is a chain of DB communications, ZMQ messages and callbacks, so testing is more about integration of those parts than the individual calls.  How should we handle this?


andre-merzky:
PS.: not sure about the remaining linter comments - pylint seems to ignore the respective `noqa` directives?

src/radical/pilot/task_overlay/worker.py
```diff
@@ -31,7 +31,7 @@ def __init__(self, cfg, session=None):
         # FIXME: this should be delegated to ru.generate_id
 
         # FIXME: why do we need to import `os` again after MPI Spawn?
-        import os
+        import os                                                         # noqa
```
iparask, line 34: Line 2 imports os. There is no need to import it here.

src/radical/pilot/task_overlay/worker.py
```diff
@@ -254,7 +239,7 @@ def _eval(self, data):
         '''
 
         try:
-            out = eval(data['code'])
+            out = eval(data['code'])                                      # noqa
```
iparask, line 242: If you add `# pylint:disable= eval-used` and `# pylint:enable= eval-used` after this line the linter will silence.

src/radical/pilot/db/database.py
```diff
@@ -259,12 +259,71 @@ def pilot_command(self, cmd, arg=None, pids=None):
             raise RuntimeError ('pymongo error: %s' % e.details) from e
 
 
+    # --------------------------------------------------------------------------
+    #
+    def pilot_rpc(self, pid, rpc, args=None):
```
iparask, line 264: Can we add a test for this method?

iparask (changes requested):

src/radical/pilot/task_overlay/worker.py
andre-merzky: For some reason,  the `os` module was not in the namespace after fork/exec.  Don't ask, but thus the import and the comment...

src/radical/pilot/task_overlay/worker.py
andre-merzky: Any idea why `noqa` does not work?   I am not sure if we should add two lines of linter control for one line of correct code.  Maybe disable that one for the file then?

src/radical/pilot/db/database.py
andre-merzky: Should that be an integration test then?    The logic and mechanics of the method is all around pushing something to MongoDB and waiting for a reply to pop up.  If MongoDB is mocked away, there is not much to test.  Should I convert some RPC example to a test, would that work in your opinion?

src/radical/pilot/agent/agent_0.py
```diff
@@ -532,13 +535,134 @@ def _check_commands(self):
...
+        # control channel
+        self.publish(rpc.CONTROL_PUBSUB, {'cmd' : 'rpc_req',
+                                          'arg' : rpc_req})
+
+        return True  # keeb cb registered (self._check_rpc)
+
+
+    # --------------------------------------------------------------------------
+    #
+    def _check_control(self, _, msg):
```
iparask, line 611: Would you mind if I try to create a unit test for this method? I think I have an idea of how to do it.

andre-merzky: > Would you mind if I try to create a unit test for this method?

Would I mind?   WOULD *I* MIND?  I thought you'd knew me better than that... :-P

src/radical/pilot/db/database.py
andre-merzky: Makes sense IMHO.

andre-merzky: That test has been added.

src/radical/pilot/task_overlay/worker.py
andre-merzky: should be fixed...

src/radical/pilot/agent/agent_0.py
iparask: Done

tests/test_agent_0/test_unit/test_agent_0.py
```diff
@@ -0,0 +1,86 @@
...
+                                           {'cmd': 'rpc_res', 
+                                            'arg': {'uid': 'rpc.0000', 
+                                                    'err': "'arg'", 
+                                                    'ret': None}
+                                           }))
+
+        msg = {'cmd': 'rpc_req',
+               'arg': {'uid': 'rpc.0000',
+                       'rpc': 'hello',
+                       'arg': 'World'}
```
andre-merzky, line 64: arg should be a list (`['World']`)

tests/test_agent_0/test_unit/test_agent_0.py
```diff
@@ -0,0 +1,86 @@
...
+              }
+        self.assertTrue(agent_cmp._check_control(None, msg))
+        print(global_control)
```
andre-merzky, line 67: The print can probably go.

tests/test_agent_0/test_unit/test_agent_0.py
```diff
@@ -0,0 +1,86 @@
...
+        self.assertEqual(global_control[1], ('control_pubsub', 
+                                             {'cmd': 'rpc_res', 
+                                              'arg': {'uid': 'rpc.0000', 
+                                                      'err': None, 
+                                                      'ret': 'hello W o r l d'}
```
andre-merzky, line 72: This will become `hello World` with the above change to list.

andre-merzky (commented):
Thanks for adding that test! :-)

andre-merzky:
@iparask : I did some small changes on the test you added.  Please give this PR another review - anything amiss before we can merge?

src/radical/pilot/staging_directives.py
```diff
@@ -199,10 +184,6 @@ def complete_url(path, context, log=None):
         log.debug('   expand with %s', context[schema])
         ret = ru.Url(context[schema])
 
-        if schema in ['resource', 'pilot']:
-            # use a dedicated staging area dir
-            ret.path += '/staging_area'
-
```
iparask, line 205: Does this mean that the pilot sandbox gets renamed? Is it not `session_id/pilot.0000/sandbox` anymore? This is something that goes all over the PR.

iparask (approved):
I'm fine with this PR. It looks okay to me. The linters fail because of archived executing components. @andre-merzky do we want them linted?

If yes, fixing them should be fast. If not, does it make sense to exclude them?

The following line:
```
$(echo $DIFF | grep -o -e '\b[^ ]*.py\b' | sed "s|^|$PWD/|")
```
filters the python files. Should we update it?

mtitov (approved):
LGTM!

andre-merzky:
We should exclude `archive`'d files from linting (we did the same in travis).

src/radical/pilot/staging_directives.py
andre-merzky: Oh no - the pilot sandbox remains as it is.  The special `staging_area/` subdir within the pilot sandbox gets removed though.

iparask: And any data are moved to the same folder with all the log files?

andre-merzky: Yes, indeed.  I realize that this is not optimal, and I suggest to succesively move log, prof and cfg/json files out of the pilot sandbox.  Ideally, the user should be able to inspect the sandbox and mostly recognize workload related data and directories.  But the motivation for this specific change was that it turned out to be difficult to construct relative paths in relation to the `staging_area/`, and the change simplifies sandbox references a lot: `pilot://` refers to the pilot sandbox, `unit://` to the unit sandbox, etc.  This in turn simplifies data management for several projects (scalems, repex, covid).

PR #2277, opened 2020-12-02
Updated pip command in bootstrap script

mtitov:


andre-merzky (approved):
LGTM, thanks

iparask (approved):
I executed a small example on Bridges and it works

Issue #270, opened 2014-08-04
SAGA-Pilot needs to be changed to RADICAL-Pilot -- consistently and globally


shantenujha:
For example see use of "SAGA-Pilot" at:

http://radicalpilot.readthedocs.org/en/latest/examples/remote_hpc_pilots.html

(immediately after 4.4 heading)

PR #2280, opened 2020-12-08
Env variable RADICAL_BASE for a job description

mtitov:
This is related to RU ticket [#270](https://github.com/radical-cybertools/radical.utils/issues/270)


andre-merzky (approved):
Thanks!

PR #2281, opened 2020-12-08
setting RP workflow and templates

iparask:


iparask:
This PR does two things:
1. Sets up Github actions that run our tests
2. Creates bug templates and feature request templates

mturilli (commented):
I iterated on the templates. Codecov highlights missing tests for the PR. Any viable hint to add some test?

iparask:
Yes, but I would say it make sense not to block this one because of missing tests. There are several PRs that are stalling because they are trying to run on Travis.

Would you agree to merge this one to get testing running? I'm picking up RP testing as well.

mturilli (approved):
Absolutely, you are right! Ready to merge for me, thank you.

Issue #515, opened 2015-03-19
check for zero core counts on CU submission


mturilli:
RP: feature/aimes-experiments
SP: master
RU: master

Running a bag of 256 tasks, 64 of which with 2 cores, 64 with 4 cores, 64 with 8 cores, 64 with 16 cores for a total of 1920 cores. I scheduled the tasks with the back filling scheduler on 3 pilots of 640 cores,  submitted respectively to Stampede, Trestles, and Gordon. Gordon was the first pilot coming online and a total of 90 units got scheduled on it. 62 units executed correctly but 4 failed and for 24 no lunch script was created. The error reported printed on the origin STDERR was:
```
'550b009a23769c3b65904b64' stderr: 
Pilot cannot start compute unit:
list index out of range
Traceback (most recent call last):
  File "radical-pilot-agent.py", line 1159, in run
    self._launch_task(task, launch_method, launch_command)
  File "radical-pilot-agent.py", line 1404, in _launch_task
    cu_environment=self.cu_environment)
  File "radical-pilot-agent.py", line 2421, in __init__
    host = task.slots[0].split(':')[0] # Get the host of the first entry in the acquired slot
IndexError: list index out of range

Pilot cannot start compute unit: 'list index out of range'.
'550b009a23769c3b65904b64' stdout: .
``` 
Reviewing my code I spotted an error that made all the CU running on 16 cores. I have now fixed that error and I am waiting for another test to go through the queue. I report the error anyway as I never saw it before and it does not necessarily seem related to the number of cores required by each CU. In case it might be useful, here the graph of the run:
![run-256_late_uniform_1_550b006823769c3b65904b1d](https://cloud.githubusercontent.com/assets/2323492/6740678/60f410d8-ce58-11e4-9462-7dbb67631882.png)

marksantcroos:
Thanks for the report. That is certainly not good.

That code has been quite reshuffled in devel, although the issue might still exist.

Could you make the agent.log available?

mturilli:
Please find the log file at: https://drive.google.com/open?id=0B25uRSKrOw87UEZUMFFGazlET3c&authuser=0

Unfortunately, the log is shared among multiple runs - this is due to RP appending to the log files. I think a safe bet is to start looking at around `2015:03:19 16:43:49` but I am not positive about it.

marksantcroos:
Thanks. However, I meant the agent.log from the pilot directory on gordon.
You still have that one?

[...]

PR #2283, opened 2020-12-08
Remove TMPDIR setup for pip installation

mtitov:
Related to EnTK ticket [#515](https://github.com/radical-cybertools/radical.entk/issues/515)

p.s. just for records: option `--build` was introduced [here](https://github.com/radical-cybertools/radical.pilot/commit/951adf520397033224688a2ccd0e66f44e06edfd#diff-56c4dddccccd0fff0a9907f6d9a54b6aed2124a4724e38f49295e3e79d962e2a), I would think as a standard approach for py2

### DIFF
--- a/src/radical/pilot/agent/bootstrap_0.sh
+++ b/src/radical/pilot/agent/bootstrap_0.sh
@@ -1269,10 +1269,6 @@ rp_install()
   #     fi
   # fi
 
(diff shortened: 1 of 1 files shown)


mturilli (approved):
LGTM. Thanks!

PR #2291, opened 2020-12-15
Introducing simple integration test with Comet

iparask:
This PR introduces an integration test for Comet. It provides the test and utility scripts to open an issue if needed.

I'm also introducing a reorganization of the testing folders. That will allow us to have everything in the same place and also be able to run what is needed with our CI. Let me know if you agree with the changes so that I can go and fix the tests that fail because of reorganizing.

Thank you!


testing_suite/batch_scripts/comet.sh
```diff
@@ -0,0 +1,33 @@
...
+source activate testing
+pip install ./radical.pilot --upgrade
+pytest -vvv $TEST > output.log 2>&1
+
+if test "$?" = 1
+then
+    python radical.pilot/tests/utils/integration_test_issue.py 'SDSC Comet' output.log
+fi
+
+sbatch comet.sh
```
andre-merzky, line 33: We discussed this on the call, and could consider decreasing the test frequency on errors:
```sh
if test "$?" = 1
then
    python radical.pilot/tests/utils/integration_test_issue.py 'SDSC Comet' output.log
    sbatch --begin='now+1month' comet.sh 
else
    sbatch --begin='now+1week' comet.sh 
fi

```

iparask:
I am not sure why coverage falls. :( The same tests are present in devel and this branch. But the reorganization is complete.

andre-merzky (approved):
Thanks @iparask , that looks great.  Would it make sense to move the issue creation scritp from utils to bin maybe?  Either way, approved.

iparask:
> Would it make sense to move the issue creation script from utils to bin maybe?

Do you mean RP's general bin or tests' bin?

andre-merzky:
> > Would it make sense to move the issue creation script from utils to bin maybe?
> 
> Do you mean RP's general bin or tests' bin?

Oh, sorry for the ambiguity - I meant `tests/bin/`.

tests/utils/integration_test_issue.py
```diff
@@ -0,0 +1,18 @@
...
+
+with open(sys.argv[2]) as f:
+    lines = f.readlines()
+
+i = repo.create_issue(
+    title="Integration Tests failed on %s : %s" % (sys.argv[1], date.today()),
+    body='```python\n' + ''.join(x for x in lines) + '\n```',
+    labels=[repo.get_label(name="topic:testing"),
+            repo.get_label(name="type:bug"),
+            repo.get_label(name="layer:rp")])
```
mturilli: Can we get the list of open issues and search for a string in the issue title? If so, we could create a unique ID for each failed-test issue and open a ticket only if that token is not already in any of the issue titles.

iparask: We can. How do you propose to create that unique ID?

tests/integration_tests/batch_scripts/comet.sh
```diff
@@ -0,0 +1,33 @@
...
+
+conda create -p testing python=3.7 pytest PyGithub -y
+
+source activate testing
+pip install ./radical.pilot --upgrade
+pytest -vvv $TEST > output.log 2>&1
+
+if test "$?" = 1
+then
+    tests/bin/integration_test_issue.py -r 'SDSC Comet' -l output.log
```
mtitov, line 29: `tests/bin/radical-pilot-test-issue ...`

tests/integration_tests/README.md
```diff
@@ -0,0 +1,28 @@
+# Integration testing with resources
+
+Integration testing on resources requires periodic jobs. To achieve this, batch scripts are offered under the `batch_scripts` folder. Currently, there is a script for SDSC Comet.
+
+The script defines the following to execute the test:
+
+- A `TEST` environment variable that contains the path to the test files to be executed. All test paths should be included and be space-separated.
+    Example:
+    ```
+    TEST="radical.pilot/tests/test_resources/test_rm/test_slurm.py"
```
mtitov, line 10: `radical.pilot/tests/integration_tests/test_rm/test_slurm.py`

tests/integration_tests/batch_scripts/comet.sh
```diff
@@ -0,0 +1,33 @@
...
+#SBATCH -o rp_integration_test.%j.out   # Name of stdout output file(%j expands to jobId)
+#SBATCH -e rp_integration_test.%j.err   # Name of stderr output file(%j expands to jobId)
+#SBATCH -p compute
+#SBATCH -N 1                # Total number of nodes requested (16 cores/node)
+#SBATCH -n 1                # Total number of mpi tasks requested
+#SBATCH -t 00:30:00         # Run time (hh:mm:ss) - 1.5 hours
+# The next line is required if the user has more than one project
+#SBATCH -A  # Allocation name to charge job against
+
+TEST="radical.pilot/tests/test_resources/test_rm/test_slurm.py"
```
mtitov, line 12: `radical.pilot/tests/integration_tests/test_rm/test_slurm.py`

tests/integration_tests/batch_scripts/comet.sh
iparask: Thanks! I missed those

tests/utils/integration_test_issue.py
mturilli: I would probably look at https://docs.python.org/3/library/uuid.html. We need information about: test ID, status of the ticket:

andre-merzky: Creating some ID is certainly doable - but how do we know that a certain failure belongs to a certain ticket?  Adding the error message to the issue maybe, and parse for that error message in all issues labeled `testing` if that error message is logged?

mturilli: I would assume the error message is in the body of the ticket? But the test does not need to know. Each test just check whether its ID is in the subject of any existing ticket. We do need to know the relationship between test, ID and error but that should be all derivable from the ticket itself: ID in the subject; name/location of the test in the body of the message; error in the body of the message. Was this what you were suggesting?

tests/integration_tests/batch_scripts/comet.sh
```diff
@@ -0,0 +1,33 @@
...
+
+conda create -p testing python=3.7 pytest PyGithub -y
+
+source activate testing
+pip install ./radical.pilot --upgrade
+pytest -vvv $TEST > output.log 2>&1
+
+if test "$?" = 1
+then
+    tests/bin/radical-pilot-test-issue -r 'SDSC Comet' -l output.log
```
andre-merzky, line 29: @iparask : can you help me please to sort this out?  Why is it `radical.pilot/tests/.../test_slurm.py`, but `tests/bin/radical-pilot-test-issue` (without the `radical-pilot/` prefix)?

tests/utils/integration_test_issue.py
iparask: > I would assume the error message is in the body of the ticket? But the test does not need to know. 

I am not sure what you mean here? The tests do not know the state of their previous run. `Pytest` runs multiple tests at a time. 

> Each test just checks whether its ID is in the subject of any existing ticket.

A test may fail for different reasons and it is not necessary that the same failed test will fail with the same error.

I think it may make sense to calculate the checksum of error that is included in the body of an issue, and include that in the issue name. If the failure is the same, then we do not need to report a new issue. Would a ping in the issue make sense? And we can ping a team instead of a person.

tests/integration_tests/batch_scripts/comet.sh
iparask: oops :)

tests/utils/integration_test_issue.py
andre-merzky: We maybe should not overthink this too much, otherwise we spend time on test fingerprinting and issue automation etc.  The goal is integration testing, and opening issues is a nice bonus.  I would suggest we roll with the 1week/4week cycle we have defined now, and we'll see how it goes.  If in 6 months we have 6 identical issues, we can pick this up and think of something better...

mturilli: OK, let's see how it goes then.

mturilli (commented):
Thanks, this looks promising!

PR #2298, opened 2021-01-11
fix node counting for FORK rm

andre-merzky:
The current scheme under-estimated the required number of nodes for certain CPU/GPU configurations.

### DIFF
--- a/src/radical/pilot/agent/agent_0.py
+++ b/src/radical/pilot/agent/agent_0.py
@@ -627,8 +627,9 @@ class Agent_0(rpu.Worker):
             # we don't handle that request
             return True
 
-        rpc_res = {'uid': req['uid']}
+        rpc_res = {'uid': arg['uid']}
         try:
+            print(arg)
             if req == 'hello'   :
                 ret = 'hello %s' % ' '.join(arg['arg'])
 
@@ -639,7 +640,7 @@ class Agent_0(rpu.Worker):
 
         except Exception as e:
             # request failed for some reason - indicate error
-            rpc_res['err'] = str(e)
+            rpc_res['err'] = repr(e)
             rpc_res['ret'] = None
 
         else:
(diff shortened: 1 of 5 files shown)


mtitov (approved):
I've added test for this case and applied a small fix. LGTM

andre-merzky:
I don't get that `agent_0` RPC test fixed...  I am fairly sure the code and test are correct, and it also works locally without any problems...  @mtitov , @iparask : do you have any idea what's up, and what causes that additional comma?

src/radical/pilot/agent/agent_0.py
```diff
@@ -627,19 +627,21 @@ def _check_control(self, _, msg):
...
                 env_id   = arg['arg']['env_id']
                 env_spec = arg['arg']['env_spec']
-                ret = self._prepare_env(env_id, env_spec)
+                self._prepare_env(env_id, env_spec)
+                ret = (env_id, env_spec)
 
         except Exception as e:
             # request failed for some reason - indicate error
-            rpc_res['err'] = str(e)
+            rpc_res['err'] = repr(e)
```
mtitov, line 644: actually, if revert this back and in test revert back to `'err': "'arg'"`, then it works

tests/unit_tests/test_agent_0/test_agent_0.py
```diff
@@ -54,7 +55,7 @@ def _prepare_env_side_effect(env_id, spec):
         self.assertEqual(global_control[0], ('control_pubsub',
                                            {'cmd': 'rpc_res',
                                             'arg': {'uid': 'rpc.0000',
-                                                    'err': "'arg'",
+                                                    'err': "KeyError('arg')",
```
mtitov, line 58: to the comment above (about reverting back)

src/radical/pilot/agent/agent_0.py
andre-merzky: Which means that `repr` adds a comma on the GH test VM, but not otherwise.  Urgh...  Let me look into that!

andre-merzky: https://bugs.python.org/issue30399 - this seems to / should have been fixed in 2017...  Hmm...  I don't really want to revert, returning `repr` (and thus the exception type) is actually the right thing to do...

mtitov: it also says about py3.7 only... (even if it would be fixed for 3.7 we still support 3.6 as well)

PR #2299, opened 2021-01-13
fix bootstrapper timestamps and resource details

andre-merzky:
Somewhere on the way, we lost the alignment of bootstrapper timestamps
in certain conditions.  This hotfix provides:

  - recording of timestamps independent of locale (LC_NUMERIC)
  - time alignment with `sync_abs` (previous `sync_rel`)

The PR also has a small fix which avoids overwriting of resource details
in some cases.

### DIFF
--- a/src/radical/pilot/agent/bootstrap_0.sh
+++ b/src/radical/pilot/agent/bootstrap_0.sh
@@ -3,6 +3,8 @@
 # Unset functions/aliases of commands that will be used during bootstrap as
 # these custom functions can break assumed/expected behavior
 export PS1='#'
+export LC_NUMERIC="en_US.UTF-8"
+
 unset PROMPT_COMMAND
 unset -f cd ls uname pwd date bc cat echo grep
 
@@ -200,6 +202,7 @@ create_gtod()
         test -x '/bin/bash' && shell=/bin/bash
 
         echo "#!$SHELL"                                > ./gtod
+        echo "export LC_NUMERIC=en_US.UTF-8"          >> ./gtod
         echo "if test -z \"\$EPOCHREALTIME\""         >> ./gtod
         echo "then"                                   >> ./gtod
         echo "  awk 'BEGIN {srand(); print srand()}'" >> ./gtod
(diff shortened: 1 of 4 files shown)


src/radical/pilot/agent/bootstrap_0.sh
```diff
@@ -210,8 +213,20 @@ create_gtod()
 
     chmod 0755 ./gtod
 
-    TIME_ZERO=`./gtod`
-    export TIME_ZERO
+    # initialize profile
+    PROFILE="bootstrap_0.prof"
+    now=$(./gtod)
+    echo "#time,name,uid,state,event,msg" > "$PROFILE"
```
mtitov, line 219: update as `#time,event,comp,thread,uid,state,msg`?

src/radical/pilot/agent/bootstrap_0.sh
```diff
@@ -210,8 +213,20 @@ create_gtod()
...
+
+    ip=$(ip addr \
+         | grep 'state UP' -A2 \
+         | grep 'inet' \
+         | awk '{print $2}' \
+         | cut -f1 -d'/')
+    printf "%.4f,%s,%s,%s,%s,%s,%s\n" \
+        "$now" "sync_abs" "bootstrap_0" "MainThread" "$PILOT_ID" \
+        "PMGR_ACTIVE_PENDING" "$(hostname):$ip:$now:$now:$now" \
```
mtitov, line 228: is this required to repeat `$now` 3 times?

mtitov (approved):
couple comments, but overall LGTM

src/radical/pilot/agent/bootstrap_0.sh
andre-merzky: Woah - good catch, that line must have quite a history of blind copy-and-paste... :-P

src/radical/pilot/agent/bootstrap_0.sh
andre-merzky: The `sync_abs` timestamps are normally derived from an NTP query, and for that we include the query send time, query receive time, and estimated time.  Here we basically fake a perfect NTP service - which is usually fine as we have second granularity only anyway.

PR #2300, opened 2021-01-13
Hotfix/tags

andre-merzky:
forward compatibility for tags - see radical-cybertools/radical.entk/issues/429


iparask (approved):
I'm approving of the change since I know it works! I'll try to create a test case for tags.

andre-merzky:
>  I'll try to create a test case for tags.

Thank you!

iparask:
@andre-merzky I added the test case. I'll do more while writing more tests.

andre-merzky:
> @andre-merzky I added the test case. I'll do more while writing more tests.

Is this good to merge, or should we wait for those other tests?

iparask:
I see that one of the tests fails. After that is fixed we can merge and I will add tests during my testing day. :)

andre-merzky:
> I see that one of the tests fails. After that is fixed we can merge and I will add tests during my testing day. :)

This should be fixed now.  I don't see how the test can have worked before though: the unit fed into the scheduler is `unit.000000`, but the test checks if the tag history has been filled for `unit.000001`?

PR #2301, opened 2021-01-15
Feature/resource plots

andre-merzky:
The main contribution of this PR is to add `utils.get_resource_transitions` which inverts a given metric dictionary to derive the events during which resource control shifts from one metric to another.  That method is used for additional resource utilization plots.

Other changes in the PR are related to profiling, but partially overlaps with changes in other branches (gtod integration / fixes), and some scattered fixes / clean-ups to scheduler event recording and profile flushing.


src/radical/pilot/agent/scheduler/base.py
```diff
@@ -748,6 +748,8 @@ def _unschedule_completed(self):
             while not self._proc_term.is_set():
                 unit = self._queue_unsched.get(timeout=0.001)
                 to_unschedule.append(unit)
+                if len(to_unschedule) > 128:
+                    break
```
mturilli, line 753: Both the timeout and the threshold seem arbitrary. Something to try to fix with a different design? Meanwhile, should we expose them via (private) conf parameters?

src/radical/pilot/utils/prof_utils.py
```diff
@@ -53,8 +55,6 @@
     #
     # resources on agent nodes are consumed for all of the pilot's lifetime
     'agent' : {
```
mturilli, line 57: Did we loose `bootstrap_0_start` and `bootstrap_0_stop` in favor of `sub_agent_start` and `sub_agent_stop`? What happens if we have multiple subagents?

src/radical/pilot/utils/prof_utils.py
```diff
@@ -83,10 +83,10 @@
...
       # # if we have cmd_start / cmd_stop:
       # 'exec_sh'     : [{ru.EVENT: 'cu_start'               },
-      #                  {ru.EVENT: 'cmd_start'              }],
-      # 'exec_cmd'    : [{ru.EVENT: 'cmd_start'              },
-      #                  {ru.EVENT: 'cmd_stop'               }],
-      # 'term_sh'     : [{ru.EVENT: 'cmd_stop'               },
+      #                  {ru.EVENT: 'app_start'              }],
+      # 'exec_cmd'    : [{ru.EVENT: 'app_start'              },
+      #                  {ru.EVENT: 'app_stop'               }],
+      # 'term_sh'     : [{ru.EVENT: 'app_stop'               },
```
mturilli, line 89: Why the change in naming? In our lab parlance, `application` refers to the user-facing application while `command` to the executable of a task. While I do not care about our parlance, I care about consistency,

src/radical/pilot/utils/prof_utils.py
```diff
@@ -1236,5 +1236,160 @@ def _get_unit_consumption(session, unit, udurations=None):
...
+    # [
+    #   # event                          from         to
+    #   [{ru.EVENT: 'schedule_ok'    },  None,        'exec_queue'],
+    #   [{ru.STATE: s.AGENT_EXECUTING}, 'exec_queue', 'exec_prep'],
+    #   [{ru.EVENT: 'exec_start'     }, 'exec_prep',  'exec_rp'],
+    #   [{ru.EVENT: 'cu_start'       }, 'exec_rp',    'None]
+    # ]
+    #
+    # which will use like this:
+    #
```
mturilli, line 1274: we will?

src/radical/pilot/utils/prof_utils.py
```diff
@@ -1236,5 +1236,160 @@ def _get_unit_consumption(session, unit, udurations=None):
...
+    #   - go through all events for a task
+    #   - if event is in list above
```
mturilli, line 1276: in the list?

src/radical/pilot/utils/prof_utils.py
```diff
@@ -1236,5 +1236,160 @@ def _get_unit_consumption(session, unit, udurations=None):
...
+    #     - reduce resources in `from` metric
+    #     - increase resource in `to` metric
+    #
+    # If `from` or `to` are None, then the resources are taken from / given to
+    # the pilot's idle resources.  We thus rename the 'None' to 'idle'.
+
```
mturilli, line 1282: Does that assume a 'perfect' profile in which events are not missing? That is a fairly rare event.

src/radical/pilot/utils/prof_utils.py
```diff
@@ -1236,5 +1236,160 @@ def _get_unit_consumption(session, unit, udurations=None):
...
+    # dig though metric,  find all pairs of matching start/stop events
```
mturilli, line 1283: `metric,  find` -> `metric, find`

src/radical/pilot/utils/prof_utils.py
```diff
@@ -1236,5 +1236,160 @@ def _get_unit_consumption(session, unit, udurations=None):
...
+    task_transitions = list()
+    for metric,spec in task_metrics['consume'].items():
+
+        m = None
+        for m in task_transitions:
+            if m[0] == spec[0]:
+                break
+            m = None
+        if m:
+            assert(m[2] is None)
```
mturilli, line 1293: ugly :)

src/radical/pilot/utils/prof_utils.py
```diff
@@ -1236,5 +1236,160 @@ def _get_unit_consumption(session, unit, udurations=None):
...
+            else              : t[2] = 'system'
+
+    return pilot_transitions, task_transitions
+
+
+# ------------------------------------------------------------------------------
+#
+def get_resource_timelines(session, task, transitions):
+    '''
+    For each the specific task, return a set of tuples of the form:
```
mturilli, line 1363: each specific

src/radical/pilot/utils/prof_utils.py
```diff
@@ -1236,5 +1236,160 @@ def _get_unit_consumption(session, unit, udurations=None):
...
+
+        [start, stop, metric]
+
+    which reports during time what metric has been used.
```
mturilli, line 1367: I am not sure I understand that

src/radical/pilot/utils/prof_utils.py
```diff
@@ -1236,5 +1236,160 @@ def _get_unit_consumption(session, unit, udurations=None):
...
+    if not pid:
+        # task was never assigned to a pilot
+        return dict()
+
+    if 'slots' not in task.cfg:
+        # the task was never scheduled
+        return dict()
+
+    # we heuristically switch between PRTE event traces and normal (fork) event
+    # traces
```
mturilli, line 1384: Not sure I understand that comment

mturilli (commented):
Left some comments but looks good!

src/radical/pilot/agent/scheduler/base.py
andre-merzky: They are very arbitrary indeed, but the behaviour is very benign: the goal is to avoid corner cases: for the sleep, avoid no sleep (busy idle) and avoid sleep which introduces significant latencies (throughput).  Anything smaller than 0.01 is under our noise level and works ok for the latency, and anything larger than 0 is sufficient to avoid busy idle.

For the `unschedule` bulk, the corner case to avoid is waiting for too long to fill a bulk so that latencies add a up and negate the bulk optimization.  For the 0.001 sleep, 128 as bulk size results in a max added latency of about 0.1 second, which is one order of magnitude above our noise level again.

I'll add a comment :-)

src/radical/pilot/utils/prof_utils.py
andre-merzky: > Did we loose `bootstrap_0_start` and `bootstrap_0_stop` in favor of `sub_agent_start` and `sub_agent_stop`?
No, those are still used for the full set of resources owned by the pilot.  The `sub_agent_*` are only for the resources owned by the sub agents.

> What happens if we have multiple subagents?

That works transparently :-)  Multiple event pairs lead to multiple resource accounts.

src/radical/pilot/utils/prof_utils.py
andre-merzky: Apologies, this came from a merge, and I resolved this apparently the wrong way.  Thanks for catching, will fix!

src/radical/pilot/utils/prof_utils.py
andre-merzky: Thanks!

src/radical/pilot/utils/prof_utils.py
andre-merzky: Thanks!

src/radical/pilot/utils/prof_utils.py
andre-merzky: No, that mechanism also works for missing events.  I mean, we then obviously miss the respective resource shifts.  For example, if `cu_exec_stop` is missing we would implicitly delay the change of resource ownership from the executor to the scheduler - but that's to be expected, right?

src/radical/pilot/utils/prof_utils.py
andre-merzky: ack

src/radical/pilot/utils/prof_utils.py
andre-merzky: Yes.  We might consider completely switching from the metric based dictionary format to the transition focused list format at some point - the reverse transition is a bit simpler.  But I added some comments for now.

src/radical/pilot/utils/prof_utils.py
andre-merzky: Thanks

src/radical/pilot/utils/prof_utils.py
andre-merzky: I fixed the comment - better?

src/radical/pilot/utils/prof_utils.py
andre-merzky: Yeah, that didn't make sense - removed (cut'n'paste error)

andre-merzky:
> Left some comments but looks good!

Thanks Matteo, that was helpful!

src/radical/pilot/utils/prof_utils.py
mturilli: Yes, it is expected but it will drug the profile and might be difficult to understand that it was indeed drugged. I guess we will have to catch that in analytics.

mturilli (approved):
Left a reminder for us to keep in mind when working on analytics. Apart adding at least a test if possible, LGTM!

src/radical/pilot/utils/prof_utils.py
andre-merzky: Yes - completeness of events has to be caught elsewhere - the plotting can't do anything about missing information...

andre-merzky:
@mtitov : since you give this branch a try, I'll take the opportunity to also ask you for a review here :-)

PR #2303, opened 2021-01-21
PRTE2 (multi-DVM)

mtitov:


src/radical/pilot/agent/launch_method/prte2.py
```diff
@@ -82,141 +95,195 @@ def rm_config_hook(cls, name, cfg, rm, log, profiler):
...
+        # get `dvm_count` from resource config
+        # FIXME: another option is to introduce `nodes_per_dvm` in config
+        #        dvm_count = math.ceil(len(rm.node_list) / nodes_per_dvm)
+        dvm_count      = cfg['resource_cfg'].get('dvm_count') or 1
+        nodes_per_dvm  = math.ceil(len(rm.node_list) / dvm_count)
+
+        dvm_uri_list   = []
+        dvm_hosts_list = []
+
+        dvm_ready_list = []  # threading.Event that checks DVM's readiness
```
andre-merzky, line 107: You have three lists which need to have the same synced index schema.  Would it make sense to switch to a dict like
```py
dvm_info = {
    'dvm_id': {
        'uri' ; ...,
        'hosts': [...],
        'readly': mt.Event()
    }
        ...
}
```
?

mtitov: so, I thought about this for couple days :)
1) I would think that `dvm_ready` flags should be separate (from other dvm-related info, which will go to `lm_info`), since it is [temporary] internal data, and we don't modify a corresponding list variable (that keeps them), where index of each flag corresponds to `dvm_id`
2) I agree to keep together `dvm_id`, `uri` and `hosts`, but would the following structure fit better?
```python
# record format: (<uri>, <dvm_id>): [<node_id>,...]
        dvm_info = {
            ('prte-batch464440.0;tcp4://10.134.0.44:56089', 0): ['node.0000', 'node.0001']
        }
.....
        lm_info = {'dvm_info'    : dvm_info,
                   'version_info': prte_info,
                   'cvd_id_mode' : 'physical'}
# thus to have key `dvm_info` instead of previous two keys `dvm_uri` and `dvm_hosts`
```

andre-merzky: > **edit:** OR dvm record format `<dvm_id>: (<uri>, [<node_id>,...])`

At the moment, the scheduler sets the `dvm_uri` for the task.  Eventually, it should only set a partition ID, and then we need to look up dvm_uri and nodes via that partition ID.  In that case, the structure above works well IMHO.

mtitov: #2323

andre-merzky (approved):
Thanks!

PR #2311, opened 2021-01-29
Hotfix/stage bs 0

andre-merzky:
This PR fixes remote staging for `bootstrap_0.sh`.

### DIFF
--- a/src/radical/pilot/pmgr/launching/default.py
+++ b/src/radical/pilot/pmgr/launching/default.py
@@ -116,10 +116,6 @@ class Default(PMGRLaunchingComponent):
         except:
             self._log.exception('finalization error')
 
-        self.unregister_timed_cb(self._pilot_watcher_cb)
-        self.unregister_input(rps.PMGR_LAUNCHING_PENDING,
-                              rpc.PMGR_LAUNCHING_QUEUE, self.work)
-
 
     # --------------------------------------------------------------------------
     #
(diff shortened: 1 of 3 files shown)


src/radical/pilot/pmgr/launching/default.py
```diff
@@ -116,6 +116,10 @@ def finalize(self):
         except:
             self._log.exception('finalization error')
 
+        self.unregister_timed_cb(self._pilot_watcher_cb)
+        self.unregister_input(rps.PMGR_LAUNCHING_PENDING,
+                              rpc.PMGR_LAUNCHING_QUEUE, self.work)
+
```
mtitov, line 122: this block is a duplicate, the same block is set earlier inside `try...except`

src/radical/pilot/pmgr/launching/default.py
```diff
@@ -427,6 +432,10 @@ def _start_pilot_bulk(self, resource, schema, pilots):
         rcfg = self._session.get_resource_config(resource, schema)
         sid  = self._session.uid
 
+        # we create a deep-copy of the resource cfg, so that this code is fee to
+        # change it as needed
+        rcfg = ru.Config(cfg=rcfg.as_dict())
```
mtitov, line 437: this one could be skipped, since `self._session.get_resource_config` already returns a deepcopy of config

mtitov (approved):
have only 2 comments (above), with everything else LGTM

andre-merzky:
Sorry, I missed your comments - thanks for reviewing, will address them.

src/radical/pilot/pmgr/launching/default.py
andre-merzky: Thanks, fixed.

src/radical/pilot/pmgr/launching/default.py
andre-merzky: Indeed - fixed.

PR #2312, opened 2021-01-30
Feature/env isolation

andre-merzky:
Finally proper env isolation! :-)  Or at least a first attempt at such for mpirun, fork launch methods and popen executor... The PR also paves the way for heterogeneous tasks as it handles ranks individually and independently during launch.  In relation to that it changes the notion of `slots` to the notion of `ranks` (a slot could in principle host multiple ranks in the past).

Note: this requires radical-cybertools/radical.utils/pull/281

This is a draft PR as of now, as it only can be merged once all LMs and executors are converted.  I would appreciate feedback / reviews either way.

Resolves radical-cybertools/radical.entk#561


src/radical/pilot/agent/agent_0.py
```diff
@@ -374,25 +373,26 @@ def _start_sub_agents(self):
...
                     'cpu_processes'    : 1,
                     'cpu_threads'      : 1,
                     'gpu_processes'    : 0,
                     'gpu_threads'      : 0,
-                  # 'nodes'            : [[node[0], node[1], [[0]], []]],
-                    'nodes'            : [{'name'    : node[0],
-                                           'uid'     : node[1],
+                  # 'ranks'            : [[node[0], node[1], [[0]], []]],
+                    'ranks'            : [{'node'    : node[0],
+                                           'node_id' : node[1],
```
mtitov, line 386: to check: `node_id` or `node_uid`? (if the second one then there are several files where it should be updated)

src/radical/pilot/agent/agent_0.py
```diff
@@ -404,63 +404,40 @@ def _start_sub_agents(self):
...
-            # ------------------------------------------------------------------
+
+                # find a launcher to use
+                launcher = self._rm.find_launcher(agent_task)
+                if not launcher:
+                    raise RuntimeError('no launch method found for sub agent')
+
+                with open(launch_script, 'w') as fout:
+
+                    fout.write('#!/bin/sh\n\n')
```
mtitov, line 415: to have one `write` call (prepare text before to write it down)

p.s. for both launch- and exec-scripts

src/radical/pilot/agent/executing/popen.py
```diff
@@ -122,222 +104,312 @@ def command_cb(self, topic, msg):
...
 
-            # Free the Slots, Flee the Flots, Ree the Frots!
-            self._prof.prof('unschedule_start', uid=task['uid'])
-            self.publish(rpc.AGENT_UNSCHEDULE_PUBSUB, task)
+                # Free the Slots, Flee the Flots, Ree the Frots!
+                self._prof.prof('unschedule_start', uid=task['uid'])
+                self.publish(rpc.AGENT_UNSCHEDULE_PUBSUB, task)
 
-            self.advance(task, rps.FAILED, publish=True, push=False)
+                task['control'] = 'umgr_pending'
```
mtitov, line 132: `tmgr_pending`

src/radical/pilot/agent/executing/popen.py
```diff
@@ -122,222 +104,312 @@ def command_cb(self, topic, msg):
...
+            raise RuntimeError('no launcher foud for task %s' % tid)
+
+        self._log.debug('Launching task with %s', launcher.name)
+
+        launch_script = '%s.launch.sh' % tid
+        exec_script   = '%s.exec.sh'   % tid
+
+        ru.rec_makedir(sbox)
+
+        with open('%s/%s' % (sbox, launch_script), 'w') as fout:
```
mtitov, line 230: flag this part to have one `write` call, same as above

src/radical/pilot/agent/executing/popen.py
```diff
@@ -462,10 +534,226 @@ def _check_running(self):
...
+        ret  = 'while test $(cat %s.sig | wc -l) -lt $RP_RANKS; do\n' % sig
+        ret += '    sleep 1\n'
+        ret += 'done\n'
+
+        return ret
+
+
+    # --------------------------------------------------------------------------
+    #
+    def _get_rank_exec(self, task, rank_id, rank, launcher):
```
mtitov, line 721: consistency in method names: `get_launch_cmd` -> `get_exec_cmd`? (+ `get_rank_setup`?)

src/radical/pilot/agent/launch_method/base.py
```diff
@@ -293,7 +184,7 @@ def get_rank_cmd(self):
 
     # --------------------------------------------------------------------------
     #
-    def construct_command(self, t, launch_script_hop):
+    def get_rank_exec(self, task, rank_id, rank):
```
mtitov, line 187: as a part of having one approach in method names (e.g., getting command or environment setup)

src/radical/pilot/agent/launch_method/mpirun.py
```diff
@@ -15,114 +15,235 @@ class MPIRun(LaunchMethod):
...
+
         else:
-            self.launch_command = ru.which(['mpirun-mpich-mp',    # Mac OSX
-                                            'mpirun-openmpi-mp',  # Mac OSX
-                                            'mpirun',             # general case
-                                           ])
+            # mpirun-mpich-mp  : Mac OSX mpich
+            # mpirun-openmpi-mp: Mac OSX openmpi
+            # mpirun           : general case
+            cmd = 'which mpirun_mpich_mp   || ' \
```
mtitov, line 117: `mpirun-mpich-mp`

src/radical/pilot/agent/launch_method/mpirun.py
```diff
@@ -15,114 +15,235 @@ class MPIRun(LaunchMethod):
...
+        mpi_version = None
+        mpi_flavor  = None
+        env_sh      = '%s/env/%s.env' % (self._pwd, self.name)
+
+        pre_exec  = lmcfg.get('pre_exec', [])
+        cmds      = lmcfg.get('cmds',     [])
+
+        # set up LM environment: first recreate the agent's original env, then
+        # apply the launch method's env_prep commands.  Store the resulting env
+        # in `env/lm_<name>.env`
```
mtitov, line 85: `env/<name>.env`?

src/radical/pilot/agent/launch_method/mpirun.py
```diff
@@ -15,114 +15,235 @@ class MPIRun(LaunchMethod):
...
-        env_list   = self.EXPORT_ENV_VARIABLES + list(task_env.keys())
-        if env_list:
 
-            if self.mpi_flavor == self.MPI_FLAVOR_HYDRA:
-                env_string = '-envlist "%s"' % ','.join(env_list)
+    # --------------------------------------------------------------------------
+    #
+    def get_launcher_env(self):
+
+        return ['. %s' % self._env_sh]
```
mtitov, line 190: naming: with this method name I would expect to have a dict of environment variables, maybe something like `get_launcher_env_cmds`?

p.s. when there is no input parameters we can have it as a property:
```python
    @property
    def env_setup_cmds(self):
         return ['. %s' % self._env_sh]
...
# in Popen:
for cmd in launcher.env_setup_cmds:
    ...
```

src/radical/pilot/agent/executing/popen.py
```diff
@@ -462,10 +534,226 @@ def _check_running(self):
...
+
+        # FIXME: this assumes that the rank has a `gpu_maps` and `core_maps`
+        #        with exactly one entry, corresponding to the rank process to be
+        #        started.
+
+        ret  = ''
+        gmap = rank['gpu_map'][0]
+        if gmap:
+            gpus = ','.join([str(gpu) for gpu in gmap])
+            ret += '        export CUDA_VISIBLE_DEVICES=%s\n' % gpus
```
mtitov, line 731: just a future reminder to check LMs, which have `cvd_id_mode = lm_info.get('cvd_id_mode', 'physical')`

p.s. from obsolete Scheduler's method `_handle_cuda(self, task)`

src/radical/pilot/configs/resource_ncar.json
```diff
@@ -21,21 +21,24 @@
...
         "agent_scheduler"             : "CONTINUOUS",
         "agent_spawner"               : "POPEN",
-        "agent_launch_method"         : "MPIRUN",
-        "task_launch_method"          : "MPIRUN",
-        "mpi_launch_method"           : "MPIRUN",
+        "launch_methods"              : {
+                                         "order" : ["FORK", "MPIRUN"],
+                                         "FORK"  : {},
+                                         "MPIRUN": {
+                                             "env_prep": [
```
mtitov, line 28: propose to have it as `'setup'`, but to have `env_prep` in sync with Task's `env_prep` also works

mtitov: BTW or should it be "in sync" with TaskDescription's attribute `pre_launch`?

src/radical/pilot/task_description.py
```diff
@@ -32,8 +32,12 @@
 
 INPUT_STAGING          = 'input_staging'
 OUTPUT_STAGING         = 'output_staging'
+PRE_LAUNCH             = 'pre_launch'
 PRE_EXEC               = 'pre_exec'
```
mtitov, line 36: - add a comment that `ENVIRONMENT = 'environment'` will be merged with `pre_exec`?
- add a placeholder for `LAUNCH_METHOD`?

mtitov:
got more comments/suggestions to raise (not to force it, but to think about it):
1. LM class methods: `initialize` & `finalize` methods are used in `resource_manager.base` only and other LM methods are used in Executor only
   - Should we have 2 separate classes: `LaunchMethod` and `LauncherService`, which corresponds to initial `rm_..._hook` methods? Thus `LauncherService` will be used in RM and `LaunchMethod` in Executor
   - Should LM configuration be taken out from initialize? To configure LM on a side of Executor (especially if sub-agents are used)
2. Propose to have a list of "active" LMs (provided either by user and/or default list) that will be initialized/used, so not to activate all LMs available in config
   - Move attribute `order` to of `launch_methods`, since it kinda breaks this internal structure (similar as with `schemas`)
   - Maybe to have additional attributes per LM config as `default:bool`, `agent_node:bool`
      - Speaking about agent nodes -> to have a special method `get_agent_launcher` in RM rather than `find_launcher`
3. Re-think how to deal with partitions per LM in RM: `self.lm_info` is dict with keys per LM ([link](https://github.com/radical-cybertools/radical.pilot/blob/feature/env_isolation/src/radical/pilot/agent/resource_manager/base.py#L154)), and we move partitions definition to RM ([link](https://github.com/radical-cybertools/radical.pilot/blob/feature/env_isolation/src/radical/pilot/agent/resource_manager/base.py#L166))
   - back to p.1 -> to have a list of active LM will help to control which LM will bring partitioning
   - Maybe to move partitions from RM to Executor? (in case we will have several LMs which deals with partitions, so their internal partitions organization will not overlap)

src/radical/pilot/agent/launch_method/base.py
```diff
@@ -53,57 +53,34 @@
...
+        self._cfg  = cfg
+        self._log  = log
+        self._prof = prof
+        self._pwd  = os.getcwd()
 
-        # A per-launch_method list of env vars to remove from the Task env
-        self.env_removables = []
-
-        self._configure()
+        self._info = self._cfg.get('lm_info', {}).get(self.name)
```
mtitov, line 72: - inside RM, this attribute will be empty here, but will be set during `initialize` method (from provided config file), and structures inside `self._info` for these cases will be different
- inside Executor, this attribute will be set here, but will not be used anyhow, since configuration was done within `initialize` method

these two points are in favor of what I wrote earlier: to split LM class in a way that one [sub-class] will be used in RM and another in Executor; to have configuration procedure within Executor.

andre-merzky:
`partitions` should live in `rm_info`, not in `lm_info` - the RM needs to define it.  Only cover in this PR if `lm_info` restucturing requires it.

andre-merzky:
`self._launchers` is replicated in RM and LM base - only keep in RM?

andre-merzky:
Either return to `task_lm`, `mpi_lm` etc and describe them explicitly, or allow user to define what LMs should be used in a pilot.

src/radical/pilot/agent/agent_0.py
andre-merzky: `node_id` (for no particular reason).  I made the name more consistent in the code now.

src/radical/pilot/agent/executing/popen.py
andre-merzky: thanks, fixed.

src/radical/pilot/agent/launch_method/mpirun.py
andre-merzky: thanks, fixed.

src/radical/pilot/agent/launch_method/mpirun.py
andre-merzky: I want to avoid collisions with other env names, thus the prefix.

lee212:
FYI, my test on Summit has an issue with this error message:

```
$ more agent.0.out
bootstrap agent agent.0
failed agent.0
--------------
RADICAL Utils -- Stacktrace [32101] [MainThread]

hrlee     32101  27797 22 11:33 ?        00:00:00  |               \_ rp.agent.0
File "/gpfs/alpine/bip212/scratch/hrlee/radical.pilot.sandbox/re.session.login4.hrlee.018764.0003/pilot.0000/rp_install/bin/radical-pilot-agent", line 51, in bootstrap_3
if aid == 'agent.0': agent = rp.Agent_0(cfg, session)
File "/gpfs/alpine/bip212/scratch/hrlee/radical.pilot.sandbox/re.session.login4.hrlee.018764.0003/pilot.0000/rp_install/lib/python3.7/site-packages/radical/pilot/agent/agent_0.py", line 76, in __init__
self._configure_rm()
File "/gpfs/alpine/bip212/scratch/hrlee/radical.pilot.sandbox/re.session.login4.hrlee.018764.0003/pilot.0000/rp_install/lib/python3.7/site-packages/radical/pilot/agent/agent_0.py", line 162, in _configure_rm
cfg=self._cfg, session=self._session)
File "/gpfs/alpine/bip212/scratch/hrlee/radical.pilot.sandbox/re.session.login4.hrlee.018764.0003/pilot.0000/rp_install/lib/python3.7/site-packages/radical/pilot/agent/resource_manager/base.py", line 265, in create
return impl(cfg, session)
File "/gpfs/alpine/bip212/scratch/hrlee/radical.pilot.sandbox/re.session.login4.hrlee.018764.0003/pilot.0000/rp_install/lib/python3.7/site-packages/radical/pilot/agent/resource_manager/lsf.py", line 91, in __init__
self._configure()
File "/gpfs/alpine/bip212/scratch/hrlee/radical.pilot.sandbox/re.session.login4.hrlee.018764.0003/pilot.0000/rp_install/lib/python3.7/site-packages/radical/pilot/agent/resource_manager/lsf.py", line 192, in _configure
raise RuntimeError("$LSB_DJOB_HOSTFILE not set!")
RuntimeError: $LSB_DJOB_HOSTFILE not set!

--------------
finalize agent.0

```
and the stack info:
```

  python               : /ccs/home/hrlee/.conda/envs/env_isolation/bin/python3
  pythonpath           : /sw/summit/xalt/1.2.1/site:/sw/summit/xalt/1.2.1/libexec
  version              : 3.7.10
  virtualenv           : env_isolation

  radical.entk         : 1.6.5
  radical.gtod         : 1.5.0
  radical.pilot        : 1.6.5-v1.6.5-249-g3d4f403@feature-env_isolation
  radical.saga         : 1.6.5
  radical.utils        : 1.6.5-v1.6.5-19-ge6c3783@feature-env_isolation

```

andre-merzky:
@lee212 : would you mind attaching the `env/` subtree of the pilot sandbox, please?  Thank you!

lee212:
[env.tar.gz](https://github.com/radical-cybertools/radical.pilot/files/6529104/env.tar.gz)
@andre-merzky , sorry for the late response, please check the attached here!

andre-merzky:
> [env.tar.gz](https://github.com/radical-cybertools/radical.pilot/files/6529104/env.tar.gz)
> @andre-merzky , sorry for the late response, please check the attached here!

Thanks @lee212!  The problem should be fixed now, the canonical examples seem to work on summit now.  Please give it another try if you can.  Note that you will also need to update RU (same branch name).

src/radical/pilot/agent/launch_method/mpirun.py
```diff
@@ -15,114 +15,176 @@ class MPIRun(LaunchMethod):
...
+    # --------------------------------------------------------------------------
+    #
+    def get_launcher_env(self):
+
+        return ['. $RP_PILOT_SANDBOX/%s' % self._env_sh]
+
+
+    # --------------------------------------------------------------------------
+    #
+    def get_task_env(self, spec):
```
mtitov, line 136: this should be moved to LM base class? (currently only 3 LMs have it, and it is duplicated for all of them)

src/radical/pilot/agent/launch_method/mpirun.py
```diff
@@ -15,114 +15,176 @@ class MPIRun(LaunchMethod):
...
 
-        # Construct the executable and arguments
-        if task_argstr: task_command = "%s %s" % (task_exec, task_argstr)
-        else          : task_command = task_exec
+    # --------------------------------------------------------------------------
+    #
+    def can_launch(self, task):
+
+        # mpirun can launch any executable
+        if task['description']['executable']:
```
mtitov, line 122: There is a different check in LM MPIExec, should for both of them be used a "combined" approach?
```python
        td = task['description']
        if td['executable'] and td['cpu_process_type'] == rpc.MPI:
            return True
```

src/radical/pilot/agent/scheduler/base.py
```diff
@@ -903,82 +907,12 @@ def _try_allocation(self, task):
         self._change_slot_states(slots, rpc.BUSY)
         task['slots'] = slots
 
-        if slots['lfs_per_node']['path']:
-            task['description']['environment']['NODE_LFS_PATH'] = \
-                slots['lfs_per_node']['path']
```
mtitov, line 908: Should this part be kept? or it was moved?

lee212:
@andre-merzky , I confirm that the issue on Summit has been resolved, I was able to run an example with one change, CPU core count per node.
I used to define 168 to use a single node in the entk's `resource_des` but now it seems I need to set physical core count 42 instead otherwise I see:
```
RADICAL Utils -- Stacktrace [109962] [MainThread]

hrlee    109962 109212 12 09:32 ?        00:00:00                  \_ rp.agent.0
hrlee    110206 109962  0 09:32 ?        00:00:00                  |   \_ [rp.agent.0] <defunct>
File "/gpfs/alpine/scratch/hrlee/bip212/radical.pilot.sandbox/re.session.login1.hrlee.018785.0003/pilot.0000/rp_install/bin/radical-pilot-agent", line 51, in
 bootstrap_3
if aid == 'agent.0': agent = rp.Agent_0(cfg, session)
File "/gpfs/alpine/scratch/hrlee/bip212/radical.pilot.sandbox/re.session.login1.hrlee.018785.0003/pilot.0000/rp_install/lib/python3.7/site-packages/radical/p
ilot/agent/agent_0.py", line 76, in __init__
self._configure_rm()
File "/gpfs/alpine/scratch/hrlee/bip212/radical.pilot.sandbox/re.session.login1.hrlee.018785.0003/pilot.0000/rp_install/lib/python3.7/site-packages/radical/p
ilot/agent/agent_0.py", line 165, in _configure_rm
cfg=self._cfg, session=self._session)
File "/gpfs/alpine/scratch/hrlee/bip212/radical.pilot.sandbox/re.session.login1.hrlee.018785.0003/pilot.0000/rp_install/lib/python3.7/site-packages/radical/p
ilot/agent/resource_manager/base.py", line 271, in create
return impl(cfg, session)
File "/gpfs/alpine/scratch/hrlee/bip212/radical.pilot.sandbox/re.session.login1.hrlee.018785.0003/pilot.0000/rp_install/lib/python3.7/site-packages/radical/p
ilot/agent/resource_manager/lsf.py", line 44, in __init__
ResourceManager.__init__(self, cfg, session)
File "/gpfs/alpine/scratch/hrlee/bip212/radical.pilot.sandbox/re.session.login1.hrlee.018785.0003/pilot.0000/rp_install/lib/python3.7/site-packages/radical/p
ilot/agent/resource_manager/base.py", line 75, in __init__
self._init_from_scratch()
File "/gpfs/alpine/scratch/hrlee/bip212/radical.pilot.sandbox/re.session.login1.hrlee.018785.0003/pilot.0000/rp_install/lib/python3.7/site-packages/radical/p
ilot/agent/resource_manager/base.py", line 197, in _init_from_scratch
% (str(cores_avail), str(self.requested_cores)))
ValueError: Not enough cores available (42 < 168).
```
my updated stack:
```

  python               : /ccs/home/hrlee/.conda/envs/env_isolation/bin/python3
  pythonpath           : /sw/summit/xalt/1.2.1/site:/sw/summit/xalt/1.2.1/libexec
  version              : 3.7.10
  virtualenv           : env_isolation

  radical.entk         : 1.6.5
  radical.gtod         : 1.5.0
  radical.pilot        : 1.6.6-v1.6.6-75-g14ca46c@feature-env_isolation
  radical.saga         : 1.6.5
  radical.utils        : 1.6.6-v1.6.6-23-gbc34a24@feature-env_isolation
```

src/radical/pilot/agent/executing/popen.py
```diff
@@ -143,13 +126,18 @@ def work(self, tasks):
                 self._prof.prof('unschedule_start', uid=task['uid'])
                 self.publish(rpc.AGENT_UNSCHEDULE_PUBSUB, task)
 
+                task['control'] = 'tmgr_pending'
+                task['$all']    = True
```
mtitov, line 130: The same is set at the staging output phase, should it be there only? (we give a control back to TMGR after staging output is passed)

src/radical/pilot/agent/executing/popen.py
```diff
@@ -474,5 +559,236 @@ def _check_running(self):
...
+        ret += 'export RP_RESOURCE="%s"\n'         % self.resource
+        ret += 'export RP_RESOURCE_SANDBOX="%s"\n' % self.rsbox
+        ret += 'export RP_SESSION_SANDBOX="%s"\n'  % self.ssbox
+        ret += 'export RP_PILOT_SANDBOX="%s"\n'    % self.psbox
+        ret += 'export RP_TASK_SANDBOX="%s"\n'     % sbox
+        ret += 'export RP_LFS="%s"\n'              % self.lfs
+        ret += 'export RP_GTOD="$RP_PILOT_SANDBOX/gtod"\n'
+
+        if self._prof.enabled:
+            ret += 'export RP_PROF="$RP_PILOT_SANDBOX/prof"\n'
```
mtitov, line 653: `RP_GTOD` and `RP_PROF` are set as attributes in `AgentExecutingComponent` base class, should they be deleted from there? (I would prefer to delete them from there, otherwise they should be referenced here, e.g., `ret += 'export RP_GTOD="%s"\n' % self.gtod`)

src/radical/pilot/agent/executing/popen.py
```diff
@@ -164,174 +152,271 @@ def _handle_task(self, task):
...
+            fout.write(self._get_prof('launch_stop', tid))
+            fout.write('exit $RP_RET\n')
+
+            fout.write(self._separator)
+            fout.write('\n')
+
+        ranks   = task['slots']['ranks']
+        n_ranks = len(ranks)
+
+        with open('%s/%s' % (sbox, exec_script), 'w') as fout:
```
mtitov, line 291: sorry for duplication, just want to keep it up - it's important to have only one call for `write` method per file

src/radical/pilot/agent/executing/popen.py
```diff
@@ -474,5 +559,236 @@ def _check_running(self):
...
+
+
+    # --------------------------------------------------------------------------
+    #
+    def _get_launch_cmds(self, task, launcher, exec_path):
+
+
+        ret  = ''
+        cmds = ru.as_list(launcher.get_launch_cmds(task, exec_path))
+        for cmd in cmds:
```
mtitov, line 609: all launch methods return a string (one command), should we have just `cmd = launcher.get_launch_cmd(task)`? (and not to have `exec_path` as an attribute there, since we build a command without that argument)

src/radical/pilot/agent/launch_method/mpirun.py
andre-merzky: Hmm - they are identical, true - but also they need to remove the launcher env, and in principle that is launcher specific.  Well, I can move it, let's see how it goes...

src/radical/pilot/agent/launch_method/mpirun.py
andre-merzky: Thanks - I made the check the same, but for the time being would like to keep the two LMs separate.

src/radical/pilot/agent/scheduler/base.py
andre-merzky: Ack, readded (changed the name for consistency, always set it)

src/radical/pilot/agent/agent_0.py
andre-merzky: Thanks, done.

src/radical/pilot/agent/executing/popen.py
andre-merzky: Same, done.

src/radical/pilot/agent/executing/popen.py
andre-merzky: would like to keep for now.

src/radical/pilot/agent/launch_method/base.py
andre-merzky: Does that comment belong to the previous one?  If not I am missing the point I'm afraid... :-/

src/radical/pilot/agent/executing/popen.py
andre-merzky: right - I added a FIXME for now.

src/radical/pilot/task_description.py
andre-merzky: `environment` is not really merged with `pre_exec`, is it?  They are both applied though.  Would the user benefit from learning how we apply it, or get confused (like, does it really matter where to specify what when it's merged anyway)?

For the `LAUNCH_METHOD` - I would like to keep that out for the moment.  Mostly because I really want to make sure that we don't *need* to specify a LM for now (RP should be able to infer), and to make sure that j/psi can map to RP even if RP does not have an LM attribute.

andre-merzky:
> got more comments/suggestions to raise (not to force it, but to think about it):
> 
>     1. LM class methods: `initialize` & `finalize` methods are used in `resource_manager.base` only and other LM methods are used in Executor only
>        
>        * Should we have 2 separate classes: `LaunchMethod` and `LauncherService`, which corresponds to initial `rm_..._hook` methods? Thus `LauncherService` will be used in RM and `LaunchMethod` in Executor
>        * Should LM configuration be taken out from initialize? To configure LM on a side of Executor (especially if sub-agents are used)
> 
>     2. Propose to have a list of "active" LMs (provided either by user and/or default list) that will be initialized/used, so not to activate all LMs available in config
>        
>        * Move attribute `order` out of `launch_methods`, since it kinda breaks this internal structure (similar as with `schemas`)
>        * Maybe to have additional attributes per LM config as `default:bool`, `agent_node:bool`
>          
>          * Speaking about agent nodes -> to have a special method `get_agent_launcher` in RM rather than `find_launcher`
> 
>     3. Re-think how to deal with partitions per LM in RM: `self.lm_info` is dict with keys per LM ([link](https://github.com/radical-cybertools/radical.pilot/blob/feature/env_isolation/src/radical/pilot/agent/resource_manager/base.py#L154)), and we move partitions definition to RM ([link](https://github.com/radical-cybertools/radical.pilot/blob/feature/env_isolation/src/radical/pilot/agent/resource_manager/base.py#L166))
>        
>        * back to p.1 -> to have a list of active LM will help to control which LM will bring partitioning
>        * Maybe to move partitions from RM to Executor? (in case we will have several LMs which deal with partitions, so their internal partitions organization will not overlap)

All good points, but maybe not quite in scope for the PR?  Let's see how things look after the initialize / finalize are fixed...

src/radical/pilot/agent/launch_method/mpirun.py
mtitov: sorry, I didn't mean to merge them, but to have a check with both `executable` and `cpu_process_type` (probably no need to check for `executable`, since it always should be set)

src/radical/pilot/configs/resource_ornl.json
```diff
@@ -114,35 +126,43 @@
...
+                                             "pre_exec_cached": [
+                                                 "module unload xl",
+                                                 "module unload xalt",
+                                                 "module unload spectrum-mpi",
+                                                 "module load   gcc/6.4.0",
+                                                 "module use    /sw/summit/ums/ompix/gcc/6.4.0/modules",
+                                                 "module load   prrte/master"
+                                             ],
+                                             "pre_exec": [
+                                                 "ulimit -u 65536"
```
mtitov, line 147: should launch methods have `pre/post_launch` parameter? since `pre_exec` is what should be set within `exec_script` (which is executed with launch command)

As follow up to this I got confused (sorry if this was discussed already) with new `TaskDescription` attributes: `pre_launch` and `post_launch`. These attributes are related to LM and should be described in resource config (and `TaskDescription` is only responsible for task related attributes `pre/post_exec` and `pre/post_rank`)

mtitov:
> All good points, but maybe not quite in scope for the PR? Let's see how things look after the initialize / finalize are fixed...

I think at least p.2 should be part of this PR

src/radical/pilot/agent/agent_0.py
```diff
@@ -4,13 +4,10 @@
 
 
 import os
-import sys
 import copy
 import stat
 import time
 import pprint
-import subprocess          as sp
-import multiprocessing     as mp
```
mtitov, line 13: deleted libs are in use below (L414, 427-428, 431)

mtitov:
method `Agent_0._start_services` is not in sync with the new approach  ([link](https://github.com/radical-cybertools/radical.pilot/blob/47ba0d13dee622f442275a9f928ca752638ecdfd/src/radical/pilot/agent/agent_0.py#L342))

src/radical/pilot/agent/agent_0.py
mtitov: this is related to the comment below about `_start_services`, thus could be skipped if that method will be updated accordingly

src/radical/pilot/agent/launch_method/base.py
```diff
@@ -53,56 +50,61 @@
...
-
-        # A per-launch_method list of env vars to remove from the Task env
-        self.env_removables = []
-
-        self._configure()
+    def __init__(self, name, lm_cfg, cfg, log, prof):
+
+        log.debug('===== lm base init start')
+        self.name    = name
+        self._lm_cfg = cfg
```
mtitov, line 64: seems that this attribute `self._lm_cfg` is never used (and should input parameter `lm_cfg` be assigned?)

src/radical/pilot/agent/executing/popen.py
andre-merzky: This task is failed and will not go through output staging, that's why we have to set it here.

src/radical/pilot/agent/executing/popen.py
andre-merzky: ack, done.

src/radical/pilot/agent/executing/popen.py
andre-merzky: I wanted to give the LMs the chance to add commands, like set an env, lookup some endpoint, or whatever.

I don't understand the comment about `exec_path` - it is passed to `launcher.get_launch_cmds(task, exec_path)` as you say, but it is actually used there, and the LM cannot infer the location of the exec script otherwise.  What am I missing?

src/radical/pilot/agent/executing/popen.py
andre-merzky: Thanks, I cleaned that up (fixed here)

src/radical/pilot/agent/launch_method/base.py
andre-merzky: Thanks, I fixed the assignment.  Not sure if we need to keep it around, it's indeed not used, but we usually keep configs in other classes also, so... *shrug*

src/radical/pilot/agent/launch_method/base.py
andre-merzky: *ping*  I am not sure how to interpret this comment ...

src/radical/pilot/agent/launch_method/mpirun.py
andre-merzky: Well, the nice thing is that they also can launch non-MPI executables, so the test for `cpu_process_type` is actually not needed.  `executable` may not be set in the future for function based tasks.  So yeah, this is a bit premature, but ok I hope...

src/radical/pilot/agent/scheduler/base.py
mtitov: is the name `NODE_LFS_PATH` actually used by the machine and not by RP? If it is so then it shouldn't be renamed

src/radical/pilot/agent/bootstrap_0.sh
```diff
@@ -232,6 +223,25 @@ create_gtod()
...
+#
+create_prof(){
+
+    cat > ./prof <<EOT
+#!/bin/sh
+
+test -z "\$RP_PROF_TGT" && exit
+
+now=\$(\$RP_PILOT_GTOD)
+printf "%.7f,\$1,\$RP_SPAWNER_ID,MainThread,\$RP_UNIT_ID,AGENT_EXECUTING,\\\n" \$now\\
```
mtitov, line 236: `RP_PILOT_GTOD` -> `RP_GTOD`
`RP_SPAWNER_ID` -> ?
`RP_UNIT_ID` -> `RP_TASK_ID`

src/radical/pilot/agent/scheduler/base.py
mtitov: btw, `Popen` also sets `RP_LFS` ([link](https://github.com/radical-cybertools/radical.pilot/blob/f333d086aeeef51cf4f34c707ceb0aabc4514acb/src/radical/pilot/agent/executing/popen.py#L661)) is it related to this as well? (also don't see where it is used by RP)

mturilli:
resolves radical-cybertools/entk#561

src/radical/pilot/agent/launch_method/base.py
mtitov: that was a proposal of having a certain approach in method names (`.._cmd`, `.._env_setup`, `get_.._env`, etc.). But could be skipped. _Resolve this conversation_

src/radical/pilot/task_description.py
mtitov: > environment is not really merged with pre_exec, is it? They are both applied though. Would the user benefit from learning how we apply it, or get confused (like, does it really matter where to specify what when it's merged anyway)?

Having both attributes (which do the same?) means having them in sync in a way how they are treated, and thus seems like duplication. And with new attributes, which explicitly point where they do env setup, attribute `environment` may seem confusing - where it is applied - is it at every time when new environment is set (e.g., pre_launch, pre_exec, pre_rank)? is it only when task exec called? etc. Probably not the case for this PR, but would be good to review it later.

> For the LAUNCH_METHOD - I would like to keep that out for the moment

agree not to keep it within this PR, but for further development should be considered that not all LMs, which are in config, should be activated (since that could be a waste of resources, e.g., activate PRTE LM and not use it). Thus should be some list of "active" LMs with "primary" and "auxiliary" roles, and possibility to reset it without changing resource config file.

src/radical/pilot/agent/executing/popen.py
mtitov: Understand about having multiple commands, but then there is some inconsistency in what you expect as output and what it gives -> currently these methods give `str` as output, but `list` is expected.

With `exec_path` - it is just added at the end of the produced string (not used to make any decisions inside the method and is not modified either), and my thought was (considering method provides a single command) that method generates a command, which is set within the script and represented as a cmd-line while adding a corresponding executable path -> request for a command (`tail -f`) which is used to have a [complete] cmd-line (`cat <file_name>`). Sorry, could sound over-complicated, but also could be skipped if we talk about list of commands (cmd-lines)

src/radical/pilot/configs/resource_ornl.json
mtitov: Second paragraph could be skipped (we discussed it - should be a possibility to set additional parameters if needed at every level)

First paragraph - we agreed to have `setup` or similar key instead of `pre_exec[_..]`, right?

mtitov (approved):
LGTM

andre-merzky:
Hey @mtitov  - I worked on the pylint issues, but those will need https://github.com/radical-cybertools/radical.utils/pull/322 to be released before showing improvement.

andre-merzky:
@mtitov : fixed all  but one test / lint error: the integration test with JSRUN would need rewriting.  I am actually not sure how to do that now since the LM does not spit out a simple command anymore.  Any idea how to handle this?  Don't really want to reimplement the popen executor for a test... :-/

PR #2314, opened 2021-02-01
Extend partitioning from DVMs to a general approach

mtitov:
This PR as a proposal to update dvm-partitioning as a general approach
- a step forward partitioning
- will help to fix tagging/partitioning in PRTE (when I started to use `tag` as `dvm_id` I skipped a possibility to use `tag` for co-location...)

p.s. this is an intermediate PR for `fix/prte2` branch


mtitov:
Comments of the corresponding sections are not updated [completely], so if this PR is accepted, then I'll update comments accordingly before PR approval

PR #2315, opened 2021-02-01
fix help string in bin/radical-pilot-create-static-ve

andre-merzky:


bin/radical-pilot-create-static-ve
```diff
@@ -58,12 +58,16 @@ help(){
...
 
-    usage: $0 <target> [-h]
+    usage: $0 -p <path> [-v <version>] [-t <type>] [-m <modules>]
 
-    This script creates a virtualenv at the given target location.  That
+    This script creates a virtualenv at the given target path.  That
     virtualenv should be suitable to be used as static VE for a radical.pilot
     target resource, and can be specified in a resource configuration for RP.
 
+    type   :  bw, bwpy, *default*
```
iparask, line 67: What is the difference between these types?

iparask (approved):
Looks good to me.

bin/radical-pilot-create-static-ve
andre-merzky: Thanks - I simplified that (`bw` was only used internally), and added a short docstring.

PR #2316, opened 2021-02-02
Keep default cfg without sub-agent

mtitov:
In case having sub-agent (for executing component) in default agent config was not intentionally, otherwise this PR should be cancelled.

p.s. sub-agent will reserve 1 node, which could be implicit for the user


andre-merzky (approved):
Thanks for catching that **again**!

PR #2325, opened 2021-02-10
Create pypi publish action

iparask:
This PR introduces a publish action for Github. As soon as a github release is made, the action gets triggered. It checks out master, builds it, and pushes it to pypi.

One thing we need to decide is whose credentials will become the pypi username and password secrets.

### DIFF
--- a/.github/workflows/python-publish.yml
+++ b/.github/workflows/python-publish.yml
@@ -1,0 +1,33 @@
+# This workflow will upload a Python Package using Twine when a release is created
+# For more information see: https://help.github.com/en/actions/language-and-framework-guides/using-python-with-github-actions#publishing-to-package-registries
+
+name: Upload Python Package
+
+on:
+  release:
+    types: [created]
+
+jobs:
+  deploy:
+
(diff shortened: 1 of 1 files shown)


andre-merzky (approved):
The release should also trigger a tag:
```sh
git tag "v$(cat VERSION)"
git push --tags
```

> actions/setup-python@v2
This does not mean Python-2, right?

I appreciate the idea - but don't think I would use it myself, really.  I like github and the features it provides around pull requests and issues - but then again, I am not a fan of button pushing on web pages when a command line version of the same takes me literally 3 seconds...  Don't want to be too negative, so apologies, just liking my old habits :-)  I am not opposed to have this option!

iparask:
```
    - name: Set up Python
       uses: actions/setup-python@v2
       with:
         python-version: '3.6'
```
`v2` is the version of the action. Python is 3.6

iparask:
I'll add the tag and get back at you

iparask:
You do not need to push the tag, because the release process creates it as well. I'm merging it

PR #2326, opened 2021-02-10
add missing entry

andre-merzky:
As discussed...


iparask:
Hello @andre-merzky, the linters fail because we missed two entries in the CU to task refactoring. Should it change?

mtitov (approved):
LGTM

PR #2327, opened 2021-02-10
Updated RM LSF_SUMMIT the same way as RM base class (partitions and prte related)

mtitov:


src/radical/pilot/agent/resource_manager/lsf_summit.py
```diff
@@ -146,6 +147,14 @@ def __init__(self, cfg, session):
...
             self._log.info("ResourceManager config hook succeeded (%s)" % lm)
 
+        # check for partitions after `rm_config_hook` was applied
+        # NOTE: partition ids should be of a string type, because of JSON
+        #       serialization (agent config), which keeps dict keys as strings
+        if self.lm_info.get('dvm_hosts'):
+            # case of DVM partitioning (PRTE/2 LM)
+            self.partitions = {
+                str(k): v for k, v in enumerate(self.lm_info['dvm_hosts'])}
+
```
andre-merzky, line 157: That PR looks good to me.  I have a nitpick though: the notion of a `dvm` should not appear outside of the LM I think.  Would you mind renaming `dvm_hosts` to `partition_nodes`?  I know, that comment would have better fit to the last PR on that topic, but I missed it there... :-/  Thanks!

andre-merzky (approved):
LGTM!

src/radical/pilot/agent/resource_manager/lsf_summit.py
andre-merzky: Yes, I think that works very well - thanks!

mtitov: introduced new structure for dvm-partitioning, after couple more tests will be ready to be merged

andre-merzky: This looks great! :-)
So, LSF_Summit RM now has a `self._partitions` attribute, but is not using it - how come?  Should that go to the base class maybe, so that all RMs have that attribute in case they need it, or is it only needed by scheduler and LM?

andre-merzky (approved):
One more comment - but looks good, thank you!

src/radical/pilot/agent/resource_manager/lsf_summit.py
mtitov: currently LSF_Summit doesn't do init call from base class, that's why it has its own setup (base class has `self._partitions` as well), and that attribute then stored in `self._rm_info` and goes back to Agent, which stores all that in configs for components

andre-merzky: Ah, right, you told me already, at least twice!  Sorry for the noise...

Issue #1469, opened 2017-10-02
How many cores/nodes does the RP agent require


vivek-bala:
I wanted to know the following:

* On which resources does RP agent need 1 core?
* On which resources does RP agent need 1 node?
* Which parameter(s) does this depend on?

It might be useful to document this and put it in the RTD as well

andre-merzky:
This depends on the agent configuration: any sub-agent can be configured to run on a compute node, and in that case that compute node will not be available for the workload execution.  Its currently always a full node (not a core).

Now, the Cray's kind of require us to use sub-agents, so the default configuration on the cray has one sub-agent, so we use one node there.  Otherwise the default agent config has no sub-agents, so no compute node is reserved.

This is a somewhat simplified picture, apologies.  I agree that needs better documentation.  From the discussions over the last weeks, I would by now favor the mode where we internally add the required number of cores/nodes to the job size, so that the requested pilot size is always fully available to the workload.  Happy to use this ticket to discuss this, let me know if you have any thoughts...

vivek-bala:
> add the required number of cores/nodes to the job size, so that the requested pilot size is always fully available to the workload

I don't think we do this. Do we? I think when we require a node for the agent, we expect the user to add a node for the agent. If we require a core, then we take one core from the resources the user requested. @andre-merzky Can you confirm that?

andre-merzky:
We never reserve a core - we either reserve a node, or co-locate the agent and the tasks on the first node, or run on the MOM node (not affecting the compute nodes at all).  But no, the documentation has not been updated I think.

vivek-bala:
When we co-locate the agent and tasks on the first node, do they use the same core? Wouldn't that mean that some tasks would be slower (even if the workload is homogeneous) due the interference from RP components?

andre-merzky:
> When we co-locate the agent and tasks on the first node, do they use the same core? Wouldn't that mean that some tasks would be slower (even if the workload is homogeneous) due the interference from RP components?

Yes, indeed.  We rarely use that setup though, almost all newer and larger machines have MOM or Batch nodes which we use to host the agent.  If the agent's performance impact is impacting the workload (say, for a highly uniform BoT), then I would recommend to pick an agent config which reserves a node for the agent.

mturilli:
[...]

PR #2331, opened 2021-02-18
Feature/nodes

andre-merzky:
This PR adds 'nodes' to the pilot description


src/radical/pilot/pmgr/launching/default.py
```diff
@@ -841,7 +694,7 @@ def _prepare_pilot(self, resource, rcfg, pilot, expand, tar_name):
         rp_version              = rcfg.get('rp_version')
         virtenv_mode            = rcfg.get('virtenv_mode',        DEFAULT_VIRTENV_MODE)
         virtenv                 = rcfg.get('virtenv',             default_virtenv)
-        cores_per_node          = rcfg.get('cores_per_node', 0)
+        cores_per_node          = rcfg.get('cores_per_node', 1)
```
mtitov, line 697: for some systems this parameter is not set in a resource config, and for such cases default value `1` will mess up with `number_cores` and in this case it is tricky to use for `number_nodes` below.

src/radical/pilot/pmgr/launching/default.py
```diff
@@ -1037,19 +890,29 @@ def _prepare_pilot(self, resource, rcfg, pilot, expand, tar_name):
...
                 raise ValueError('no local env found')
 
-        # if cores_per_node is set (!= None), then we need to
-        # allocation full nodes, and thus round up
-        if cores_per_node:
-            cores_per_node = int(cores_per_node)
-            number_cores   = int(cores_per_node *
-                             math.ceil(float(number_cores) / cores_per_node))
+        if number_nodes:
+            number_cores = number_nodes * cores_per_node
```
mtitov, line 894: continue from comment above about `cores_per_node` - if default value is used, then requested `number_cores` will be incorrect (for example for Summit, `cores_per_node` will depend on `SMT` level)

src/radical/pilot/pmgr/launching/default.py
```diff
@@ -15,6 +15,8 @@
 import radical.saga            as rs
 import radical.utils           as ru
 
+rsfs = rs.filesystem
```
mtitov, line 18: this could be removed, since it was used in removed methods `_handle_pilot_input_staging` and `_handle_pilot_output_staging`

src/radical/pilot/pmgr/launching/default.py
```diff
@@ -1037,19 +890,29 @@ def _prepare_pilot(self, resource, rcfg, pilot, expand, tar_name):
...
-            cores_per_node = int(cores_per_node)
-            number_cores   = int(cores_per_node *
-                             math.ceil(float(number_cores) / cores_per_node))
+        if number_nodes:
+            number_cores = number_nodes * cores_per_node
+            number_gpus  = number_nodes * gpus_per_node
+            self._log.debug('nodes: %s [%s %s]',
+                            number_nodes, cores_per_node, gpus_per_node)
+            self._log.debug('nodes: %s [%s %s]',
+                            number_nodes, number_cores,   number_gpus)
```
mtitov, line 899: may be just to have one line?
```python
            self._log.debug('nodes: %s [%s %s], cores: %s, gpus: %s',
		            number_nodes, cores_per_node, gpus_per_node
                            number_cores, number_gpus)
```

src/radical/pilot/pmgr/launching/default.py
```diff
@@ -1264,11 +1127,6 @@ def _prepare_pilot(self, resource, rcfg, pilot, expand, tar_name):
                     'site:%s/%s > %s' % (session_sandbox, sdist, sdist)
                 ])
 
-            if stage_cacerts:
-                jd.file_transfer.extend([
-                    'site:%s/%s > %s' % (session_sandbox, certs, certs)
-                ])
```
mtitov, line 1270: two interdependent questions: why this was deleted? and is section under `if not shared_filesystem` actually needed?

mturilli:
Fixes #1469

src/radical/pilot/pmgr/launching/default.py
```diff
@@ -888,42 +890,51 @@ def _prepare_pilot(self, resource, rcfg, pilot, expand, tar_name):
...
-            elif os.environ.get('CONDA_PREFIX'):
-                python_dist = 'anaconda'
-                virtenv     = os.environ['CONDA_PREFIX']
-            else:
-                # we can't use local
-                self._log.error('virtenv_mode is local, no local env found')
-                raise ValueError('no local env found')
+        # estimate requested resources
+
+        if os.environ.get('RADICAL_SAGA_SMT'):
```
andre-merzky, line 895: This could be simpler written as:
```py
system_architecture['smt'] = int(os.environ.get('RADICAL_SAGA_SMT', 1))
assert(system_architecture['smt'] > 0)
```
Exceptions can fall through IMHO

src/radical/pilot/pmgr/launching/default.py
```diff
@@ -888,42 +890,51 @@ def _prepare_pilot(self, resource, rcfg, pilot, expand, tar_name):
...
+            try:
+                system_architecture['smt'] = int(os.environ['RADICAL_SAGA_SMT'])
+            except Exception as e:
+                self._log.debug('SAGA SMT not set: %s' % e)
```
andre-merzky, line 899: maybe use `self._log.exception`?

src/radical/pilot/pmgr/launching/default.py
mtitov: kinda initial idea was that all possible exceptions will be handled here (key and value errors) Should we still assume that it could be non-convertible string variable, e.g., `None`? or just skip it and handle it as `int` only value?

p.s. also this part is how it was treated originally, thus we didn't force to set `system_architecture['smt']`, which was used at a certain cases. With the current setup it is required now. Will update it

src/radical/pilot/pmgr/launching/default.py
mtitov: I think we can skip `try..except` and follow your proposal above
(btw `debug` was used since it wasn't a required parameter, and we didn't enforce it)

src/radical/pilot/pmgr/launching/default.py
mtitov: @andre-merzky can you please look here? (that were initial changes in this branch)

src/radical/pilot/pmgr/launching/default.py
mtitov: @andre-merzky can you also please look here ? (from initial changes as well)

src/radical/pilot/pmgr/launching/default.py
mtitov: updated (added extra check, since if there is no env var, we need to check that there was value from config)

src/radical/pilot/pmgr/launching/default.py
andre-merzky: Ack,   that can be removed as staging moved out of here.

src/radical/pilot/pmgr/launching/default.py
```diff
@@ -536,7 +538,8 @@ def _start_pilot_bulk(self, resource, schema, pilots):
             tgt_dir = os.path.dirname(tgt)
 
             if tgt_dir.startswith('..'):
-              # raise ValueError('staging tgt %s outside pilot sbox: %s' % (ft['tgt'], tgt))
+              # raise ValueError('staging tgt %s outside pilot sbox: %s'
+              #                  % (ft['tgt'], tgt))
```
andre-merzky, line 542: We probably can remove this altogether by now...

src/radical/pilot/pmgr/launching/default.py
andre-merzky: Thanks for the ping!  We don't have any systems where this is still needed, and, more importantly, pip distribution was fixed to not need this anymore in recent releases.  The assumption here is that no system will install an older version of pip anymore.
The underlying problem was that some pip installations did not manage to fetch pypi host certs during installation and thus could not fetch from pypi...

andre-merzky: the check for `shared_filesystem` is needed for condor.  Well, I don't have any idea in what state our condor support is really as we have not exercised this in the last 5 years or so, but still.
Condor compute nodes have no shared file system with the condor login nodes (not sure about node type terminology, but you get the point), and we thus need to stage in two steps: once from the client side to the login node, and once from the login node to the compute node (each compute node in Condor runs it's own agent IIRC).  That is not needed on clusters which have a shared fs.

andre-merzky (commented):
Thanks for the iteration, @mtitov!

One thing I realized is that we lost the default pilot size of 1 core.  We can either leave it like that, or switch the default to one node.  I would probably prefer the latter, but we cannot do that in the schema, but have to correct that in the launcher I think.  What's your take on this?

mtitov:
> One thing I realized is that we lost the default pilot size of 1 core. We can either leave it like that, or switch the default to one node. I would probably prefer the latter, but we cannot do that in the schema, but have to correct that in the launcher I think. What's your take on this?

@andre-merzky actually I we can keep a corresponding value in `PilotDescription`: `cores: 1` (with `node=1` we cannot guarantee that there is a `cores_per_node`) - with this we would need to update `_verify` method - user can set `cores` and `nodes`, but if `nodes` is set then `cores` will be ignored.

src/radical/pilot/pmgr/launching/default.py
mtitov: you've updated this part, should it still be removed?

src/radical/pilot/pmgr/launching/default.py
andre-merzky: `cacerts`: good catch, will clean up.

shared FS: right, good point.  Probably would be good to fix even if we can't test - let me do that.

src/radical/pilot/pmgr/launching/default.py
andre-merzky: yes, I think so.

andre-merzky:
> > One thing I realized is that we lost the default pilot size of 1 core. We can either leave it like that, or switch the default to one node. I would probably prefer the latter, but we cannot do that in the schema, but have to correct that in the launcher I think. What's your take on this?
> 
> @andre-merzky actually I we can keep a corresponding value in `PilotDescription`: `cores: 1` (with `node=1` we cannot guarantee that there is a `cores_per_node`) - with this we would need to update `_verify` method - user can set `cores` and `nodes`, but if `nodes` is set then `cores` will be ignored.

fair point, but it feels wrong to me - in practice, one core will  on most machines result in one node anyway, so the default is a bit misleading then.  Given that there is only one resource where we can't specify a useful `cores_per_node` (amarel, which we don't really use either), I would not mind using the `1 node` default still and raise an error when we don't have `cores_per_node`, just like we do when `nodes` is defined in the PD.

src/radical/pilot/pmgr/launching/default.py
andre-merzky: Cleaned it out...

src/radical/pilot/pmgr/launching/default.py
andre-merzky: done

mtitov:
@andre-merzky with current changes it is enforced to set `jd.system_architecture` (`jd = rs.job.Description()`, [link](https://github.com/radical-cybertools/radical.pilot/blob/f0a355ad8ce6f278b07dc450940c77a6eba3ea0f/src/radical/pilot/pmgr/launching/default.py#L881-L886)), but not all SAGA adaptors support attribute `rs.job.SYSTEM_ARCHITECTURE`, e.g.
```
  File "/ve_exaworks/lib64/python3.6/site-packages/radical/saga/job/service.py", line 295, in create_job
    raise se.BadParameter._log (self._logger, msg)
radical.saga.exceptions.BadParameter: 'JobDescription.SystemArchitecture' ({'smt': 1}) not supported by 
radical.saga.adaptors.shell_job (/ve_exaworks/lib64/python3.6/site-packages/radical/saga/job/service.py 
+295 (create_job)  :  raise se.BadParameter._log (self._logger, msg))
```

https://github.com/radical-cybertools/radical.saga/blob/671dd9563228cad58580017963e573f83b91cf54/src/radical/saga/adaptors/shell/shell_job.py#L158-L175

thus the question - should we set `SystemArchitecture` for all SAGA adaptors or not enforce it here?

andre-merzky:
> thus the question - should we set `SystemArchitecture` for all SAGA adaptors or not enforce it here?

Oh, good point!  We should not enforce it here - the SAGA perspective is that an error MUST be raised on unsupported attributes, so it's doing the right thing here...

mtitov:
@andre-merzky done! (cannot put you as reviewer, since you are creator of this PR)

mtitov (approved):

andre-merzky:
@mtitov : can you please have a quick look at these two commits to make sure I am not doing anything stupid here?  Thanks!
d0d12ba927 23bf5e66dc

src/radical/pilot/pilot_description.py
```diff
@@ -257,6 +262,14 @@ def _verify(self):
         if not self.get('resource'):
             raise ValueError("Pilot description needs 'resource'")
 
+        if self.get('nodes'):
+
+            if self.get('cores'):
+                raise ValueError("Pilot description needs 'cores' *or* 'nodes'")
+
+            if self.get('gpus'):
+                raise ValueError("Pilot description needs 'cores' *or* 'nodes'")
```
mtitov, line 271: just a side note - maybe combine it? -> `if self.get('cores') or self.get('gpus'):`
p.s. in gpus-part, error message points to `cores`

andre-merzky:
Finally merged :-D

PR #2332, opened 2021-02-18
Updating resource configs

AymenFJA:
This pull request update/add: 

1.  Updating the `comet_mpirun` config  (`module reset` is not supported and switched to `purge`).
2.  Adding `xsede.bridges2`  entry to `example/config.json`


andre-merzky (approved):
Thanks Aymen, that looks good.

PR #2334, opened 2021-02-24
Fix/bridges integation test

iparask:
This PR removes the integration test for Bridges and goes for Bridges2. Some additional changes were made to make sure the tests run on Comet as well.


tests/integration_tests/test_config/resources.json
```diff
@@ -1,16 +1,15 @@
 
 {
     "comet": {
-      "cpus_on_node": 24,
-      "gpus_on_node": 0,
+      "cpus_per_node": 24,
```
andre-merzky, line 4: We use `cores_per_node` otherwise in RP - why `cpus_per_node` here?  I actually like that better, but it should be consistent maybe?

andre-merzky (approved):
Apart from that one non-semantic comment, this looks good to me - thanks!

tests/integration_tests/test_config/resources.json
iparask: Sorry. I will fix it.

tests/integration_tests/test_lm/test_mpirun.py
```diff
@@ -16,27 +16,30 @@ class TestTask(TestCase):
     # --------------------------------------------------------------------------
     #
     def setUp(self) -> dict:
```
mtitov, line 18: I'd like to propose here to start to fix method `setUp` within `TestCase`-based classes. This method shouldn't return anything, rather it prepares the instance of the class for the coming test ([link](https://docs.python.org/3.7/library/unittest.html#unittest.TestCase.setUp)) and it is called before each test (thus no need to call it manually)
```python
    def setUp(self):
        path = os.path.dirname(__file__) + '/../test_config/resources.json'
        resources = ru.read_json(path)
        hostname = socket.gethostname()

        for host in resources.keys():
            if host in hostname:
                self.host = host
                self.resource = resources[host]
                break
```
but if that information will be used for every test (within the test class), then corresponding class setup method `setUpClass` ([link](https://docs.python.org/3.7/library/unittest.html#unittest.TestCase.setUpClass)) should be used, which is called once before all tests start.
```python
    @classmethod
    def setUpClass(self):
        path = os.path.dirname(__file__) + '/../test_config/resources.json'
        resources = ru.read_json(path)
        hostname = socket.gethostname()

        for host in resources.keys():
            if host in hostname:
                cls.host = host
                cls.resource = resources[host]
                break
```
(*) same applied for `tearDown` and `tearDownClass` methods (which could be used for cleaning after each test and after all tests accordingly)

tests/integration_tests/test_lm/test_mpirun.py
```diff
@@ -16,27 +16,30 @@ class TestTask(TestCase):
...
+        component.name = 'mpirun'
+        component._log = mocked_Logger
+        component._cfg = ru.Munch({'resource': host})
         component.env_removables = []
         component._configure()
+
         self.assertEqual(component.launch_command, cfg['mpirun_path'])
-        self.assertEqual(component.mpi_flavor, cfg['mpi_flavor'])
         self.assertEqual(component.mpi_version, cfg['mpi_version'])
+        self.assertEqual(component.mpi_flavor, cfg['mpi_flavor'])
```
mtitov, line 42: would be better to have `cfg.get(...)`, thus if resource config doesn't have that key, we wouldn't have `KeyError`, but rather `AssertError`.

(*) here `cfg` is `dict`, but if it would be `Munch`/`Config`-based then it would work fine without method `get()`

tests/integration_tests/test_lm/test_ssh.py
```diff
@@ -27,10 +27,11 @@ def setUp(self) -> dict:
     # --------------------------------------------------------------------------
     #
     @mock.patch.object(SSH, '__init__',   return_value=None)
-    def test_configure(self, mocked_init):
+    @mock.patch('radical.utils.Logger')
+    def test_configure(self, mocked_init, mocked_Logger):
         cfg = self.setUp()
         component = SSH(name=None, cfg=None, session=None)
-        component._log = ru.Logger('dummy')
+        component._log = mocked_Logger
```
mtitov, line 34: I like this approach, could it be extended to Slurm test as well?

tests/integration_tests/test_lm/test_mpirun.py
iparask: Not so sure. `cfg` here is the test config and not the resource config. It may be better if I name it `results` or something. If an entry is missing means that the test is missing a test case.

tests/integration_tests/test_lm/test_mpirun.py
iparask: Let me check this out and get back at you.

mtitov (approved):
LGTM

PR #2337, opened 2021-03-02
Updates for PRTE2 LM

mtitov:
- updated `prun` command (fixed threads handling);
- reorganized multi-DVM start process (every following DVM is launched only after the previous one sets the status "ready");
- list of nodes per partition (i.e., per DVM) is not kept in `task['slots']['lm_info']['partitions']`, but in `ResourceManager.partitions` (which is propagated to `Scheduler._rm_partitions`);
- `task['slots']` is extended with the attribute `partition_id`, which is set in a scheduler;


src/radical/pilot/agent/launch_method/prte2.py
```diff
@@ -239,43 +241,35 @@ def _start_dvm(dvm_id, dvm_size, dvm_ready_flag):
...
                 'nodes'  : [node[1] for node in node_list],
-                'details': {'dvm_uri': _start_dvm(_dvm_id, vm_size, dvm_ready)}}
-
-        # check that all DVMs are ready
-        for _dvm_id, _dvm_ready in enumerate(dvm_ready_list):
-
-            if not _dvm_ready.is_set():
-
-                partition_id = str(_dvm_id)
+                'details': {'dvm_uri': _dvm_uri}}
```
mtitov, line 252: @andre-merzky with proposed changes not to keep nodes list per partition in `lm_info`, should this be updated as following?
```python
            partitions[partition_id] = {
                'nodes'  : [node[1] for node in node_list],
                'dvm_uri': _dvm_uri}
```
thus list of nodes will be moved to `ResourceManager.partitions` and the rest is as "details" per partition, which will be stored in `lm_info` as
```python
lm_info = {
    ...
    'partitions': {<partition_id>: {'dvm_uri': <dvm_uri>}}
    ...
}
```

andre-merzky: :+1:

docs/source/events.md
```diff
@@ -138,15 +138,16 @@ indication on event ordering *within each individual component*.
     cmd                 : command received from pmgr                 (uid: pilot, msg: command, [API])
     get                 : tasks   received from task manager         (uid: pilot, msg: 'bulk size: %d')
     get                 : task    received from task manager         (uid: task)
-    dvm_start           : DVM startup by launch method               (uid: pilot) [CFG-R])
-    dvm_ok              : DVM startup completed                      (uid: pilot) [CFG-R])
-    dvm_fail            : DVM startup failed                         (uid: pilot) [CFG-R])
-    dvm_stop            : DVM stopped                                (uid: pilot) [CFG-R])
+    dvm_start           : DVM startup by launch method               (uid: pilot, msg: 'dvm_id=%d') [CFG-DVM])
+    dvm_uri_ok          : DVM URI is set successfully                (uid: pilot, msg: 'dvm_id=%d') [CFG-DVM])
```
andre-merzky, line 142: I can see that `dvm_ok` does not match - but maybe shorter as `dvm_uri`?

mtitov: ok, will update accordingly

andre-merzky (approved):
Thanks!

PR #2338, opened 2021-03-02
add_pilot() also accepts pilot dict

andre-merzky:
@iparask : let me know if this does or does not work.

### DIFF
--- a/src/radical/pilot/task_manager.py
+++ b/src/radical/pilot/task_manager.py
@@ -795,7 +795,7 @@ class TaskManager(rpu.Component):
 
     # --------------------------------------------------------------------------
     #
-    def list_units(self, uids=None):
+    def list_units(self, uids=None):                                      # noqa
         '''
         deprecated - use `list_tasks()`
         '''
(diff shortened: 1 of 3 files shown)


src/radical/pilot/task_manager.py
```diff
@@ -695,19 +695,24 @@ def add_pilots(self, pilots):
...
-                self._pilots[pid] = pilot
 
-                # subscribe for state updates
-                pilot.register_callback(self._pilot_state_cb)
+                if isinstance(pilot, dict):
+                    pilot_dict = pilot
+                else:
+                    pilot_dict = pilot.as_dict()
+                    # real object: subscribe for state updates
+                    pilot.register_callback(self._pilot_state_cb)
```
iparask, line 709: If we pass a pilot object it also registers the callback, but not when we pass a dict. Does this mean that when an application passes a dict, it has to make sure to register TMGR's pilot callback?

andre-merzky: That is correct.  The callback is, after all, on the pilot object which we simply do not have in that case.  The rationale is that the API should normally use an object, and that is what we document.  Using the dict only happens to work when you know RP internals - but in that case you have other means to get state notifications (like, subscribing to the state pubsub).

iparask:
It works for me. Let me finish its test and then we can merge. I will use this branch for developing in EnTK

PR #2341, opened 2021-03-03
Test/summit integration

iparask:
We have integration tests for Summit! The LSF test makes sure that the resource manager configuration is correct. The JSRUN test verifies the path for jsrun based on the modules we load, and creates and run a jsrun command based on what the launch method creates.

I opened the PR to show my approach. Next step is to integrate the test cases Vivek created.

@mtitov, is the use of `setUpClass` according to your understanding and request in PR #2334?


tests/integration_tests/batch_scripts/summit.sh
```diff
@@ -0,0 +1,50 @@
...
+module unload py-pip
+module unload py-virtualenv
+module unload py-setuptools
+module load gcc/8.1.1
+module load zeromq/4.2.5
+module load python/3.7.0-anaconda3-5.3.0
+
+conda create -p testing python=3.7 pytest PyGithub -y -c conda-forge
+
+source activate $PWD/testing
```
mtitov, line 31: `conda activate $PWD/testing`

tests/integration_tests/test_lm/test_jsrun.py
```diff
@@ -0,0 +1,82 @@
...
+    def setUpClass(cls):
+        path = os.path.dirname(__file__) + '/../test_config/resources.json'
+        resources = ru.read_json(path)
+        hostname = 'summit'
+
+        for host in resources.keys():
+            if host in hostname:
+                cls.host = host
+                cls.resource = resources[host]
+                break
```
mtitov, line 28: maybe update it as `cls.resource = resources['summit']` (since you know which configuration to pick)

tests/integration_tests/test_lm/test_jsrun.py
```diff
@@ -0,0 +1,82 @@
...
+        component._log = mocked_Logger
+        component._configure()
+        self.assertEqual(component.launch_command, self.resource['jsrun_path'])
+
+    # --------------------------------------------------------------------------
+    #
+    @mock.patch.object(JSRUN, '__init__', return_value=None)
+    @mock.patch('radical.utils.Logger')
+    def test_command(self, mocked_init, mocked_Logger):
+        cfg = {
```
mtitov, line 45: this will be moved to a separate json file, right?

iparask: Yes

tests/integration_tests/batch_scripts/summit.sh
iparask: `conda activate` produces this error:
```
bash-4.2$ conda activate $PWD/testing

CommandNotFoundError: Your shell has not been properly configured to use 'conda activate'.
To initialize your shell, run

    $ conda init <SHELL_NAME>

Currently supported shells are:
  - bash
  - fish
  - tcsh
  - xonsh
  - zsh
  - powershell

See 'conda init --help' for more information and options.

IMPORTANT: You may need to close and restart your shell after running 'conda init'.
```

I did `conda init` but still I got the same error. `source` on the other hand correctly activates the VE. I am not sure if it makes sense to change my .`bashrc` to support `conda` since `source` works. Do you have another idea?

mtitov: try to add this line `eval "$(conda shell.posix hook)"` before the activation line

tests/integration_tests/test_lm/test_jsrun.py
iparask: I created a task json file with the task and expected output. This is then used for executing the integration test.

tests/integration_tests/test_lm/test_jsrun.py
iparask: I think this is fixed.

tests/integration_tests/batch_scripts/summit.sh
iparask: I added it before the VE creation and it seems to work.

tests/integration_tests/test_rm/test_lsf.py
```diff
@@ -0,0 +1,62 @@
...
+#
+class TestTask(TestCase):
+
+    # --------------------------------------------------------------------------
+    #
+    @classmethod
+    def setUpClass(self) -> dict:
+        path = os.path.dirname(__file__) + '/../test_config/resources.json'
+        resources = ru.read_json(path)
+        hostname = 'summit'
```
mtitov, line 24: here is the same as for jsrun test, everything else looks good!
```python
    @classmethod
     def setUpClass(cls):
         path = os.path.dirname(__file__) + '/../test_config/resources.json'
         resources = ru.read_json(path)

         cls.host = 'summit'
         cls.resource = resources[cls.host]
```
btw, if `cls.host` is not used later then no need to keep this class attribute

mtitov (approved):
LGTM

PR #2343, opened 2021-03-08
New resource - Andes@ORNL

mtitov:
since compute nodes don't have a connection to outside then virtual environment should be pre-set
```
source /sw/andes/python/3.7/anaconda-base/etc/profile.d/conda.sh
conda create -y -p $HOME/ve.rp python=3.7
pip install radical.pilot
```
**EDIT**: any virtual env could be configured (either conda env or python virtualenv as `python3 -m venv <path>`)


andre-merzky (approved):
I'm fine with the Andes changes - but are you  sure about the summit ones?  If so, this can be merged.

mtitov:
@andre-merzky thanks!! I re-checked it and yes, wrong line for conda activation, everything else is correct.

p.s. codecov got delayed, if you don't mind can you please merge this PR

mtitov:
one more comment: should we do for Summit - use virtual env from where RP is launched? (I forgot that we have it and just tested with Andes and updated it accordingly)
```
         "python_dist"                 : "local",
         "virtenv_mode"                : "local",
```
and remove `rp_version`, `virtenv_dist` which will be set automatically

src/radical/pilot/configs/resource_ornl.json
```diff
@@ -1,93 +1,55 @@
...
-        "rp_version"                  : "local",
-        "virtenv_mode"                : "use",
-        "virtenv"                     : "/lustre/atlas2/%(pd.project)s/world-shared/ve.rp.titan.2018.05.23",
+        "agent_launch_method"         : "SRUN",
+        "task_launch_method"          : "SRUN",
+        "mpi_launch_method"           : "SRUN",
+        "pre_bootstrap_0"             : ["source /sw/andes/python/3.7/anaconda-base/etc/profile.d/conda.sh"],
+        "valid_roots"                 : ["$MEMBERWORK/"],
+        "default_remote_workdir"      : "$MEMBERWORK/%(pd.project)s",
+        "python_dist"                 : "local",
```
andre-merzky, line 22: Sorry, didn't see this on the first read.  `local` will only reliably work if
  - the client virtualenv is on a shared filesystem (usable from the compute nodes)
  - the client virtualenv can be activated using the module loads above (so, if the user happens to use the exact same modules)
I would find this too risky for a production config (which we let end users use), and think it's safer to let the bootstrapper set up the env.  The overhead is not large enough to matter in production.  What do you think?
Same for summit (just more so ;-) )

mtitov: all that points work and we don't force user to use a specific type of virtenv or a specific name

> think it's safer to let the bootstrapper set up the env

this case requires to have a pre-set virtenv (since there is no connection to outside from compute nodes), but I would propose to extend using such approach for other similar cases (to be discussed during devel-meeting and will be a separate ticket/PR)

PR #2346, opened 2021-03-15
python 2 -> 3 fix for shebangs

andre-merzky:
how did we miss this?


mtitov:
these changes are not much necessary (since within VE it is fine to refer to python as `python` or `python3`, and the second approach is applicable since we restrict to use python >=3.6), but ok to have them

mtitov (approved):

andre-merzky:
> these changes are not much necessary

We also need to support when not running in a VE, and on system level, `python` frequently referes to `python2`...

PR #2350, opened 2021-03-16
fix ve isolation for prep_env

andre-merzky:
also make prep_env async and let scheduler wait as needed.  And finally install module dependencies by default now.


bin/radical-pilot-create-static-ve
```diff
@@ -5,38 +5,42 @@ PREFIX=         # full path or virtenv name
...
 fi
 
+echo
+echo "script  : $SCRIPT"
+echo "type    : $TYPE"
+echo "prefix  : $PREFIX"
+echo "version : $VERSION"
+echo "modules : $MODULES"
+echo "defaults: $DEFAULTS"
+echo
```
mtitov, line 41: maybe merge this with the printout at the bottom of this file?

bin/radical-pilot-create-static-ve
```diff
@@ -187,18 +191,18 @@ then
 fi
 
 echo -n "create  virtualenv "
-stdbuf -oL $VIRTENV_CMD -p "python$VERSION" "$PREFIX" | progress
+stdbuf -oL $VIRTENV_CMD --no-site-packages -p "python$VERSION" "$PREFIX" | progress
```
mtitov, line 194: with this one I have an error: `virtualenv: error: unrecognized arguments: --no-site-packages`, it seems like this option was dropped, skip it?

src/radical/pilot/agent/scheduler/base.py
```diff
@@ -653,18 +653,29 @@ def _schedule_waitpool(self):
...
+        to_wait = list()
+        to_test = list()
+
+        for task in tasks:
+            named_env = task['description'].get('named_env')
+            if named_env:
+                if os.path.exists('%s.ok' % named_env):
+                    to_test.append(task)
+                else:
+                    to_wait.append(task)
```
mtitov, line 666: need to limit IO calls, thus to have corresponding flags for `named_env`
as an example:
```python
        named_envs = {}
        for task in tasks:
            named_env = task['description'].get('named_env')
            if named_env:
                if named_env not in named_envs:
                    named_envs[named_env] = os.path.exists('%s.ok' % named_env)
                if named_envs[named_env]: to_test.append(task)
                else                    : to_wait.append(task)
```

src/radical/pilot/agent/scheduler/base.py
```diff
@@ -653,18 +653,29 @@ def _schedule_waitpool(self):
...
+
+        to_test.sort(key=lambda x:
                 (x['tuple_size'][0] + x['tuple_size'][2]) * x['tuple_size'][1],
                  reverse=True)
 
         # cycle through waitpool, and see if we get anything placed now.
-        scheduled, unscheduled = ru.lazy_bisect(tasks,
+        scheduled, unscheduled = ru.lazy_bisect(to_test,
```
mtitov, line 673: I got a little bit confused, if there is no tasks with `named_env` set, then all that tasks will not reach here?

src/radical/pilot/agent/scheduler/base.py
```diff
@@ -720,6 +731,18 @@ def _schedule_incoming(self):
...
+            # FIXME: This is a slow and inefficient way to wait for named VEs.
+            #        The semantics should move to the upcoming eligibility
+            #        checker
+            # FIXME: Note that this code is duplicated in _schedule_waitpool
+            named_env = task['description'].get('named_env')
+            if named_env:
+                if not os.path.exists('%s.ok' % named_env):
+                    to_wait.append(task)
+                    self._log.debug('delay %s, no env %s',
+                                    task['uid'], named_env)
```
mtitov, line 743: same here with "flags" as above

mtitov (approved):
LGTM (I made updates according to comments)

andre-merzky:
Thanks @mtitov !

PR #2351, opened 2021-03-16
ensure that RU lock names are unique

andre-merzky:
The IDs will otherwise collide when multiple mgrs are used in the same session.


src/radical/pilot/utils/component.py
```diff
@@ -419,9 +419,9 @@ def __init__(self, cfg, session):
         self._workers    = dict()       # methods to work on things
         self._publishers = dict()       # channels to send notifications to
         self._threads    = dict()       # subscriber and idler threads
-        self._cb_lock    = ru.RLock('comp.cb_lock.%s' % self._name)
+        self._cb_lock    = ru.RLock('%s.cb_lock' % self._name)
                                         # guard threaded callback invokations
-        self._work_lock  = ru.RLock('comp.work_lock.%s' % self._name)
+        self._work_lock  = ru.RLock('%s.work_lock' % self._name)
```
mtitov, line 424: maybe to use `self._uid` instead of `self._name`?

src/radical/pilot/task_manager.py
```diff
@@ -101,12 +101,12 @@ def __init__(self, session, cfg='default', scheduler=None):
...
         self._pilots      = dict()
-        self._pilots_lock = ru.RLock('tmgr.pilots_lock')
+        self._pilots_lock = ru.RLock('%s.pilots_lock' % self._uid)
         self._uids        = list()   # known task UIDs
         self._tasks       = dict()
-        self._tasks_lock  = ru.RLock('tmgr.tasks_lock')
+        self._tasks_lock  = ru.RLock('%s.tasks_lock' % self._uid)
         self._callbacks   = dict()
-        self._cb_lock     = ru.RLock('tmgr.cb_lock')
+        self._cb_lock     = ru.RLock('%s.cb_lock' % self._uid)
```
mtitov, line 109: this one will be overwritten in `rpu.Component`, should it be `self._tcb_lock`? (similar to PilotManager)

mtitov (approved):
LGTM!

andre-merzky (commented):
Thanks for those commits, @mtitov !

PR #2353, opened 2021-03-23
No nodes overlapping between different tags (Continuous scheduler)

mtitov:


src/radical/pilot/agent/scheduler/continuous.py
```diff
@@ -422,8 +423,12 @@ def schedule_task(self, task):
...
             # If a tag exists, continue to consider this node if the tag was
             # used for this node - else continue to the next node.
-            if tag is not None and tag in self._tag_history:
-                if node_uid not in self._tag_history[tag]:
+            if tag is not None:
+                if tag in self._tag_history:
+                    if node_uid not in self._tag_history[tag]:
+                        continue
+                # for a new tag check that nodes were not used for previous tags
+                elif node_uid in self._tagged_nodes:
```
mtitov, line 431: @andre-merzky along with these changes I would propose to have an extra attribute `exclusive` within `tags`: `td['tags'] = {'colocate': <tag>, 'exclusive': False}`

It would have default value as `True` (to use nodes assigned to a certain tag exclusively), and if it is ok to share nodes among different tags then to set it `False`

```python
                is_exclusive = td.get('tags', {}).get('exclusive', True)
                ...
                if ...
                elif is_exclusive and node_uid in self._tagged_nodes:
                    continue
```

mtitov:
@andre-merzky was it closed accidentally? (if I'm not mistaken we agreed to have it)

andre-merzky:
Oh, what did I do?  yes, accidentally - I thought this was a leftover branch of mine and cleaned it out...  Sorry for that!

mtitov:
@andre-merzky I did cleanup tests a little bit and added check for `exclusive` flag in tags. Ready for review

src/radical/pilot/agent/scheduler/continuous.py
```diff
@@ -422,9 +423,16 @@ def schedule_task(self, task):
...
+            if tag is not None:
+                if tag in self._tag_history:
+                    if node_uid not in self._tag_history[tag]:
+                        continue
+                # for a new tag check that nodes were not used for previous tags
+                else:
+                    # `exclusive` -> not to share nodes between different tags
+                    is_exclusive = td.get('tags', {}).get('exclusive', True)
+                    if is_exclusive and node_uid in self._tagged_nodes:
+                        continue
```
andre-merzky, line 435: Should we maybe set the default to `False`?  If there are more tags than nodes, the execution will stall otherwise.  Maybe that's also a check we should perform (when we add new tags to the exclusive list):
```py
if len(self._tagged_nodes) >= len(self._nodes):
  task.state = rp.FAILED
  ...
```
although that would only work when all nodes in the `_tagged_nodes` are tagged for exclusivity - which we don't know.  Any idea?

Actually, what happens then - will `_iterate_nodes` cycle around and we are stuck in this loop forever?

src/radical/pilot/agent/scheduler/continuous.py
```diff
@@ -422,9 +423,16 @@ def schedule_task(self, task):
...
-                if node_uid not in self._tag_history[tag]:
-                    continue
+            if tag is not None:
+                if tag in self._tag_history:
+                    if node_uid not in self._tag_history[tag]:
+                        continue
+                # for a new tag check that nodes were not used for previous tags
+                else:
+                    # `exclusive` -> not to share nodes between different tags
+                    is_exclusive = td.get('tags', {}).get('exclusive', True)
```
andre-merzky, line 433: I think this line should be:
```py
    is_exclusive = tag.get('exclusive', True)
```

andre-merzky (commented):
Some comments - thanks!

src/radical/pilot/agent/scheduler/continuous.py
mtitov: good points, thanks! didn't think about that... I agree to keep default value as `False` and will think more about exception case (first thought: if there are not more exclusive nodes for new tags, then let it go, but print a warning in logs)

src/radical/pilot/agent/scheduler/continuous.py
mtitov: within the code `tag` is a string, but since `'tags'` is in scheme, then it always will be in task description -> `is_exclusive = td['tags'].get('exclusive', True)`. 

Or you want move it inside `'colocate'`? (`td = {..'tags': {'colocate': {'name': <tag_name>, 'exclusive': False}}..}`) I would like to keep it as a separate key within `tags`

andre-merzky: We have in line 383:
```py
tag = td.get('tag')
```
and thus the first part of 
```py
td.get('tags', {}).get('exclusive', True)
```
kind of duplicates it.  But you are right - the former one would need a dict default value then.

Anyway, not important, this comment was not meant to mean much more than this :-)

mtitov: right, that is `tag` (of `str` type), not `tags` :) and that `tag` attribute was temporary and is close to become obsolete

andre-merzky (approved):
LGTM - thanks!

PR #2354, opened 2021-03-24
Feature/outside triggered action

iparask:
This is ready for Summit, Bridges, and Comet. Please let me know if you see anything I missed. Also, I've set up the corn job to run every month and check if there was a trigger a week before the date it runs. We can change that as we please.

### DIFF
--- a/tests/bin/radical-pilot-test-verify
+++ b/tests/bin/radical-pilot-test-verify
@@ -1,0 +1,33 @@
+#!/usr/bin/env python3
+
+import requests
+import sys
+
+from datetime import datetime
+from argparse import ArgumentParser
+
+def main():
+
+    parser = ArgumentParser()
+    parser.add_argument('--yaml', '-y', type=str, required=True,
+                        help='Name of workflow YAML file')
+    parser.add_argument('--days', '-d', type=int, required=True,
+                        help='Number of days from last triggered check')
+    args = parser.parse_args()
+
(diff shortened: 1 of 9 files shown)


andre-merzky (commented):
The workflow yml scrtipts are identical apart from the resource name.  Would it be possible to reduce that to one file with an argument?  Same for the `batch_scripts/*sh`.

tests/bin/radical-pilot-test-verify
```diff
@@ -0,0 +1,51 @@
...
+        finished = last_run['updated_at']
+        status   = last_run['conclusion']
+
+        now      = datetime.now()
+        last     = datetime.strptime(finished, '%Y-%m-%dT%H:%M:%SZ')
+        diff     = now - last
+
+        if status == 'success' and diff.days > args.days:
+            sys.exit(1)
+        else:
```
andre-merzky, line 38: this implies:
```py
if status == failed:
     exit(0)
```
Is that intentional?

iparask: Yes, this is intentional. I wanted to have the github action fail. For this to happen, an executable needs to exit with a code different from 0.

iparask:
> The workflow yml scrtipts are identical apart from the resource name. Would it be possible to reduce that to one file with an argument? Same for the `batch_scripts/*sh`.

It can be. The question I have is how do you know which resource failed. I need to see if we can have multiple badges triggered by a single workflow and change their state based on the workflow state. This is currently simpler and I think a good first version.

PR #2356, opened 2021-03-26
remove add_resource_config

andre-merzky:
this is currently broken: the added configs will not get communicated to the pmgr.launcher where they are used.  Also, this is really hard to fix: we would need to distinguish cases where the launcher already started / where it is not yet started; where it opened it's own session instance / where it did not, etc.  This is a rabbit hole of race conditions waiting to happen...

The code likely has last worked before the Component base class was introduced.  The fact that until now (i.e., until @eirrgang ) nobody stumbled over that problem motivates the removal.

PS.: the unit test only checked if the current session has the added config registered, not if it is used during launch - that would only be covered by an integration test which we don't have to programmatic resource configs...


mtitov:
@andre-merzky can you please describe more why it should be removed? I through this is a convenient way to prepare a resource config for the session (and should be executed before `PilotDescription` is set). It could be used instead of creating resource config file manually (in as user space)

andre-merzky:
Sorry for the lack of details.

The method changes the set of resources a session sees - but the resource configs are not used in the same process where the API changes that list, but instead are used in the `pmgr.launching` component process.  That process will not see any changes done by the user, so the changes are basically useless.

The mechanism worked in the past when we just forked the process, and the session maintained state.  By now we use `bin/radical-pilot-component` to bootstrap new component processes from scratch, and as such create new session instances - those won't pick up the resource configs from the original session, and the whole thing is thus broken right now.

We may want to re-introduce the feature at some point - but then it will need a synchronization mechanism between the API session and all running (and yet to be started) components, so the code will have to look quite different I'm afraid.

mtitov:
@andre-merzky thank you for explanation! so, yeah, I see the point

mturilli (approved):
LGTM. Should we open a discussion about this capability so not to forget about it?

PR #2358, opened 2021-03-26
removing cuda env from lm base

iparask:
I'm suggesting removing `CUDA_VISIBLE_DEVICES` from the default environment of launch methods. The specific variable is set in the scheduler [`_handle_cuda`](https://github.com/radical-cybertools/radical.pilot/blob/devel/src/radical/pilot/agent/scheduler/base.py#L889) method and added to a task's environment when needed.


andre-merzky (approved):
That makes sense, thanks!

mtitov (approved):
agree

iparask:
And I did not think to check the tests. The test for `ssh` launch method, fails when there is an environment variable in the task description. It does not add a space correctly. Do you know how to fix it without messing everything up?

mtitov:
@iparask I've added a small fix, all looks good now

mtitov:
@andre-merzky that `1` comes from `launch_script_hop` parameter in SSH LM ([link](https://github.com/radical-cybertools/radical.pilot/blob/538f98741a2604e418e9cdd81230fe6c32b62dea/src/radical/pilot/agent/launch_method/ssh.py#L90)) and it is set manually in test [here](https://github.com/radical-cybertools/radical.pilot/blob/538f98741a2604e418e9cdd81230fe6c32b62dea/tests/unit_tests/test_lm/test_ssh.py#L71)

andre-merzky:
> @andre-merzky that `1` comes from `launch_script_hop` parameter in SSH LM ([link](https://github.com/radical-cybertools/radical.pilot/blob/538f98741a2604e418e9cdd81230fe6c32b62dea/src/radical/pilot/agent/launch_method/ssh.py#L90)) and it is set manually in test [here](https://github.com/radical-cybertools/radical.pilot/blob/538f98741a2604e418e9cdd81230fe6c32b62dea/tests/unit_tests/test_lm/test_ssh.py#L71)

Ah, thanks

PR #2360, opened 2021-03-29
Fixed launch command for PRTE2 LM

mtitov:
With oncoming version of prrte,`prun` command with `--map-by hwthread` is not supported for certain machines* and raises the following error:
```
We cannot bind to hardware threads if we are not treating hardware
threads as independent CPUs since the lowest allocatable unit is
"core" by default. In order to bind to hardware threads, you must
therefore specify that we treat hardware threads as independent CPUs
by adding the "HWTCPUS" modifier to the "--map-by" directive.

The combination "--map-by core:hwtcpus --bind-to hwthread" or
"--map-by :hwtcpus --bind-to hwthread" will result in one process
being placed on each core, with the process bound to the first
hardware thread in that core.

In contrast, a case such as "--map-by package:hwtcpus --bind-to hwthread"
would result in each process being bound to successive hardware threads
on each package.
```
(*) doesn't work for Frontera, but works for Summit

`prun` command with `--map-by node:HWTCPUS` works for both machines

### DIFF
--- a/src/radical/pilot/agent/launch_method/prte2.py
+++ b/src/radical/pilot/agent/launch_method/prte2.py
@@ -340,11 +340,8 @@ class PRTE2(LaunchMethod):
         partition_id = slots.get('partition_id') or partitions.keys()[0]
         dvm_uri = '--dvm-uri "%s"' % partitions[partition_id]['dvm_uri']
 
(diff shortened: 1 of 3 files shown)


andre-merzky (approved):
Thanks!

PR #2365, opened 2021-04-02
Stampede2 migrates to work2 filesystem

iparask:
Stampede2 migrates from WORK to WORK2. This PR changes the filesystem that RP is using.


andre-merzky (approved):
Thanks @iparask!

PR #2369, opened 2021-04-06
fix missing event updates

andre-merzky:
the bulk cb PR (#1872) introduced a race that caused state updates to go missing when occuring in short order - this PR fixes that problem


mtitov (approved):
LGTM!

PR #2372, opened 2021-04-07
Remove obsolete attribute `tag` in `TaskDescription`

mtitov:


andre-merzky (approved):
I took the liberty of renaming `tag` to `colo_tag` to make room for other tag types.

PR #2375, opened 2021-04-09
Refactoring continuous scheduler unit test

iparask:
All silent conflicts were resolved.


mtitov (approved):
LGTM! (sorry for massive changes..)

.github/workflows/python-app.yml
```diff
@@ -99,6 +109,7 @@ jobs:
         pip install pylint
     - name: Lint with pylint
       run: |
+        . testenv/bin/activate
```
andre-merzky, line 112: Doesn't this need  a `python -m venv testenv` also?

eirrgang: The testenv venv is created in the previous step: "Install dependencies"

PR #2376, opened 2021-04-15
Fixed example "07_..."

mtitov:


examples/07_shared_task_data.py
```diff
@@ -100,13 +100,14 @@
             # Here we don't use dict initialization.
             td = rp.TaskDescription()
             td.executable     = '/bin/echo'
-            td.arguments      = ['-c', 'input.dat', '%d' % i]
+            td.arguments      = ['$(cat input.dat)', '%d' % i]
```
andre-merzky, line 103: This will not work with `mpirun` and the like.  Should we maybe just run `/bin/cat input.dat`?

andre-merzky (approved):
proposal for simpler command

examples/07_shared_task_data.py
mtitov: ok, will do

PR #2388, opened 2021-05-06
executor errors should not affect task bulks

andre-merzky:
instead, errors should only affect the individual task on which the
error occured.  We also need to unschedule affected tasks.


mtitov (approved):
LGTM!

PR #2389, opened 2021-05-07
Hotfix/stager error

andre-merzky:
This PR avoids hangs on staging errors


src/radical/pilot/pilot_manager.py
```diff
@@ -486,7 +486,15 @@ def _pilot_staging_output(self, sds):
                                                 if sd['uid'] in uids]
 
         if rps.FAILED in sd_states:
-            raise RuntimeError('pilot staging failed')
+            errs = list()
+            for uid in self._active_sds:
+                if self._active_sds[uid].get('error'):
+                    errs.append(self._active_sds[uid]['error'])
```
mtitov, line 492: I don't see that these elements have key `'error'` - method `expand_staging_directives` (from module `staging_directives`) defines keys for `sds` elements

p.s. I see that `stager` provides this error message, but that method will not take it into account

andre-merzky: > p.s. I see that `stager` provides this error message, but that method will not take it into account

It does:  line 476 of the manager sends the SDs to the stager, and the manager then waits in line 482ff until the stager sends the SDs back to the manager (they are received in line 512).  At that point the `error` field in the SD was (possibly) set by the stager.

mtitov: at part that starts with L512 only `'state'` is updated (`self._active_sds[sd['uid']]['state'] = sd['state']`), unless there is something more...

p.s. I tried a small test, and see that stager gives the following error (stager:L77)
> file copy failed: /bin/cp: cannot stat '/home/matitov/radical.pilot.sandbox/rp.session.js-17-185.jetstream-cloud.org.matitov.018755.0000/pilot.0000/STDOUT.000000': No such file or directory

but that error wasn't returned to pilot manager and my script finished with the following error:
```
  File "test_run.py", line 118, in <module>
    'action': rp.TRANSFER} for fname in outs])
  File "/home/matitov/.miniconda3/envs/rp/lib/python3.7/site-packages/radical/pilot/pilot.py", line 660, in stage_out
    self._pmgr._pilot_staging_output(sds)
  File "/home/matitov/.miniconda3/envs/rp/lib/python3.7/site-packages/radical/pilot/pilot_manager.py", line 497, in _pilot_staging_output
    raise RuntimeError('pilot staging failed')
RuntimeError: pilot staging failed
```

andre-merzky: Yeah, you are right here of course - thanks, fixed now.

PR #2390, opened 2021-05-07
User/repex

andre-merzky:
This PR merges some smaller fixes and cleanups - all semantic changes of this branch trickled into devel already.


src/radical/pilot/agent/executing/popen.py
```diff
@@ -210,6 +210,10 @@ def spawn(self, launcher, task):
             launch_script.write('\n%s\n\n' % pprint.pformat(task['slots']))
 
         with open(launch_script_name, "w") as launch_script:
+
+            td = task['description']
+            tid = task['uid']
```
mtitov, line 215: In L192 there is already set `descr = task['description']`, so that could be used, and maybe move `tid` there as well?

andre-merzky: Thanks, I fixed that, and made var names a bit more consistent in that file.

mtitov (approved):
LGTM!

PR #2395, opened 2021-05-13
Resolve a "FIXME"

### LINKED ISSUE #2394: Session.__exit__() does not allow full Session.close() behavior.
There is a "FIXME" note at https://github.com/radical-cybertools/radical.pilot/blob/7c5cb936a0b0e86a2f766b99c066d0c4137a5968/src/radical/pilot/session.py#L221 that I will assume means "bug" for tracking purposes.

The note indicating that users should have access to the full `Session.close()` behavioral options when leaving a `with Session():` block.

This can be resolved by allowing users to (optionally) provide the `close()` key word arguments during Session creation. Then, when `close()` is called, it can use the stored options instead of the hard-coded options.

eirrgang:
Encapsulate kwargs handling for Session.close().

Allow kwargs for Session.close() to be provided when initializing the Session.
This allows a "FIXME" to be resolved. When used as a Python context manager,
the Session will have user input with which to override the default arguments
at close().

If key word arguments are provided both at instantiation and when calling
`close()`, the kwargs provided to `close()` take precedence.

An attempt is made to merge the new arguments with the previous arguments.
`close()` is called with `cleanup=True` but an unspecified *terminate*, and
`cleanup=False, terminate=False` was provided to `Session()`, a warning is
issued and the old value is overwritten with `terminate=True`.

Close options are evaluated for self-consistency during both
`Session.__init__()` and `Session.close()`

## Behavior changes

If the user provides inconsistent values for *cleanup* and *terminate*,
ValueError is raised.

Previously, if the user explicitly specified `cleanup=None`, this was
interpreted as `cleanup=True` instead of the default `cleanup=False`.

Fixes #2394

### DIFF
--- a/src/radical/pilot/session.py
+++ b/src/radical/pilot/session.py
@@ -233,7 +233,7 @@ class Session(rs.Session):
 
     # --------------------------------------------------------------------------
     #
-    def close(self, cleanup=None, terminate=None, download=None):
+    def close(self, **kwargs):
         '''
 
         Closes the session.  All subsequent attempts access objects attached to
@@ -259,19 +259,10 @@ class Session(rs.Session):
         self._prof.prof("session_close", uid=self._uid)
 
         # Merge kwargs with current defaults stored in self._close_options
-        self._close_options.update(
-            {
-                'cleanup': cleanup,
-                'download': download,
-                'terminate': terminate
-            }
-        )
+        self._close_options.update(kwargs)
         self._close_options.verify()  # in case to call for `_verify` method and to convert attributes
                                       # to their types if needed (but None value will stay if it is set)
 
-        del cleanup
-        del terminate
-        del download
         options = self._close_options
 
         for tmgr_uid, tmgr in self._tmgrs.items():
--- a/src/radical/pilot/session.py
+++ b/src/radical/pilot/session.py
@@ -76,7 +76,7 @@ class Session(rs.Session):
         Session is used as a Python context manager, such that close() is called
         automatically at the end of a ``with`` block.)
         '''
-        self._close_options = _CloseOptions(**close_options)
+        self._close_options = _CloseOptions(close_options)
         # NOTE: `name` and `cfg` are overloaded, the user cannot point to
         #       a predefined config and amend it at the same time.  This might
         #       be ok for the session, but introduces a minor API inconsistency.
@@ -261,19 +261,16 @@ class Session(rs.Session):
         self._prof.prof("session_close", uid=self._uid)
 
         # Merge kwargs with current defaults stored in self._close_options
-        if terminate is None and cleanup:
-            if not self._close_options.terminate:
-                warnings.warn(
-                    'Overriding *terminate=False* set during Session creation. '
-                    '*cleanup* implies *terminate*.'
-                )
-            terminate = True
-        self._close_options = dataclasses.replace(
-            self._close_options,
-            cleanup=cleanup,
-            terminate=terminate,
-            download=download
+        self._close_options.update(
+            {
+                'cleanup': cleanup,
+                'download': download,
+                'terminate': terminate
+            }
         )
+        self._close_options.verify()  # in case to call for `_verify` method and to convert attributes
+                                      # to their types if needed (but None value will stay if it is set)
+
         del cleanup
         del terminate
         del download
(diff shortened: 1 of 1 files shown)


src/radical/pilot/session.py
```diff
@@ -968,3 +985,29 @@ def excuse():
...
+        * **cleanup**   (`bool`):
+          Remove session from MongoDB (implies * terminate). (default False)
+        * **terminate** (`bool`):
+          Shut down all pilots associated with the session. (default True)
+        * **download** (`bool`):
+          Fetch pilot profiles and database entries. (default False)
+
+    """
+    cleanup: bool = False
+    # Warning: Previously, passing `cleanup=None` would reverse the default and
```
eirrgang, line 1004: Was this logic intentional? Is it okay to remove the distinction between unspecified *cleanup* and explicit *cleanup=None*? If not, the field is really an explicit Union rather than an implicit Optional, and it might be more appropriate to use either a TypedDict or just a simple dict without encapsulated logic.

src/radical/pilot/session.py
```diff
@@ -968,3 +985,29 @@ def excuse():
...
+    # result in cleanup == True.
+
+    terminate: bool = True
+    download: bool = False
+
+    def __post_init__(self):
+        if self.cleanup and not self.terminate:
+            # cleanup implies terminate
+            raise ValueError('cleanup implies terminate.')
```
eirrgang, line 1013: There was previously no enforcement, so I may have misinterpreted the intention. Previously, if a user provided `cleanup=True` and `terminate=False`, this was not an error as far as the `close()` function call was concerned. I assume that was an oversight. If it was intentional, then this commit is wrong.

eirrgang:
If you need to continue to support Python 3.6 for a while, I can resubmit without `dataclasses`, if you like. But please clarify the desired logic.

mtitov:
@eirrgang hi Eric, I like your approach, but if you don't mind, can I propose to use `radical.utils.Munch` class (our typed dictionary class)? as well it will help to resolve issue with python version (as you mentioned, `dataclasses` is not supported by python3.6)
```python
class _CloseOptions(ru.Munch):

    _schema = {
        'cleanup'  : bool,
        'download' : bool,
        'terminate': bool
    }

    _defaults = {
        'cleanup'  : False,
        'download' : False,
        'terminate': True
    }

    def _verify(self):
        if self.cleanup and not self.terminate:
            # cleanup implies terminate
            raise ValueError('cleanup implies terminate.')

# within initialization method
    def __init__(...
        self._close_options = _CloseOptions(close_options)
# within close method
    def close(self, **close_options):
        self._close_options.update(close_options)
        self._close_options.verify()  # in case to call for `_verify` method and to convert attributes 
                                      # to their types if needed (but None value will stay if it is set)
        # available attributes:
        #    self._close_options.cleanup
        #    self._close_options.download
        #    self._close_options.terminate
```
@andre-merzky Andre, what do you think?

andre-merzky:
> @andre-merzky Andre, what do you think?

Ack, thanks.

eirrgang:
> @eirrgang hi Eric, I like your approach, but if you don't mind, can I propose to use `radical.utils.Munch` class

Done.

Is the primary (explicit) `Session.close()` use case tested? It seems like it might be a good idea to make sure that the options are merged as expected (and to encoded in unit-test form what those expectations are). I'm not sure what is supposed to happen with `None` values to Munch.update(), for instance.

src/radical/pilot/session.py
```diff
@@ -254,13 +258,21 @@ def close(self, cleanup=False, terminate=True, download=False):
...
-            # cleanup implies terminate
-            terminate = True
+        # Merge kwargs with current defaults stored in self._close_options
+        self._close_options.update(
+            {
+                'cleanup': cleanup,
+                'download': download,
+                'terminate': terminate
+            }
+        )
```
mtitov, line 268: I think, it is better to use keyword arguments here, thus if certain argument is set then only the corresponding close-option will be updated and others will keep their default values
```python
    def close(self, **kwargs):
        ...
        self._close_options.update(kwargs)
        ...
# s.close(download=True) -> cleanup: False, download: True, terminate: True
```
(*) `close(hello=True)` will raise an error, that `hello` is not in schema

eirrgang: Done.

mtitov (approved):
Eric, thank you! LGTM

src/radical/pilot/session.py
mtitov: resolved (if user will set `s.close(cleanup=None)` then it will be interpreted as `not True`, but only `True` values are targeted)

PR #2405, opened 2021-06-25
Fix/issue 2129

### LINKED ISSUE #2129: Scheduler needs to fail tasks which can never be placed

andre-merzky:
This fixes #2129: if scheduling a task fails, and at the same time the scheduler has all resources available ( no other tasks have resources allocated), then the scheduler will fail the task.

### DIFF
--- a/src/radical/pilot/agent/scheduler/base.py
+++ b/src/radical/pilot/agent/scheduler/base.py
@@ -684,9 +684,10 @@ class AgentSchedulingComponent(rpu.Component):
                                                 log=self._log)
 
         for task, error in failed:
-            task['stderr']  = error
-            task['control'] = 'tmgr_pending'
-            task['$all']    = True
+            task['stderr']       = error
+            task['control']      = 'tmgr_pending'
+            task['target_state'] = 'FAILED'
+            task['$all']         = True
 
             self._log.error('bisect failed on %s: %s', task['uid'], error)
             self.advance(scheduled, rps.FAILED, publish=True, push=False)
@@ -783,9 +784,10 @@ class AgentSchedulingComponent(rpu.Component):
 
             except Exception as e:
 
-                task['stderr']  = str(e)
-                task['control'] = 'tmgr_pending'
-              # task['$all']    = True
+                task['stderr']       = str(e)
+                task['control']      = 'tmgr_pending'
+                task['target_state'] = 'FAILED'
+                task['$all']         = True
 
                 self._log.exception('scheduling failed for %s', task['uid'])
 
--- a/src/radical/pilot/agent/scheduler/base.py
+++ b/src/radical/pilot/agent/scheduler/base.py
@@ -216,8 +216,6 @@ class AgentSchedulingComponent(rpu.Component):
     def __init__(self, cfg, session):
 
         self.nodes = None
-        self._uid  = ru.generate_id(cfg['owner'] + '.scheduling.%(counter)s',
-                                    ru.ID_CUSTOM)
         rpu.Component.__init__(self, cfg, session)
 
 
--- a/src/radical/pilot/agent/scheduler/base.py
+++ b/src/radical/pilot/agent/scheduler/base.py
@@ -261,7 +261,7 @@ class AgentSchedulingComponent(rpu.Component):
         self._waitpool   = dict()  # map uid:task
         self._ts_map     = dict()
         self._ts_valid   = False   # set to False to trigger re-binning
-        self._active_cnt = 0       # count of currently schedled tasks
+        self._active_cnt = 0       # count of currently scheduled tasks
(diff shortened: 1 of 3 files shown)


src/radical/pilot/agent/scheduler/base.py
```diff
@@ -587,6 +589,9 @@ def _schedule_tasks(self):
         self.register_output(rps.AGENT_EXECUTING_PENDING,
                              rpc.AGENT_EXECUTING_QUEUE)
 
+        self._publishers = dict()
+        self.register_publisher(rpc.STATE_PUBSUB)
```
mtitov, line 593: these operations are already applied in `rpu.Component` (in `__init__` and `_initialize` methods)

andre-merzky: Indeed - but the scheduler spawns a new process, and the ZMQ threads are there gone - that is the cause of the problem :-)

mtitov: hmm, not clear... any RP scheduler is based on `rpu.Component`, thus after scheduler instance is created (`__init__`) it will have that attribute `_publishers` (as any other component). All components are created [here](https://github.com/radical-cybertools/radical.pilot/blob/73fc5a8c43db20ea8ac7dece32b13daa95f853b6/src/radical/pilot/utils/component.py#L538) and in case of scheduler it will return a corresponding scheduler (according to what is set in config). Then scheduler is started [here](https://github.com/radical-cybertools/radical.pilot/blob/73fc5a8c43db20ea8ac7dece32b13daa95f853b6/src/radical/pilot/utils/component.py#L458) and method `_initialize` is called, which is inherited from `rpu.Component` and not overloaded in scheduler. Thus scheduler will call `self.register_publisher(rpc.STATE_PUBSUB)`

cannot figure out what I'm missing..

p.s. there probably something wrong with component `uid`: Scheduler sets `uid` [here](https://github.com/radical-cybertools/radical.pilot/blob/73fc5a8c43db20ea8ac7dece32b13daa95f853b6/src/radical/pilot/agent/scheduler/base.py#L219), but then `rpu.Component` overwrite it [here](https://github.com/radical-cybertools/radical.pilot/blob/73fc5a8c43db20ea8ac7dece32b13daa95f853b6/src/radical/pilot/utils/component.py#L400) with value from config

andre-merzky: [this](https://github.com/radical-cybertools/radical.pilot/blob/73fc5a8c43db20ea8ac7dece32b13daa95f853b6/src/radical/pilot/agent/scheduler/base.py#L297) is the line I meant, sorry for being vague: the scheduler spawns it's own sub-process (beyond the component's main process), and that process will need it's own set of ZMQ comm threads.

Ack on the uid though.

mtitov: with the current setup
```
  python               : /home/matitov/.miniconda3/envs/rct/bin/python3
  pythonpath           : 
  version              : 3.7.10
  virtualenv           : rct

  radical.gtod         : 1.5.0
  radical.pilot        : 1.6.6
  radical.saga         : 1.6.6
  radical.utils        : 1.6.6
```
when I added the following lines instead of that above (within method `_schedule_tasks`)
```python
        self._log.debug('>>>: %s', self._publishers)
        self._log.debug('>>>: %s', self._outputs)
```
I've got the following output in logs
```
DEBUG    : >>>: {'state_pubsub': <radical.utils.zmq.pubsub.Publisher object at 0x7f16ccf632d0>, 'control_pubsub': <radical.utils.zmq.pubsub.Publisher object at 0x7f16ccf63490>}
DEBUG    : >>>: {'AGENT_EXECUTING_PENDING': <radical.utils.zmq.queue.Putter object at 0x7f16ccf6a650>}
```
as what I would expect (since when thread is started, all initializations are done, thus instance already has that attribute set and publishers registered)

Thus that two lines above are a duplication

andre-merzky: Yes, the data structure are still there, as the memory of the process is copied during the fork.  But the actual threads are gone - they will not survive the fork.  That the data structures still exist hid the problem, otherwise the component base class would have complained that no listener exists for the channel...

mtitov: yeah, thank you for you patience! understood the cause of this issue

mtitov (approved):
LGTM!

src/radical/pilot/agent/scheduler/base.py
andre-merzky: > yeah, thank you for you patience! understood the cause of this issue

Not at all!  That those points aren't clear in the code is a bit of a problem.  Not to be addressed right now I guess, but eventually we may want to make this more obvious, explicit and consistent...

PR #2415, opened 2021-07-02
Separate cpu/gpu utilization

mturilli:
This PR separate CPU and GPU utilization. The assumption is that we will plot those resource on separate plots. For now, no single plot with all the utilized resources will be available.


andre-merzky (approved):
LGTM, thanks!

You may want to delete the lines you commented out.

PR #2416, opened 2021-07-02
Cleaned list of dependencies

mtitov:
Removed from the list: `netifaces`, `ntplib` (they are dependencies in RU)
Removed from the list in bootstrapper and in script for VE creation: `future` (dependency that was removed from RU)


andre-merzky (approved):
LGTM - thanks!

PR #2421, opened 2021-07-08
Use methods from the new module `host` within RU (>=1.6.7)

mtitov:
- removed `python-hostlist` dependency (from setup, bootstrapper, Slurm RM);
- removed host-related methods from LM base class, which are moved to RU;
- updated Cobalt RM to use host names (and updated corresponding resource config for `anl.theta/theta_gpu`);

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -245,10 +245,9 @@ setup_args = {
     'package_data'       : {'': ['*.txt', '*.sh', '*.json', '*.gz', '*.c',
                                  '*.md', 'VERSION', 'SDIST', sdist_name]},
   # 'setup_requires'     : ['pytest-runner'],
-    'install_requires'   : ['radical.utils>=1.6.6',
+    'install_requires'   : ['radical.utils>=1.6.7',
                             'radical.saga>=1.6.6',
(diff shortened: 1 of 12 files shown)


setup.py
```diff
@@ -245,10 +245,9 @@ def run(self):
     'package_data'       : {'': ['*.txt', '*.sh', '*.json', '*.gz', '*.c',
                                  '*.md', 'VERSION', 'SDIST', sdist_name]},
   # 'setup_requires'     : ['pytest-runner'],
-    'install_requires'   : ['radical.utils>=1.6.6',
+    'install_requires'   : ['radical.utils>=1.6.7',
```
mtitov, line 248: new releases will be of version `1.6.7`, right?

mtitov:
Test with Slurm: `Found SLURM_NODELIST c455-[072-074]. Expanded to: ['c455-072', 'c455-073', 'c455-074']` -> `ru.get_hostlist(hoststring=os.environ['SLURM_NODELIST'])` works as expected

andre-merzky (approved):
LGTM, Thanks!

PR #2424, opened 2021-07-14
Fixed pilot staging for input directories

mtitov:


andre-merzky (approved):
Thanks for adding the test also!

mtitov:
keep it open until related issue is confirmed to be resolved

PR #2431, opened 2021-07-30
Update launch method `aprun`

mtitov:


src/radical/pilot/agent/launch_method/aprun.py
```diff
@@ -1,221 +1,84 @@
...
     # --------------------------------------------------------------------------
     #
-    def __init__(self, name, cfg, session):
-
-        LaunchMethod.__init__(self, name, cfg, session)
+    def _configure(self):
 
+        self.launch_command = ru.which('aprun')
+        # only one instance of `aprun` is allowed per node at any time,
+        # otherwise multiple instances will be executed sequentially
```
andre-merzky, line 22: Do you know if this universally true for aprun, or is that  a system config setting on Theta?

andre-merzky (commented):
The PR seems to change the LM to ignore the scheduler decision (`slots` is not evaluated).  Is that the intent, and why?

andre-merzky (approved):
We decided tyo merge w/o explicit placement as Theta gives trouble.  We'll open an issue for that.

mtitov:
closed in favor #2502

PR #2434, opened 2021-08-09
Change custom `repr()` output.

### LINKED ISSUE #2391: radical.pilot `__repr__()` methods have confusing output.
RP objects can have confusing `repr()` output.

For Task and Pilot, at least, `__repr__()` gives a string representation of a list object, which is unconventional and confusing. It gives the appearance that a Pilot instance *is* a list. Furthermore, the elements represented are not related to the `__init__()` signature in any way.

Since this output duplicates the `__repr__()` output, it actually removes utility. If the `__repr__()` function were removed, it would be more obvious (at least in common debugging scenarios) that an object is in fact a Pilot instance and not an unexpected list.

My request would be either
1. to redefine `__repr__()` to return the conventional output that could conceivably be `eval` ed to instantiate the represented object,
2. format the output with a description between angle brackets (`<...>`), clearly indicating the object is a Pilot instance, or
3. to remove the `Pilot.__repr__()` definition so that the interpreter default is used.

A reasonable alternative could be to return the full dictionary representation. This would seem more appropriate for `__str__()`, but could be made more conventional by formatting in the form of a `PilotManager.submit_pilots()` call.

eirrgang:
Let `repr()` provide more conventional debugging
output for Pilot and Task instances.

Resolves #2391.


andre-merzky:
@mtitov: can you review this please?

mtitov (commented):
Agree with the `repr` format (looks more clear), but can we have a different (aka old) style for string formatting `'<%s object, uid %s>' % (self.__class__.__qualname__, self._uid)`? (to be consistent through all the code)

mtitov (approved):
LGTM! Thanks!

Issue #2436, opened 2021-08-12
Add more documentation linking targets.


eirrgang:
Please consider adding `py` domain Sphinx directives to the RST docs for Python details that are not auto-documented.

For example, the modules do not typically have docstrings, but a `.. py:module:: radical.pilot` directive in `apidoc.rst` or `intro.rst` would allow convenient linking, even across projects (with intersphinx).

I think each RST document could be the target of at least one otherwise undocumented `py` entity.

And several of the modules have natural connections to existing documentation:
```
:py:mod:`radical.pilot.constants`
:py:mod:`radical.pilot.context`
:py:mod:`radical.pilot.staging_directives`
:py:mod:`radical.pilot.states`
:py:mod:`radical.pilot.configs`

PR #2437, opened 2021-08-12
Support :py:mod:`radical.pilot`.

eirrgang:
Add a Sphinx directive to the main RP documentation entry point so that
    :py:mod:`radical.pilot`
can have a target.

Refs #2436

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -3,6 +3,8 @@
 RADICAL-Pilot |version| Documentation
 #####################################
 
+.. py:module:: radical.pilot
+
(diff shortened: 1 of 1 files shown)


docs/source/index.rst
```diff
@@ -3,6 +3,8 @@
 RADICAL-Pilot |version| Documentation
 #####################################
 
+.. py:module:: radical.pilot
+
```
eirrgang, line 7: Of course, some amount of the subsequent text could be indented to become explicitly module documentation, but I don't think there is substantial utility in doing so.

andre-merzky (approved):
LGTM, thanks!

PR #2439, opened 2021-08-20
Hotfix/updated module list on Summit

lee212:
According to the software change, https://docs.olcf.ornl.gov/software/software-news.html,

No longer exists:
- `py-pip` 
- `py-virtualenv`
- `py-setuptools` (is listed but not available)

Updated:
- `gcc/9.1.0` (from 8.1.1)
- `libzmq/4.3.3` (from zeromq/4.2.5, renamed)
- `python/3.7-anaconda3`


andre-merzky (approved):
LGTM, thanks both!

src/radical/pilot/configs/resource_ornl.json
```diff
@@ -70,31 +70,22 @@
...
+                                         "module load   gcc/9.1.0",
+                                         "module load   libzmq/4.3.3",
+                                         "module load   python/3.7-anaconda3",
                                          # increase process limit on node
                                          "ulimit -u 65536"],
         "valid_roots"                 : ["$MEMBERWORK/"],
         "default_remote_workdir"      : "$MEMBERWORK/%(pd.project)s",
-        "rp_version"                  : "local",
-        "virtenv_mode"                : "create",
+        "virtenv_mode"                : "local",
```
lee212, line 87: @mtitov, can this prevent that task executable changes conda envs? I've noticed that python env wasn't changed although each task specifies different conda activation, and if I reverse this `virtenv_mode` and corresponding configs, it seems working correctly.

mtitov: @lee212 this mode tells to RP bootstrapping to use the same env as it is set for "client" side (since it is within the system), thus that shouldn't affect tasks if they use the same env as it is set by bootstrapping, do you change env for each task? can you please show an example?

lee212: For example task.0000.sh can have lines to change python env to `openmm` like:
```
# Pre-exec commands
prof task_pre_start
. /sw/summit/python/3.7/anaconda3/5.3.0/etc/profile.d/conda.sh ||  (echo "pre_exec failed"; false) || exit
conda activate /gpfs/alpine/world-shared/chm155/hrlee/conda/openmm ||  (echo "pre_exec failed"; false) || exit
prof task_pre_stop

# The command to run
prof task_exec_start
/opt/ibm/jsm/bin/jsrun --erf_input /gpfs/alpine/scratch/hrlee/chm155/radical.pilot.sandbox/re.session.login2.hrlee.018869.0011/pilot.0000/task.0000//task.000
0.rs --smpiargs="off"  /gpfs/alpine/world-shared/chm155/hrlee/conda/openmm/bin/python "/gpfs/alpine/world-shared/chm155/hrlee/git/DeepDriveMD-pipeline/deepdr
ivemd/sim/openmm/run_openmm.py" "-c" "/gpfs/alpine/world-shared/chm155/hrlee/runs/bba-test-4th/molecular_dynamics_runs/stage0000/task0000/stage0000_task0000.
yaml"

```

while the client env is not `openmm`. Does it have to be differently loaded to switch python env?

mtitov: correct, it shouldn't be affected, and `...openmm` env should be used by task, maybe I misread, but what is the issue then?

p.s. this line `. /sw/summit/python/3.7/anaconda3/5.3.0/etc/profile.d/conda.sh` should be updated accordingly

lee212: Task executable still loads libraries from client envs not from openmm in this example.

mtitov: Thank you @lee212 ! I was able to reproduce the issue, and let's postpone introducing local-mode here (keep it for other PR)

mtitov (approved):

PR #2442, opened 2021-08-21
Hotfix/flux update

andre-merzky:
This PR fixes several bugs which were introduced by either Flux changes (new event types), or merges (termination problem), and adds a proper conversion from RP tasks to Flux jobs.

### DIFF
--- a/src/radical/pilot/agent/scheduler/flux.py
+++ b/src/radical/pilot/agent/scheduler/flux.py
@@ -1,6 +1,6 @@
 
-__copyright__ = "Copyright 2017, http://radical.rutgers.edu"
-__license__   = "MIT"
+__copyright__ = 'Copyright 2017, http://radical.rutgers.edu'
+__license__   = 'MIT'
 
 
 import json
@@ -70,7 +70,7 @@ class Flux(AgentSchedulingComponent):
           # # FIXME: transfer from executor
           # self._task_environment = self._populate_task_environment()
 
-            jd  = json.dumps(ru.read_json('/home/merzky/projects/flux/spec.json'))
+            jd  = json.dumps(self.task_to_spec(task))
             jid = flux_job.submit(self._flux, jd)
             task['flux_id'] = jid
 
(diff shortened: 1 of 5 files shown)


src/radical/pilot/agent/scheduler/flux.py
```diff
@@ -144,5 +144,49 @@ def work(self, tasks):
   #     return new_env
 
 
+    # --------------------------------------------------------------------------
+    #
+    def task_to_spec(self, task):
```
mtitov, line 149: I would propose to move it into LM (or at least to executor), but if to keep it here is just a temporary solution then ok

src/radical/pilot/agent/scheduler/flux.py
```diff
@@ -144,5 +144,49 @@ def work(self, tasks):
...
+
+        td   = task['description']
+        cmd  = '%s %s 1>%s 2>%s' % (td['executable'],
+                      ' '.join(td['arguments']), task['stdout'], task['stderr'])
```
mtitov, line 153: instead of `task['stdout'], task['stderr']` should be `td['stdout'], td['stderr']`, which also should be checked - e.g., `td.get('stdout') or '%s.out' % tid` (similar to what `Popen` does)

mtitov (commented):
couple more comments related to Flux LM:
- missed `rm_shutdown_hook` in Flux LM, or corresponding method in LM base class should be edited (remove lines L249 and L259, [link](https://github.com/radical-cybertools/radical.pilot/blob/2ac753c5a867ec6e307546fbca5a65ec329824fe/src/radical/pilot/agent/launch_method/base.py#L249))
- as for docker container `FLUX_URI` should not be updated with `ssh` [link](https://github.com/radical-cybertools/radical.pilot/blob/2ac753c5a867ec6e307546fbca5a65ec329824fe/src/radical/pilot/agent/launch_method/flux.py#L72-L75) or this could be configured with resource config - this one comment is just to raise a point

overall PR looks good and works fine

src/radical/pilot/agent/scheduler/flux.py
andre-merzky: Yes, I agree with that certainly.  If you don't mind though, let's clean this up after `env_isolation` merge.  I'll open a ticket to keep track of it.  Thanks!

src/radical/pilot/agent/scheduler/flux.py
andre-merzky: Ack, thanks!

andre-merzky:
Sorry, I had the wrong base branch - hotfix was supposed to go to master...

mtitov (approved):
LGTM!

PR #2448, opened 2021-09-29
Resource config for Amarel@Rutgers

mtitov:
Based on ticket [#2079](https://github.com/radical-cybertools/radical.pilot/issues/2079)

additional material: https://sites.google.com/view/cluster-user-guide/amarel

### DIFF
--- a/src/radical/pilot/configs/resource_rutgers.json
+++ b/src/radical/pilot/configs/resource_rutgers.json
@@ -1,0 +1,36 @@
+{
+    "amarel":
+    {
+        "description"                 : "Heterogeneous community-model Linux cluster",
+        "notes"                       : "Access from registered IP address",
+        "schemas"                     : ["local", "ssh"],
+        "local"                       :
+        {
+            "job_manager_endpoint"    : "slurm://amarel.rutgers.edu/",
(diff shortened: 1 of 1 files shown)


andre-merzky (approved):
LGTM

AymenFJA:
@mtitov, @andre-merzky. Ok, so here are some updates: RP is failing on Amarel for the following reason: ```/usr/bin/srun: unrecognized option '--nodefile=/home/afa64/radical.pilot.sandbox/rp.session.amarel2.amarel.rutgers.edu.afa64.018907.0010/pilot.0000/task.000013//task.000013.nodes'```. 

Now, it is obvious why this is happening, Amarel uses an old version of SLURM (18.0.8) which if you check the archive documentation (https://slurm.schedmd.com/archive/slurm-18.08.8/srun.html) they do not support the flag `--nodefile`. How should we proceed? should we support backward compatibility of the old version of SLURM?

andre-merzky:
Thanks for the feedback, @AymenFJA !  We should indeed switch to `--nodefile` in RP also - SAGA did the switch already, there is no reason to not keep RP in sync.  Would you mind taking an attempt at a PR?

PR #2466, opened 2021-11-09
Add fallback for ssh tunnel on ifconfig-less nodes

vrpascuzzi:
Currently, `ifconfig` is the primary executable for determining
the bind address for creating an SSH tunnel for nodes without an
outside-world connection. However, some machine nodes are not
equipped with `ifconfig` -- e.g, anl.arcticus -- and the last-case
scenario using `ip` returns a string of IP addresses instead of a
single one.

This commit employs the use of `host` to get the node IP address
for SSH tunnel binding (thanks to @andre-merzky for the tip!) when
`ifconfig` is unavailable in the 'usual' location.


andre-merzky (approved):
Great, thanks for the PR - LGTM, feel free to push the merge button!

PR #2477, opened 2021-12-01
Hotfix/env isolation

andre-merzky:
Some changes for the env isolation release:
  - fix linebreaks in LMs / popen
  - fix env setup and dumping in bootstrapper (use stages RU shell tools)
  - limit mongoDB version due to conflicts with recent updates


src/radical/pilot/agent/agent_0.py
```diff
@@ -784,14 +784,12 @@ def _check_tasks_cb(self):
...
 
-        print(env_spec)
-
         etype = env_spec['type']
         evers = env_spec['version']
         emods = env_spec.get('setup')    or []
         pre   = env_spec.get('pre_exec') or []
 
-        pre_exec = '-P ". env/bs0_orig.sh" '
+        pre_exec = '-P ". env/bs0_pre_0.sh" '
```
mtitov, line 792: not clear with this, which script is assumed to be changed `bs0_orig.sh` or `bs0_pre_0.sh`? (I understand that at this point they are the same)

mtitov (commented):
couple comments, but other changes LGTM

src/radical/pilot/agent/agent_0.py
andre-merzky: I'm positive on this one - changes need to apply to the pre_bootstrap_0 also (which results in the `bs0_pre_0` env).  Thanks for checking!

Issue #2451, opened 2021-10-12
RP fails on Amarel due to SLURM old version (--nodefile not recognized)


AymenFJA:
Based on the PR #2448, RP is failing on Amarel for the following reason:

 ```bash
/usr/bin/srun: unrecognized option '--nodefile=/home/afa64/radical.pilot.sandbox/rp.session.amarel2.amarel.rutgers.edu.afa64.018907.0010/pilot.0000/task.000013//task.000013.nodes'
```

Now, it is obvious why this is happening, Amarel uses an old version of SLURM (18.0.8) which if you check the archive documentation (https://slurm.schedmd.com/archive/slurm-18.08.8/srun.html) they do not support the flag `--nodefile`. How should we proceed? should we support backward compatibility of the old version of SLURM?

AymenFJA:
@andre-merzky medium!

AymenFJA:
Hello @andre-merzky . Any news regarding this issue? Thanks.

AymenFJA:
This is related to #2480 and it is solved now! I will reopen it if something comes up.

PR #2480, opened 2021-12-12
detect slurm version and use node file/list

AymenFJA:
This PR is regarding `srun` launch method and this issue #2451 , the old slurm (<=18) does not support `--nodefile` which what `srun` is using by default this PR will help to detect the major version of slurm and decide to use `nodefile` or `nodelist`.

### DIFF
--- a/src/radical/pilot/agent/launch_method/srun.py
+++ b/src/radical/pilot/agent/launch_method/srun.py
@@ -93,7 +93,7 @@ class Srun(LaunchMethod):
     def get_slurm_ver(self):
 
         version = self._version.strip('\n')
-        major_version = eval(version.strip('\n slurm').strip(".")[:2])
+        major_version = int(version.strip('\n slurm').strip(".")[:2])
         return major_version
 
 
--- a/src/radical/pilot/agent/launch_method/srun.py
+++ b/src/radical/pilot/agent/launch_method/srun.py
@@ -88,6 +88,15 @@ class Srun(LaunchMethod):
         return True, ''
 
 
+    # -------------------------------------------------------------------------
+    #
+    def get_slurm_ver(self):
+        
+        version = eval(self._version.strip('\n'))
+        major_version = version.strip('\n slurm').strip(".")[:2]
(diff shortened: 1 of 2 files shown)


src/radical/pilot/agent/launch_method/srun.py
```diff
@@ -88,6 +88,15 @@ def can_launch(self, task):
         return True, ''
 
 
+    # -------------------------------------------------------------------------
+    #
+    def get_slurm_ver(self):
+        
+        version = eval(self._version.strip('\n'))
```
andre-merzky, line 95: I don't understand what the purpose of the `eval` is here - can you explain please?

andre-merzky (commented):
Thanks for taking care of this!

src/radical/pilot/agent/launch_method/srun.py
AymenFJA: @andre-merzky I committed a change. The eval should be on the major version not here! The eval should evaluate the `str` to `int` to be used in the version check later. I replaced it with `int` as it is safer!

src/radical/pilot/agent/launch_method/srun.py
```diff
@@ -93,7 +93,7 @@ def can_launch(self, task):
     def get_slurm_ver(self):
 
         version = self._version.strip('\n')
-        major_version = eval(version.strip('\n slurm').strip(".")[:2])
+        major_version = int(version.strip('\n slurm').strip(".")[:2])
```
andre-merzky, line 96: This will break in 43 years when slurm reaches version 100.0.0 :-P  Maybe
```
major_version = self._version.split('.')[0].split()[-1]
```
(take element before first dot, pick last word (strip `slurm ` and whatever they decide to print up front)?

AymenFJA: Even better! thanks for the suggestion. I have committed the change accordingly.

andre-merzky (approved):
LGTM - thanks!

PR #2484, opened 2022-01-21
set RP_RANK on non-MPI tasks

andre-merzky:
Non-MPI tasks miss the `RP_RANK` env setting and fall through the rank switch in `task_uid.exec.sh`.

See also https://github.com/radical-cybertools/radical.entk/issues/599


mtitov (approved):
LGTM!

PR #2487, opened 2022-01-24
fixing srun version command

AymenFJA:
This PR is to fix the `srun` `slurm` version here: https://github.com/radical-cybertools/radical.pilot/blob/2c0f72511f50bfe966ca36041473a8c05e824489/src/radical/pilot/agent/launch_method/srun.py#L93. 

`self._version` is not defined in the initialization of the class so the function `_get_slurm_ver` does not have that value. I propose this solution in the PR (but it introduces a code duplication) as the same chunk of code are in the `_init_from_scratch`. 

There is another option, which defines the version in the initializer of the class so all functions can see the value of the `Srun` version, but it might fail if `Srun` doesn't exist:

```python

    #
    def __init__(self, name, lm_cfg, rm_info, log, prof):

        self.command = ru.which('srun')

        LaunchMethod.__init__(self, name, lm_cfg, rm_info, log, prof)

        out, err, ret = ru.sh_callout('%s -V' % command)
        if ret:
            raise RuntimeError('cannot use srun [%s] [%s]' % (out, err))

        self.version = out.split()[-1]
```
Any suggestions please?

### DIFF
--- a/src/radical/pilot/agent/launch_method/srun.py
+++ b/src/radical/pilot/agent/launch_method/srun.py
@@ -130,8 +130,7 @@ class Srun(LaunchMethod):
             # mapping (see above), but we at least honor the nodelist.
             nodelist = [rank['node_name'] for rank in slots['ranks']]
 
-            # newer slurm versions need a nodefile
-            # FIXME: is this really true?
+            # older slurm versions don't accept nodefiles
             if self._vmajor > 18:
                 nodefile = '%s/%s.nodes' % (sbox, uid)
                 with ru.ru_open(nodefile, 'w') as fout:
--- a/src/radical/pilot/agent/launch_method/srun.py
+++ b/src/radical/pilot/agent/launch_method/srun.py
@@ -44,18 +44,19 @@ class Srun(LaunchMethod):
     def _init_from_scratch(self, env, env_sh):
 
         command = ru.which('srun')
-
         out, err, ret = ru.sh_callout('%s -V' % command)
         if ret:
             raise RuntimeError('cannot use srun [%s] [%s]' % (out, err))
 
-        self._version = out.split()[-1]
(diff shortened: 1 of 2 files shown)


andre-merzky:
@AymenFJA : thanks for opening the PR.  It turned out that the version check was crooked in multiple ways.  I pushed some commits to hopefully correct that.

Out of scope for this PR, but I am still uncertain about the underlying assumption though: we switched to 'nodelist' for older versions of slurm (because of Amarel) - but shouldn't we then use `nodelist` for all slurm versions?  That would reduce the load on the file system.  We should only switch to nodefiles once the command line with the node list becomes too long.

Anyway, please do a review and test with these changes and let us know how it goes!

AymenFJA:
> @AymenFJA : thanks for opening the PR. It turned out that the version check was crooked in multiple ways. I pushed some commits to hopefully correct that.
> 
> Out of scope for this PR, but I am still uncertain about the underlying assumption though: we switched to 'nodelist' for older versions of slurm (because of Amarel) - but shouldn't we then use `nodelist` for all slurm versions? That would reduce the load on the file system. We should only switch to nodefiles once the command line with the node list becomes too long.
> 
> Anyway, please do a review and test with these changes and let us know how it goes!

@andre-merzky  thank you. My assumption was based on what I have experienced, and it is as follows: I was requesting only 1 node (which in my understanding it shouldn't require the setup of `--nodfile` to launch a task on 1 node/4 cores).  However, RP used `--nodefile` instead of `nodelist`, and from here the problem started, as slurm version was pretty old at `Amarel` and complained about `--nodefile` option. 

I totally agree that we should use `--nodelist` if the number of characters is not exceeding the limit (I don't know the max number), otherwise we switch to `--nodefile`.

Happy to work on it if necessary.

Finally, the PR/latest commits fixed the problem for now.

PR #2501, opened 2022-02-07
fix metric setup / lookup in tmgr

### LINKED ISSUE #2444: Logic error in TaskManager.unregister_callback for metrics=None
At
https://github.com/radical-cybertools/radical.pilot/blob/68f455bae19fa6479a25f1113093d3b49938384b/src/radical/pilot/task_manager.py#L1201
the default value for `metrics` becomes `[['TASK_STATE', 'WAIT_QUEUE_SIZE']]`.

I believe that `['TASK_STATE', 'WAIT_QUEUE_SIZE']` is what is intended. E.g.

    if not metrics: metrics = rpc.TMGR_METRICS

To reduce future errors, maybe remove the `else` from the following line, as well.

andre-merzky:
This fixes #2444

### DIFF
--- a/src/radical/pilot/task_manager.py
+++ b/src/radical/pilot/task_manager.py
@@ -1195,8 +1195,8 @@ class TaskManager(rpu.Component):
 
             for metric in metrics:
 
-                if  metric not in rpc.TMGR_METRICS:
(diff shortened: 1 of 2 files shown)


mtitov (approved):
LGTM

PR #2502, opened 2022-02-07
Update launch method `aprun`

mtitov:


mtitov:
> This version now ignores the scheduler defined slots.

right, Theta doesn't like it (it is possible to use the current implementation of `aprun`, but only without option `-L`)

> Should we add a code path where slots are used again?

yeah, I keep the corresponding code commented out and did extend it with possibility to share nodes (option `-F`)
https://github.com/radical-cybertools/radical.pilot/blob/726762222fdee76daa98d07286a64ef670c44ad4/src/radical/pilot/agent/launch_method/aprun.py#L115-L119

andre-merzky (approved):
Fair - since we don't have other `aprun` machines at the moment, that makes sense.  We can revisit in case other machines appear...

andre-merzky:
@mtitov : can this be merged?

mtitov:
Andre, it is, thank you for remind!

PR #2507, opened 2022-02-12
Adding MPI as optional dependency in installation.rst Fixes #2506

### LINKED ISSUE #2506: Documenting MPI Library  as optional dependency
The current installation doesn't list MPI as optional dependency. 
https://radicalpilot.readthedocs.io/en/stable/installation.html#software

We need to document this as a dependency to easier "Up and Running Process"

kartikmodi:


docs/source/installation.rst
```diff
@@ -111,6 +111,14 @@ If you don't have access to a suitable machine with a public IP, you can open
 a ticket with RADICAL lab and we will provide a testing MongoDB instance for
 you.
 
+MPI Implementation Library (Optional)
+-------
+
+The MPI is used in launch method. There are several MPI flavours and implementations available
```
andre-merzky, line 117: For
```
The MPI is used in launch method
```
I would suggest
```
MPI is used as a task launch method
```

andre-merzky (approved):
Left one comment, otherwise this looks good to me - thanks!

mtitov:
@andre-merzky have a general comment - should we use a dedicated branch (`docs/nov21`) for any changes in documentation? (even if some section are missing there, just fill up with such new descriptions)

AymenFJA:
> @andre-merzky have a general comment - should we use a dedicated branch (`docs/nov21`) for any changes in documentation? (even if some section are missing there, just fill up with such new descriptions)

agreed!

andre-merzky:
@AymenFJA @mtitov : yes, we can do that.  Small PRs still make sense, but I guess they should be directed toward that documentation branch then...

kartikmodi:
Have to redo the work, because of documentation being in a separate branch.

PR #2514, opened 2022-02-15
Remove task log and the state_history as they are no longer supported Fixes #2508

### LINKED ISSUE #2508: Remove task log and the state_history from documentation
Quoting from https://radicalpilot.readthedocs.io/en/devel/user_guide/02_failing_tasks.html#handle-failing-tasks
`...array of timed log messages, similar to the state_history array discussed before.`

But, the state_history is discussed before. The link on `before` points to the previous document ->
https://radicalpilot.readthedocs.io/en/devel/user_guide/01_task_details.html#chapter-user-guide-01
which doesn't mention state_history anywhere. 

Are we talking of task doc from  _get_session_docs_  which has a state_history key? 
Is state_history recognized, such an important attribute that we need to describe it here on the 3rd page of documentation?

kartikmodi:


andre-merzky (approved):
LGTM - thanks!

PR #2515, opened 2022-02-15
fix capture of exit code

### LINKED ISSUE #2509: Something is broken is task state, task.exit_code
I observe task state and exit_code are reporting wrong values.

**Console Output -** 

**task's state with debugger -** 

**To confirm, I checked the script by running from the sandbox -** 
pasted image 0

I don't know if it's something specific to my env / setup but it's re-creatable on my system. 
Can anyone check on their env as well?

**Env Tested on-**
Branch - devel
OS - Debian 11
local.Locahost - Single machine

andre-merzky:
This fixes #2509
A backslash was missing :-(   Also let the exit code fall through.


mtitov (approved):
LGTM!

PR #2517, opened 2022-02-16
RM parameter `_launch_order`

mtitov:


src/radical/pilot/configs/resource_anl.json
```diff
@@ -16,6 +16,7 @@
         "agent_scheduler"             : "CONTINUOUS",
         "agent_spawner"               : "POPEN",
         "launch_methods"              : {
+                                         "order": ["APRUN"],
```
andre-merzky, line 19: What happens if that is not specified - should't the launch order fall back to `['APRUN']` anyway?

andre-merzky (approved):
LGTM, thanks.

src/radical/pilot/configs/resource_anl.json
andre-merzky: Thanks, makes sense.

PR #2525, opened 2022-02-22
Tests fixes

mtitov:


src/radical/pilot/pmgr/launching/default.py
```diff
@@ -986,7 +986,7 @@ def _prepare_pilot(self, resource, rcfg, pilot, expand, tar_name):
 
         # Convert dict to json file
         self._log.debug("Write agent cfg to '%s'.", cfg_tmp_file)
-        ru.write_json(agent_cfg, cfg_tmp_file)
+        agent_cfg.write(cfg_tmp_file)
```
andre-merzky, line 989: I am ok with the change - but the former version should still work, right?

andre-merzky (approved):
LGTM, thanks!

src/radical/pilot/pmgr/launching/default.py
mtitov: I doesn't.. when we call `json.dump` it cannot serialize `Config` into JSON, but method `write` does convert data into `dict` and it is serializable

andre-merzky: Oh, but that did work before since the Config identified as a dict (IIRC).  Hmm, that's a bit inconvenient - we serialize dict-like thingies all over the place (ZMQ)...

mtitov: `Munch` was inherited from `dict`, and this inheritance is removed for `TypedDict`

andre-merzky: Why did we remove it?

mtitov: The idea was to have a base class not being inherited
Could this PR [RU#335](https://github.com/radical-cybertools/radical.utils/pull/335) be considered?

andre-merzky: Yes, that PR would likely fix it - but why did we remove the inheritance? :-)  I looked back, and we inherited from `DictMixin` in fact - which would also reduce the implementation size (right now we basically re-implement `DictMixin`). 

Sorry for opening a discussion on a PR which was already merged - but wouldn't it be simpler code if we'd revert to the `DictMixin` inheritance?

mtitov: just for a record the summary is:
- `TypedDict` is not inherited from `DictMixin` or `dict` for now, but keep possibility to inherit it from `dict` in case of issues caused by serialization process;
- `DictMixin` is kept for now, but with high probability to be obsolete later (and `TypedDict` might be used instead, for `Registry` implementation)

Issue #2488, opened 2022-01-26
srun should use nodelist instead of nodefile


andre-merzky:
up to a certain number of nodes

AymenFJA:
Few things to highlights and consider:

1. `srun` should use by default `--nodelist` and not `--nodefile` for 2 reasons:

    -  SLURM versions <=18 do not support the option `--nodfile`. (Example SLURM on Amarel)
    -  Using `--nodefile` will leads to creating a file with node names in it, leading to I/O operation that can be expensive, if not needed.

2. Switching to `--nodefile` should happen in one scenario: requesting a certain number of nodes (beyond x numbers) that leads to exceeding the allowed number of characters in a single shell file? 

3. A corner case that I have no clue of how to proceed with if happened, we have a SLURM version that does not support `--nodefile` and the user requested a large number of nodes that exceed the number of allowed characters. This will leads to a certain failure based on what we mentioned earlier in 1 and 2.

AymenFJA:
Update regarding the ticket from yesterday's development meeting: 

1.  The separator between the `--nodelist` and `--nodefile` is the number of nodes. Assuming that SLURM supports the `--nodefile`.
2. If the number of nodes is >42 nodes (that is what we agreed on based on Andre's advice) then we switch to `--nodefile`.
3. 
    ```
    If SLURM_version <=18 & nodes >42:
         set task to fail.
         exit
    ```
Based on these comments, we will add the mentioned cases to the `srun` launcher code.

AymenFJA:
fixed and PR merged.

PR #2536, opened 2022-03-04
Fix nodefile/nodelist

AymenFJA:
This fix is related to #2488.

### DIFF
--- a/src/radical/pilot/agent/launch_method/srun.py
+++ b/src/radical/pilot/agent/launch_method/srun.py
@@ -129,15 +129,24 @@ class Srun(LaunchMethod):
             # the scheduler did place tasks - we can't honor the core and gpu
             # mapping (see above), but we at least honor the nodelist.
             nodelist = [rank['node_name'] for rank in slots['ranks']]
+            n_nodes  = len(set(nodelist))
 
             # older slurm versions don't accept nodefiles
+            # 42 node is the upper limit to switch from `--nodelist`
+            # to `--nodefile`
             if self._vmajor > 18:
-                nodefile = '%s/%s.nodes' % (sbox, uid)
-                with ru.ru_open(nodefile, 'w') as fout:
-                    fout.write(','.join(nodelist))
-                    fout.write('\n')
+                if n_nodes > 42:
+                    nodefile = '%s/%s.nodes' % (sbox, uid)
+                    with ru.ru_open(nodefile, 'w') as fout:
+                        fout.write(','.join(nodelist))
+                        fout.write('\n')
+                    nodelist.clear()
 
-            n_nodes = len(set(nodelist))
+            # corener case that we can not address, related ticket:
+            # https://github.com/radical-cybertools/radical.pilot/issues/2488
+            elif self._vmajor <= 18 and n_nodes > 42:
+                raise Exception('srun <= 18 (nodes: %d) do not support nodefile' % (n_nodes)) 
+
 
         # use `--exclusive` to ensure all tasks get individual resources.
         # do not use core binding: it triggers warnings on some installations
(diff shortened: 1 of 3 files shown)


src/radical/pilot/agent/launch_method/srun.py
```diff
@@ -152,12 +161,11 @@ def get_launch_cmds(self, task, exec_path):
...
-        if self._vmajor > 18:
-            if nodefile:
-                mapping += ' --nodefile=%s' % nodefile
-        else:
-            if nodelist:
-                mapping += ' --nodelist=%s' % ','.join(str(n) for n in nodelist)
+        if nodefile:
+            mapping += ' --nodefile=%s' % nodefile
+
+        if nodelist:
```
andre-merzky, line 167: `elif`?  We should never have both.  In fact, I would say it is 
```
if nodefile:
    ...
else:
    ...
```

src/radical/pilot/agent/launch_method/srun.py
```diff
@@ -129,15 +129,24 @@ def get_launch_cmds(self, task, exec_path):
...
+                if n_nodes > 42:
+                    nodefile = '%s/%s.nodes' % (sbox, uid)
+                    with ru.ru_open(nodefile, 'w') as fout:
+                        fout.write(','.join(nodelist))
+                        fout.write('\n')
+                    nodelist.clear()
+
+            # corener case that we can not address, related ticket:
+            # https://github.com/radical-cybertools/radical.pilot/issues/2488
+            elif self._vmajor <= 18 and n_nodes > 42:
```
andre-merzky, line 147: The check for `self._vmajor <= 18` seems redundant - in this else it's always true.  But also, we don't need to raise - we can the LM try with that long nodelist as command line arg anyway: either it works, hurray, or it fails, then we are not worse off than raising now...

andre-merzky (commented):
Left 2 comments - thanks!

src/radical/pilot/agent/launch_method/srun.py
```diff
@@ -129,15 +129,24 @@ def get_launch_cmds(self, task, exec_path):
...
-                nodefile = '%s/%s.nodes' % (sbox, uid)
-                with ru.ru_open(nodefile, 'w') as fout:
-                    fout.write(','.join(nodelist))
-                    fout.write('\n')
+                if n_nodes > 42:
+                    nodefile = '%s/%s.nodes' % (sbox, uid)
+                    with ru.ru_open(nodefile, 'w') as fout:
+                        fout.write(','.join(nodelist))
+                        fout.write('\n')
+                    nodelist.clear()
```
mtitov, line 143: this operation is unnecessary, latter you just check
```python
if nodefile:
    ...
elif nodelist:
    ...
```

src/radical/pilot/agent/launch_method/srun.py
```diff
@@ -129,15 +129,24 @@ def get_launch_cmds(self, task, exec_path):
...
 
             # older slurm versions don't accept nodefiles
+            # 42 node is the upper limit to switch from `--nodelist`
+            # to `--nodefile`
             if self._vmajor > 18:
-                nodefile = '%s/%s.nodes' % (sbox, uid)
-                with ru.ru_open(nodefile, 'w') as fout:
-                    fout.write(','.join(nodelist))
-                    fout.write('\n')
+                if n_nodes > 42:
```
mtitov, line 138:
```python
if node > 42:

    if self._vmajor <= 18:
        raise Exception(...)

    nodefile = ...
```

btw I also would propose to move these numbers up, as constants (names for constants just an example, you can come up with yours)

```python
MIN_NUM_NODES_IN_NODELIST = 42
MIN_SLURM_VER_TO_NODELIST = 18
```

mtitov (commented):
couple general comments

src/radical/pilot/agent/launch_method/srun.py
AymenFJA: We would never have both @andre-merzky , as you can see if we have `nodfile` then once we finish writing to that `nodefile` from the `nodelist` we clear the `nodelist`. 

```python
with ru.ru_open(nodefile, 'w') as four:
        fout.write(','.join(nodelist))
        fout.write('\n')
nodelist.clear()

src/radical/pilot/agent/launch_method/srun.py
AymenFJA: @mtitov  This looks good to set the min version of `slurm` and `number of nodes`

AymenFJA:
@andre-merzky @mtitov thanks for your comments, I addressed them in the last commit, if this version is fine, then I can move forward with fixing the tests.

andre-merzky (approved):
LGTM, thanks!

mtitov (approved):
LGTM!

andre-merzky:
A test got stuck in `pip install`, thus merged manually...

PR #2542, opened 2022-03-07
removing module reset

AymenFJA:
removing `module reset` from `bridges2`  resource file as it raises warning if no module exists leading to RP bootstrapping failure.


andre-merzky (approved):
LGTM, thanks!

PR #2544, opened 2022-03-09
Feature/pyfunc

AymenFJA:
This PR propose: 

1.  An implementation of an object sterilizer (callable and non-callable).
2.  Python function wrapper class and decorator specifically for python functions with impeded serialization approach. 
3.  Adding the functionality to `raptor/worker_mpi` via `dispatch_pyfucntion` to deserialize and executing python functions that is:

- Shipped from the client-side to the agent side.
- Is not defined in the worker namespace


src/radical/pilot/pytask.py
```diff
@@ -0,0 +1,77 @@
...
+
+import functools
+
+from .  import utils     as rpu
+
+TASK = dict 
+_SER = rpu.Serializer()
+
+
+class PythonTask(object):
```
andre-merzky, line 13: Why do we actually have a class `PythonTask` if that is only holding two methods which are not even class methods (don't take `self` as argument)?

src/radical/pilot/raptor/__init__.py
```diff
@@ -3,4 +3,3 @@
 from .worker         import Worker
 from .worker_mpi     import MPIWorker
 from .worker_default import DefaultWorker
-
```
andre-merzky, line 6: Our coding guidelines actually want an empty line at end of file...

src/radical/pilot/utils/serializer.py
```diff
@@ -0,0 +1,159 @@
...
+            try:
+                with open(self._obj_file_path, 'wb') as f:
+                    dill.dump(obj, f, recurse = True)
+                    result = self._obj_file_path
+            except Exception as e:
+                exception = e
+                self._log.exception("failed to serialize object to file: %s", e)
+
+        if result is None:
+            raise Exception("object is not serializable: %s", exception)
```
andre-merzky, line 103: This should be:
```py
raise Exception("object is not serializable") from exception
```
to preserve the original exception stack for debugging.

src/radical/pilot/utils/serializer.py
```diff
@@ -0,0 +1,159 @@
...
+
+        return self._obj_file_path
+
+
+
+    def deserialize_file(self, file_obj):
```
andre-merzky, line 109: `file_obj` is not an object but a file name, so maybe `fname` which we often use for file name parameters?

src/radical/pilot/utils/serializer.py
```diff
@@ -0,0 +1,159 @@
...
+
+    """
+    The serializer should be able to (de)serialize information that we want
+    to send over the wire from the client side to the agent side via
+    1- ZMQ
+    2- MongoDB
+    """
+
+
+    def __init__(self):
```
andre-merzky, line 24: Why is this actually an object?  The only real data it holds is the file name, and that is only used in a single function.  Couldn't this all be free functions?

andre-merzky (commented):
Overall this looks good to me - thanks for the work, @AymenFJA !

I still left several comments, but those are mostly about smaller things, really.  I'm optimistic we get that merged in a couple of days...

src/radical/pilot/pytask.py
AymenFJA: This class is for wrapping purposes, it acts as a function wrapper without the need to decorate. Example: 

```python
def AA(func:Callable, x= random.random()):
    eq = func(x)
    print (eq * 4)


def AB(z):
    import random
    return random.random() * z


def AC(y):
    import time
    return time.time()


wrapped_function = partial(AA, AB)

PythonTask(wrapped_function)
```

Also, we might need it to wrap any type of function with or without `args` and `kwargs` internally: for example `RP-Parsl` executor.

src/radical/pilot/raptor/__init__.py
AymenFJA: I used our `.pylint` and `.flake8` files to check and they did not complain.

src/radical/pilot/task_description.py
```diff
@@ -26,6 +27,9 @@
 ARGS             = 'args'
 KWARGS           = 'kwargs'
 
+# mode: TASK_PY_FUNCTION
```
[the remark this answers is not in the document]
AymenFJA: It would make sense, but your suggestion was to make a totally different type of task and I ended up creating `PY_FUCNTION`, so I do not know how to proceed now.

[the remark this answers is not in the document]
andre-merzky: Ha, that puts you in a pickle, and me too! :-)  Let me look over the code again...

src/radical/pilot/raptor/__init__.py
AymenFJA: fixed.

src/radical/pilot/raptor/worker_mpi.py
```diff
@@ -1,17 +1,18 @@
-
```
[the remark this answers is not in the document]
AymenFJA: fixed

src/radical/pilot/utils/serializer.py
```diff
@@ -0,0 +1,159 @@
...
+
+        if not os.path.isfile(file_obj):
+            self._log.error("%s is not a file", file_obj)
+            return None
+
+        try:
+            with open(file_obj, 'rb') as f:
+                result = dill.load(f)
+                self._log.debug(result)
+                if not result:
```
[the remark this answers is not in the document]
AymenFJA: fixed

src/radical/pilot/utils/serializer.py
AymenFJA: fixed

src/radical/pilot/utils/serializer.py
AymenFJA: fixed

src/radical/pilot/utils/serializer.py
AymenFJA: fixed with refactoring the code.

src/radical/pilot/raptor/worker_mpi.py
```diff
@@ -547,25 +549,43 @@ def _dispatch_function(self, task, env):
...
+        to_call = ''
+        args    = task['description'].get('args',   [])
+        kwargs  = task['description'].get('kwargs', {})
+
+        # check if we have a serialized object
+        try:
+            func_info  = deserialize_bson(func)
+            to_call    = deserialize_obj(func_info['func'])
+            args       = func_info["args"]
+            kwargs     = func_info["kwargs"]
```
mtitov, line 564: why do we need to have two functions here, could we have one?
```python
func_info = deserialize_func(func)
# func_info could be either tuple `(to_call, args, kwargs)` or dict
```
and I would think the same when we do serialization
```python
return serialize_func(f, args, kwargs)
```

mtitov (commented):
have just general comments

src/radical/pilot/utils/serializer.py
```diff
@@ -0,0 +1,154 @@
...
+    result = None
+
+    try:
+        result = dill.loads(obj)
+        if result is None:
+            raise RuntimeError('failed to deserialize')
+        return result
+
+    except Exception as e:
+        raise Exception ("failed to deserialize from object") from e
```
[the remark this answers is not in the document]
andre-merzky: You don't need the `else` really as the `raise` will leave the scope?

src/radical/pilot/utils/serializer.py
```diff
@@ -0,0 +1,154 @@
...
+
+    try:
+        with open(fname, 'rb') as f:
+            result = dill.load(f)
+            if result is None:
+                raise RuntimeError('failed to deserialize')
+            return result
+
+    except Exception as e:
+        raise Exception ("failed to deserialize object from file") from e
```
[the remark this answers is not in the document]
andre-merzky: same

src/radical/pilot/utils/serializer.py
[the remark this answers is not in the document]
mtitov: but that's not right being inside `try..except` and raise your exception, kinda that's the point with `try` to check that operation will not raise an exception and then in `else` to process received result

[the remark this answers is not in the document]
andre-merzky: It's likely I am missing something as I am a bit tired, but I would think that
```py
    try:
        result = dill.loads(obj)
    except Exception as e:
        raise Exception('failed to deserialize from object') from e
    else:
        if result is None:
            raise RuntimeError('failed to deserialize')
        return result
```
is the same as
```py
    try:
        result = dill.loads(obj)
    except Exception as e:
        raise Exception('failed to deserialize from object') from e

    if result is None:
        raise RuntimeError('failed to deserialize')
    return result
```
isn't it?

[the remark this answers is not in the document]
mtitov: ah, my bad, sorry (misread...), yes!

p.s. `else` would be needed if we don't raise an exception, but want to process result only if exception was not raised

src/radical/pilot/raptor/worker_mpi.py
AymenFJA: @mtitov  They both do `serialize/deserialize` but one returns `string` (`deserialize_bson` which MongoDB accepts)  and one returns `byte` so `deserialize_obj` can be used to deserialize different types of objects which might be needed for use cases like `Colmena`. Combining both is not possible at the moment.

src/radical/pilot/utils/serializer.py
```diff
@@ -0,0 +1,154 @@
...
+
+    if callable(obj):
+        try:
+            with open(_obj_file_path, 'wb') as f:
+                dill.dump(obj, f)
+                result = _obj_file_path
+
+        except Exception as e:
+            exception = e
+            pass
```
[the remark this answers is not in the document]
AymenFJA: @mtitov  You are correct about the `pass`, but for the exception, you must keep it somewhere to raise from it in the end.
Also, we do not want the serializer to stop on one way of serialization (in case it fails). We continue to try to serialize with different approaches until all of them fail then we raise from the last try. Hope this was clear.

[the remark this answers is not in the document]
mtitov: > for the exception, you must keep it somewhere to raise from it in the end

yeah, it will stay in `e`

> we do not want the serializer to stop on one way of serialization (in case it fails)

right, I didn't mean that it will be called before other checks
```python
...
        try:
            with open(_obj_file_path, 'wb') as f:
                dill.dump(obj, f)
        except Exception as e:
            pass
        else:
            result = _obj_file_path
...
    if result is None:
        raise Exception('object is not serializable') from e
```

src/radical/pilot/utils/serializer.py
```diff
@@ -0,0 +1,154 @@
...
+
+    if callable(obj):
+        try:
+            result = dill.dumps(obj)
+        except Exception as e:
+            exception = e
+            pass
+
+        # if we fail, then pikle it by reference
+        if result is None:
```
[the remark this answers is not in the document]
mtitov: > we do not want the serializer to stop on one way of serialization (in case it fails). We continue to try to serialize with different approaches until all of them fail

@AymenFJA in case your comment was related to this part, here is what I meant:
```python
    if callable(obj):
        try:
            result = dill.dumps(obj)
        except Exception:
            try:
                result = dill.dumps(obj, byref = True)
            except Exception as e:
                pass
```

src/radical/pilot/raptor/worker_mpi.py
mtitov: So, I figured out what I didn't like (and my comment below could be skipped) - class `PythonTask` (aka FuncSer) is responsible for function serialization, THUS it should be responsible for deserialization (since object was packed here, this class **knows** its structure)

A corresponding method could be static
```python
class PythonTask:
...
    @staticmethod
    def deserialize(f):
        func_info = deserialize_bson(f)
        to_call   = deserialize_obj(func_info['func'])
        return to_call, func_info['args'], func_info['kwargs']
```
and use it as
```python
            to_call, args, kwargs = PythonTask.deserialize(func)
```

src/radical/pilot/utils/serializer.py
```diff
@@ -0,0 +1,148 @@
...
+    result = codecs.encode(pickle.dumps(obj), "base64").decode()
+
+    return result
+
+
+def deserialize_bson(obj):
+
+    result = pickle.loads(codecs.decode(obj.encode(), "base64"))
+
+    return result
```
mtitov, line 148: one more thing (sorry for many comments...) - why `pickle` is used here? and not `dill`? (kinda, it is good to have one approach, and as I understand `dill` is the same as `pickle`, but provides some extra functionality, so we can stick just to one of them)

src/radical/pilot/raptor/worker_mpi.py
AymenFJA: @mtitov  Why do we want to expose the mechanism of wrapping the python functions to the external API? The purpose of this class `Python Task` is to act as a wrapper, not a wrapper/deserializer. Serializing an object might fall in the category of wrapping but the opposite is not valid IMHOP.

src/radical/pilot/utils/serializer.py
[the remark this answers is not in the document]
AymenFJA: Fixed with the latest commit.

src/radical/pilot/raptor/worker_mpi.py
mtitov: > Why do we want to expose the mechanism of wrapping the python functions to the external API?

It is exposed now when you deserialize func-object outside of `PythonTask`. Kinda when you received an object, which you know was serialized by `PythonTask`, you shouldn't deserialize it by yourself.

A simple example: you have several places where you do deserialization of func-object and serialization process got extended with an extra `dumps` call, then you need to fix deserialization at every place it is done. Thus you have to carry knowledge about [de]serialization process inside `PythonTask` and outside, but it should be only within `PythonTask`.

> The purpose of this class `PythonTask` is to act as a wrapper, not a wrapper/deserializer.

`PythonTask` does wrapping of func and produces a serialized func-object, thus it knows the order of `loads` calls and a structure of a serialized object (`{'func': .., 'args': .., 'kwargs': ..}`). You shouldn't move this knowledge outside of the class.

AymenFJA: Fixed in the latest commit.

mtitov (approved):
I'm ok to merge it with the current state into `raptor_workers` and move leftover comments there

andre-merzky:
> I'm ok to merge it with the current state into `raptor_workers` and move leftover comments there

Yes, lets do that.  Thanks for working through the PR (both of you) :-)

PR #2552, opened 2022-03-21
Resource config for Crusher@OLCF/ORNL

mtitov:
Requires prepared VE: either use provided python module with `module load cray-python` or have a manual installation of conda (not provided by system)


andre-merzky (approved):
LGTM, thank you!
What is the bootstrapper's problem when creating a VE with that `cray-python` module?

mtitov:
> What is the bootstrapper's problem when creating a VE with that cray-python module?

looks like compute nodes do not have access to outside, initially I did set to create one, but got the error below, then switched to re-use a prepared VE
```
# Download virtualenv tgz
# cmd: curl -1 -k -L -O 'https://files.pythonhosted.org/packages/66/f0/6867af06d2e2f511e4e1d7094ff663acdebc4f15d4a0cb0fed1007395124/virtualenv-16.7.5.tar.gz'
#
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:--  0:02:10 --:--:--     0
curl: (7) Failed to connect to files.pythonhosted.org port 443: Connection timed out
```

andre-merzky:
I see, same issue as always then...  I would really like to figure out how to do `pip install` via a tunnel, but no luck so far...

PR #2562, opened 2022-03-28
envvar directive usage. Fixes #1578

### LINKED ISSUE #1578: Documentation: Table with all the enviroment variables of rp
A table with all the environment variables of radical pilot is needed (RADICAL_PILOT_VERBOSE, RADICAL_PILOT_PROFILE ..)

kartikmodi:


docs/source/user_guide/12_list_of_environment_variables.rst
```diff
@@ -0,0 +1,48 @@
...
+.. list-table:: User List
+   :widths: 35 100 20 20
+   :header-rows: 1
+
+   * - Environment Variable Name
+     - Description
+     - Default value
+     - Is_deprecated
+     - Importance
+   * - .. envvar:: RADICAL_UTILS_NO_ATFORK
```
eirrgang, line 13: The purpose of a directive like `.. envvar::` is to define a role that you will use later. It should appear on a line by itself exactly once in the project. Depending on what makes the most sense, you could have a list of directives to define environment variables in a single rst document, or, if there is a clear "owner" of the env var, you could put the directive in the corresponding module documentation.

Then, if it is helpful, you could create tables like these that use the roles (not the directives) to identify the env vars.

In such an approach, the repeated references could make more sense:
* at the `.. envvar:: RADICAL_UTILS_NO_ATFORK` you would provide a description of what the environment variable means and does.
* In a pair of tables that describe environment variables relevant to users or developers, you would focus your descriptive text on the target audience, and why they would want to use the particular env var.

In any case, it probably makes sense to first focus on getting the `.. envvar::` directives into the package _somewhere_ so that the documentation exists and is referentiable. _Then_, if tables like these are necessary, they can be tackled afterwards.

src/radical/pilot/pilot.py
```diff
@@ -20,9 +20,11 @@ class Pilot(object):
     """
     A Pilot represent a resource overlay on a local or remote resource.
 
+    .. envvar:: RADICAL_TEST_ENV
```
eirrgang, line 23: I'm not sure what you are trying to demonstrate here. There is a directive, but you aren't using it to provide any explanatory text about the meaning of the environment variable.

src/radical/pilot/task_description.py
```diff
@@ -85,6 +85,12 @@ class TaskDescription(ru.TypedDict):
     :meth:`radical.pilot.TaskManager.submit_tasks` to instantiate and run
     a new task.
 
+    .. envvar:: RADICAL_TEST_ENV
```
eirrgang, line 88: It doesn't make sense to have more than one directive for the same documented entity. (Also, Sphinx does not play well with it, which you'll notice if you ever accidentally include a document in more than one toctree; directives, like toctrees, establish documentation _nodes_, and nodes need to be uniquely identifiable).

If an environment variable is relevant to more than one software component, and neither component "owns" the environment variable, then they should probably just reference it as a role. If some local explanatory text is appropriate in addition to the link that the role provides, that's fine. E.g.
```
Relevant environment variables
-------------------------------
 * :envvar:`RADICAL_TEST_ENV` is used by TaskDescription to blahblah.
 * :envvar:`RADICAL_TEST_ENV2`, if defined, causes yaddayadda

**Review by eirrgang — COMMENTED** [review]
Note also: user_guide/index.rst
* does not have `12_list_of_environment_variables` in the toc list
* has the wrong name for `11_containerized_compute_tasks`

File: docs/source/user_guide/12_list_of_environment_variables.rst (hunk 1 above)
Comment (kartikmodi) [reply docs/source/user_guide/12_list_of_environment_variables.rst]: > does not have 12_list_of_environment_variables in the toc list
Added now in toctree. 

>In any case, it probably makes sense to first focus on getting the .. envvar:: directives into the package somewhere so that the documentation exists and is referentiable
I am trying to create a page where we'll have this .. envvar:: directives and then we can refer to those variable throughout the codebase. I do understand the definitions .. envvar:: directives should be unique in one place, else sphinx picks the last definition it finds and link to that definition. 
I sure there is more to this, you are saying and I ain't getting all of it.

File: src/radical/pilot/pilot.py (hunk 2 above)
Comment (kartikmodi) [reply src/radical/pilot/pilot.py:23]: Sorry I was not supposed to commit this. But I had to demonstrate sphinx behaviour in cases of multiple declarations/definitions of envvar with the same name, and then usage reference behavior. I have removed this.

**Comment by kartikmodi** [comment]
Current State of generated page (just to give a look n feel) - 
image

File: docs/source/user_guide/12_list_of_environment_variables.rst [hunk 4]
```diff
@@ -0,0 +1,149 @@
...
+    * - .. envvar:: RADICAL_PILOT_BULK_CB
+      - Enable bulk callbacks for performance boost. This changes the callback signature.
+      - FALSE
+    * - .. envvar:: RADICAL_PILOT_STRICT_CANCEL
+      - Cancel the tasks immediately. #TODO Didn't understood
+      - {NOT_SET}
+    * - .. envvar:: RADICAL_SAGA_PTY_SSH_TIMEOUT
+      - Sets the connection timeout (in seconds, default: 10) for the SAGA PTY layer. On connection timeouts should be set to 60 or more
+      - `10.0`
+    * - .. envvar:: RADICAL_DEFAULT_LOG_TGT
```
kartikmodi, line 36: Do we need these *DEFAULT* env vars ? If yes, then in user list?

docs/source/user_guide/12_list_of_environment_variables.rst
```diff
@@ -0,0 +1,149 @@
...
+
+
+.. list-table:: User List
+    :widths: 35 100 20
+    :header-rows: 1
+
+    * - Environment Variable Name
+      - Description
+      - Default value
+    * - .. envvar:: RADICAL_BASE
```
eirrgang, line 18:
```suggestion
    * - :envvar:`RADICAL_BASE`
```
Since this is not the place that you are providing a canonical documentation object, you should use the *role*, not the *directive*. It will format the way you want and will automatically link once you add a directive in the separate doc you describe.

If I have misunderstood you, and you intend for *this* to be the document that provides the definition text for the environment variable, then I think it hurts the documentation to (mis)use the directives in this way.

1. It is useful to let the directive tell sphinx what the definition is so that it can provide more than just a link.
2. Are you sure that each environment variable will fall neatly into one of these lists exactly once?

If the formatting of the environment variable definition (or of the text around any directive) is really important, you should probably be customizing a style sheet (or writing a custom directive) so that you can use the directives in a more conventional way and let sphinx do the formatting for you as a special case (decoupled from writing the documentation). But, as with LaTeX, it's probably best not to worry too much about the exact formatting, or at least to worry about the content first and the formatting second. Raw table syntax is always problematic in documentation because it is both hard to maintain the desired formatting consistently and it breaks up information that should be tightly programmatically coupled. I.e. once you've put useful information in three syntactically separate places (three table cells), neither Sphinx nor any other tool can use the relationship you intend between "RADICAL_BASE", "Root dir for temporary state files", and "`$HOME/.radical`".

docs/source/user_guide/12_list_of_environment_variables.rst
```diff
@@ -0,0 +1,184 @@
...
+      - MongoDB URL string and mandatory for RP to work.
+      - {NOT_SET}
+    * - .. envvar:: RADICAL_UTILS_NTPHOST
+      - NTP host to be used for profile syncing
+      - `0.pool.ntp.org`
+    * - .. envvar:: RADICAL_PILOT_BULK_CB
+      - Enable bulk callbacks for performance boost. This changes the callback signature.
+      - FALSE
+    * - .. envvar:: RADICAL_PILOT_STRICT_CANCEL
+      - Cancel the tasks immediately. #TODO Didn't understood
```
kartikmodi, line 25: #TODO @mtitov Need more details here. You can either comment/or push changes.

docs/source/user_guide/12_list_of_environment_variables.rst
```diff
@@ -0,0 +1,184 @@
...
+      - {NOT_SET}
+    * - .. envvar:: RADICAL_DEFAULT_LOG_TGT
+      - The log target. #TODO Need more clarification
+      - .
+    * - .. envvar:: RADICAL_DEFAULT_LOG_DIR
+      - The Log dir
+      - $PWD
+    * - .. envvar:: RADICAL_DEFAULT_LOG_LVL
+      - #TODO Description
```
kartikmodi, line 34: #TODO @mtitov I understand these *DEFAULT*, but don't know what to write in their description.

docs/source/user_guide/12_list_of_environment_variables.rst
```diff
@@ -0,0 +1,184 @@
...
+.. list-table:: Logging-related Env vars.
+    `ru.Logger` instances have a name and a name space.
+    :widths: 35 100 20
+    :header-rows: 1
+
+    * - Environment Variable Name
+      - Description
+      - Default value
+    * - <NS>_LOG_LVL
+      - Used for log level. #TODO Description
```
kartikmodi, line 65: #TODO @mtitov  Need better description here.

docs/source/user_guide/12_list_of_environment_variables.rst
```diff
@@ -0,0 +1,184 @@
...
+
+    * - Environment Variable Name
+      - Description
+      - Default value
+    * - <NS>_REPORT
+      - Boolean to turn on and off Reporter
+      - TRUE
+    * - <NS>_LOG_TGT
+      - Where to report to
+      - #TODO
```
kartikmodi, line 89: #TODO @mtitov Dont know the default values

docs/source/user_guide/12_list_of_environment_variables.rst
```diff
@@ -0,0 +1,184 @@
...
+      - {NOT_SET}
+    * - RU_RAISE_ON_*
+      - related to :envvar:`RADICAL_DEBUG` to trigger specific exceptions
+      - {NOT_SET}
+    * - .. envvar:: RADICAL_DEBUG_HELPER
+      - related to :envvar:`RADICAL_DEBUG`, enables a persistent debug helper class in the code and installs some signal handlers for extra debug output.
+      - {NOT_SET}
+    * - .. envvar:: RADICAL_DEBUG_VERBOSE
+      - related to :envvar:`RADICAL_DEBUG`, enables verbose messages for debugging.
+      - #TODO
```
kartikmodi, line 116: Dont know the default value @mtitov

docs/source/user_guide/12_list_of_environment_variables.rst
```diff
@@ -0,0 +1,184 @@
...
+    * - *_PROFILE
+      - `Profiler` is similar to `Logger` and `Reporter`
+      - {NOT_SET}
+    * - .. envvar:: RADICAL_PILOT_PRUN_VERBOSE
+      - Increase verbosity of prun output
+      - FALSE
+    * - .. envvar:: UMS_OMPIX_PRRTE_DIR
+      - #TODO Not understood
```
kartikmodi, line 124: Dont know this variable @mtitov

docs/source/user_guide/12_list_of_environment_variables.rst
mtitov: Limit task cancelation, not to force state "CANCELLED" on a Task Manager side. Note that corresponding command for pilot(s) to cancel tasks is published anyway.

p.s., [code](https://github.com/radical-cybertools/radical.pilot/blob/34db28c651a580276a660b209beed8cfe101bfdc/src/radical/pilot/task_manager.py#L1098-L1113)

docs/source/user_guide/12_list_of_environment_variables.rst
mtitov: Default value of a logging level ["DEBUG", "INFO", "WARNING", "ERROR", "CRITICAL"] to control the debug output.

@andre-merzky do we actually need such env variables as `.._DEFAULT_..`? For example `RADICAL_DEFAULT_LOG_LVL` is treated the same as `RADICAL_LOG_LVL`

docs/source/user_guide/12_list_of_environment_variables.rst
mtitov: Logging level ["DEBUG", "INFO", "WARNING", "ERROR", "CRITICAL"] to control the debug output for a corresponding namespace NS, where NS can be applied as for a specific package (e.g., "RADICAL_PILOT_LOG_LVL" or "RADICAL_UTILS_LOG_LVL") or for a whole stack (e.g., "RADICAL_LOG_LVL").

docs/source/user_guide/12_list_of_environment_variables.rst
mtitov: List of comma separated targets ["0"/"null", "1"/"stdout", "2"/"stderr", "."/"<log_name>"] to write the debug output for a corresponding namespace NS.

docs/source/user_guide/12_list_of_environment_variables.rst
mtitov: Controls "debug" module to collect stack traces. Verbose flag sets the level of details for output messages.

docs/source/user_guide/12_list_of_environment_variables.rst
mtitov: Installation directory for PMIx/PRRTE used in RP LM PRTE (optional, to be obsolete).

kartikmodi:
Output - 
Screenshot 2022-08-29 at 12-23-45 3 13  List of Env vars — RADICAL-Pilot 1 16 0 documentation

andre-merzky (approved):
This looks sufficiently complete and useful to get merged.

PR #2567, opened 2022-03-30
Fix/issue 2556

AymenFJA:
This PR fixes: if the code is changing the `current working directory` of the session somehow (Colmena specifically). Thanks to Andre this is now fixed.


andre-merzky (approved):
LGTM :-D

src/radical/pilot/session.py
```diff
@@ -295,8 +295,8 @@ def close(self, **kwargs):
 
             self._prof.prof("session_fetch_start", uid=self._uid)
             self._log.debug('start download')
-            tgt = os.getcwd()
-            self.fetch_json    (tgt='%s/%s' % (tgt, self.uid))
+            tgt = self._cfg.base
+            self.fetch_json    (tgt='%s/%s' % (self._cfg.path, self.uid))
```
mtitov, line 299: should both changes be `self._cfg.base`? (`path` contains `base`)
(*)
```python
            if not self._cfg.path:
                self._cfg.path = '%s/%s' % (self._cfg.base, self._cfg.sid)
```

andre-merzky: Yes, otherwise we end up with `base/sid/sid` like path's.  Thanks for checking though, I got that very thing wrong on the first fix attempt :-)

mturilli:
Note the failure message of the analytics test: `FileNotFoundError: [Errno 2] No such file or directory: '/home/runner/work/radical.pilot/radical.pilot/rp.session.fv-az213-85.runner.019083.0000/rp.session.fv-az213-85.runner.019083.0000/rp.session.fv-az213-85.runner.019083.0000.json` Maybe some issue with path repetition?

PR #2572, opened 2022-04-03
Fixed session path

mtitov:


andre-merzky (approved):
WTH - why did I merge #2567??  Thanks @mtitov !

PR #2535, opened 2022-03-04
Feature/psi j

andre-merzky:
This PR implements 2 things:
1) the default `pmgr.launching` component is split into two parts: a generic part in `base`, and a part which provides pilot launching via RS (`default`). 
2) an alternative implementation of the latter which provides pilot launching via PSIJ

PR #2574, opened 2022-04-04
re-apply psi-j diff

andre-merzky:
replace reverted PR #2535


mtitov:
@andre-merzky two modules from `pmgr.launching` are missing here: `saga` and `psi_j`

src/radical/pilot/pmgr/launching/psi_j.py
```diff
@@ -0,0 +1,186 @@
+
+__copyright__ = "Copyright 2013-2016, http://radical.rutgers.edu"
+__license__   = "MIT"
```
mtitov, line 3: really off topic, but better to keep up to date
```python
__copyright__ = 'Copyright 2022, The RADICAL-Cybertools Team'
__license__   = 'MIT'
```

src/radical/pilot/pmgr/launching/saga.py
```diff
@@ -0,0 +1,268 @@
...
+    # --------------------------------------------------------------------------
+    #
+    def cancel_pilots(self, pids):
+
+        tc = rs.job.Container()
+
+        for pid in pids:
+
+            job   = self._saga_jobs[pid]
+            pilot = self._pilots[pid]['pilot']
```
mtitov, line 154: `self._pilots` was not defined in `__init__`
(and then also should be filled up in `launch_pilots`)

src/radical/pilot/pmgr/launching/saga.py
```diff
@@ -0,0 +1,268 @@
...
+
+    # --------------------------------------------------------------------------
+    #
+    def __init__(self, name, log, prof, state_cb):
+
+        PilotLauncherBase.__init__(self, name, log, prof, state_cb)
+
+        self._saga_jobs = dict()      # pid      : rs.Job
+        self._saga_js   = dict()      # resource : rs.JobService
+        self._saga_lock = ru.RLock()  # lock for above
```
mtitov, line 40: `mt.RLock`?

src/radical/pilot/pmgr/launching/base.py
```diff
@@ -25,34 +123,1121 @@ def __init__(self, cfg, session):
...
+    #
+    def _stage_in(self, pilot, sds):
+        '''
+        Run some input staging directives.
+        '''
+
+        resource_sandbox = self._session._get_resource_sandbox(pilot)
+      # session_sandbox  = self._session._get_session_sandbox (pilot)
+        pilot_sandbox    = self._session._get_pilot_sandbox   (pilot)
+        client_sandbox   = self._session._get_client_sandbox()
```
mtitov, line 1131: these values are already defined within pilot
https://github.com/radical-cybertools/radical.pilot/blob/416b0ff39c0aa1559cfb72b383d9fbc07464cab7/src/radical/pilot/pmgr/launching/base.py#L696-L705

src/radical/pilot/pmgr/launching/base.py
```diff
@@ -25,34 +123,1121 @@ def __init__(self, cfg, session):
...
+    #
+    def _stage_out(self, pilot, sds):
+        '''
+        Run some output staging directives.
+        '''
+
+        resource_sandbox = self._session._get_resource_sandbox(pilot)
+      # session_sandbox  = self._session._get_session_sandbox (pilot)
+        pilot_sandbox    = self._session._get_pilot_sandbox   (pilot)
+        client_sandbox   = self._session._get_client_sandbox()
```
mtitov, line 1164: same - just use `pilot` with corresponding keys

src/radical/pilot/pmgr/launching/saga.py
```diff
@@ -0,0 +1,268 @@
...
+                    jd[key] = val
+
+            # set saga job description attribute based on env variable(s)
+            # FIXME: is this used?
+            if os.environ.get('RADICAL_SAGA_SMT'):
+                try:
+                    jd.system_architecture['smt'] = \
+                        int(os.environ['RADICAL_SAGA_SMT'])
+                except Exception as e:
+                    self._log.debug('SAGA SMT not set: %s' % e)
```
mtitov, line 122: we can remove this, since `smt` is set in base module

p.s. in base module we can either rename env variable (e.g., `RADICAL_SMT`) or even do not use at all (just use `smt` through config only)

src/radical/pilot/pmgr/launching/saga.py
```diff
@@ -0,0 +1,268 @@
...
+            pid = pilot['uid']
+
+            # FIXME: update the right pilot
+            with self._saga_lock:
+                self._saga_jobs[pid] = j
+
+
+    # --------------------------------------------------------------------------
+    #
+    def cancel_pilots(self, pids):
```
mtitov, line 147: this method is not called in base module

src/radical/pilot/pmgr/launching/base.py
```diff
@@ -25,34 +123,1121 @@ def __init__(self, cfg, session):
...
+            self.unregister_input(rps.PMGR_LAUNCHING_PENDING,
+                                  rpc.PMGR_LAUNCHING_QUEUE, self.work)
+
+            # FIXME: always kill all pilot jobs for non-final pilots at
+            #        termination, and set the pilot states to CANCELED.
+            with self._pilots_lock:
+                pids = list(self._pilots.keys())
+
+            self._cancel_pilots(pids)
+            self._kill_pilots(pids)
```
mtitov, line 209: `_cancel_pilots` is called within `_kill_pilots`, should it be removed either from here or from `_kill_pilots`?

src/radical/pilot/pmgr/launching/base.py
```diff
@@ -25,34 +123,1121 @@ def __init__(self, cfg, session):
...
+                pilots = list()
+                for pid in pids:
+
+                    pilot = self._pilots['uid']['pilot']
+                    if pilot['state'] in rps.FINAL:
+                        continue
+
+                    pilots.append(pilot)
+
+                self._cancel_pilots(pilots)
```
mtitov, line 353: should be provided `pids`

src/radical/pilot/pmgr/launching/psi_j.py
```diff
@@ -0,0 +1,186 @@
...
+            spec.arguments   = jd['arguments']
+            spec.environment = jd['environment']
+            spec.directory   = jd['working_directory']
+            spec.stdout_path = jd['output']
+            spec.stderr_path = jd['error']
+
+            spec.resources   = psij.ResourceSpecV1()
+            spec.resources.process_count         = 1
+            spec.resources.cpu_cores_per_process = jd['total_cpu_count']
+            spec.resources.gpu_cores_per_process = jd['total_gpu_count']
```
mtitov, line 163: `process_count = jd.total_cpu_count`? (since `total_cpu_count = (requested_nodes * cores_per_node) or requested_cores`)
and `processes_per_node = jd.processes_per_host`?

mtitov (commented):
have several comments below

src/radical/pilot/pmgr/launching/base.py
```diff
@@ -25,34 +123,1121 @@ def __init__(self, cfg, session):
                                    ru.ID_CUSTOM)
 
         rpu.Component.__init__(self, cfg, session)
-
         self._pmgr = self._owner
 
+        self._pilots        = dict()      # dict for all known pilots
```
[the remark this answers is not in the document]
andre-merzky: Linter does not like it.  We may want to reconsider the `initialize` anyway, there are only a few cases where it is actually useful, and it convolutes the code a bit.  But low priority...

src/radical/pilot/pmgr/launching/base.py
```diff
@@ -25,34 +123,1121 @@ def __init__(self, cfg, session):
...
+        cores_per_node          = rcfg.get('cores_per_node', 0)
+        gpus_per_node           = rcfg.get('gpus_per_node',  0)
+        blocked_cores           = rcfg.get('blocked_cores', [])
+        blocked_gpus            = rcfg.get('blocked_gpus',  [])
+        lfs_path_per_node       = rcfg.get('lfs_path_per_node')
+        lfs_size_per_node       = rcfg.get('lfs_size_per_node', 0)
+        python_dist             = rcfg.get('python_dist')
+        task_tmp                = rcfg.get('task_tmp')
+        spmd_variation          = rcfg.get('spmd_variation')
+        shared_filesystem       = rcfg.get('shared_filesystem', True)
```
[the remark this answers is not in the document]
andre-merzky: Ha, I feel like I remove this the 3rd time or so :-P  Done, thanks.

src/radical/pilot/pmgr/launching/base.py
```diff
@@ -25,34 +123,1121 @@ def __init__(self, cfg, session):
...
+            #  e : everything (== pilot sandbox)
+            if shared_filesystem:
+                cleanup = 'luve'
+            else:
+                # we cannot clean the sandbox from within the agent, as the hop
+                # staging would then fail, and we'd get nothing back.
+                # FIXME: cleanup needs to be done by the pmgr.launcher, or
+                #        someone else, really, after fetching all logs and
+                #        profiles.
+                cleanup = 'luv'
```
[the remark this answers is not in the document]
andre-merzky: same

src/radical/pilot/pmgr/launching/base.py
```diff
@@ -25,34 +123,1121 @@ def __init__(self, cfg, session):
...
+
+            # we never cleanup virtenvs which are not private
+            if virtenv_mode != 'private':
+                cleanup = cleanup.replace('v', '')
+
+        # estimate requested resources
+
+        smt = os.environ.get('RADICAL_SAGA_SMT')
+        if smt:
+            system_architecture['smt'] = int(smt)
```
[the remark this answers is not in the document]
andre-merzky: But that would imply that we have to always specify `smt` in the resource config and cannot let it default to 1?

Alternatively, we could consider getting rid of the env variable altogether and only use the resource config settings - I don't think anybody uses that env really, or do you?

src/radical/pilot/pmgr/launching/base.py
```diff
@@ -25,34 +123,1121 @@ def __init__(self, cfg, session):
...
+
+        if avail_gpus_per_node and blocked_gpus:
+            avail_gpus_per_node -= len(blocked_gpus)
+            assert (avail_gpus_per_node >= 0)
+
+        if requested_nodes:
+
+            if not cores_per_node:
+                raise RE('use "cores" in PilotDescription')
+
```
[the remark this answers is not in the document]
andre-merzky: Ack - I added an explicit default setting now:
```py
        if not requested_nodes and not requested_cores:
            requested_nodes = 1
```

src/radical/pilot/pmgr/launching/base.py
```diff
@@ -25,34 +123,1121 @@ def __init__(self, cfg, session):
...
+        if self._prof.enabled:
+            jd_dict.environment['RADICAL_PROFILE'] = 'TRUE'
+
+        jd_dict.environment['RADICAL_BASE'] = resource_sandbox
+
+        # for condor backends and the like which do not have shared FSs, we add
+        # additional staging directives so that the backend system binds the
+        # files from the session and pilot sandboxes to the pilot job.
+        jd_dict.file_transfer = list()
+        if not shared_filesystem:
```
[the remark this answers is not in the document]
andre-merzky: same as above - done.

src/radical/pilot/pmgr/launching/psi_j.py
andre-merzky: done

src/radical/pilot/pmgr/launching/saga.py
andre-merzky: This has been reworked (notifications were incomplete...)

src/radical/pilot/pmgr/launching/saga.py
andre-merzky: ack

src/radical/pilot/pmgr/launching/base.py
andre-merzky: Fair, changed it.  Although the `_get_*_sandbox` methods also cache, so it's no overhead really.

src/radical/pilot/pmgr/launching/base.py
andre-merzky: same

src/radical/pilot/pmgr/launching/saga.py
andre-merzky: Ah, great - let me remove it then...

src/radical/pilot/pmgr/launching/psi_j.py
andre-merzky: Semantics is somewhat different here than in SAGA.  Basically, `process_count = 2` would mean 2 ranks, so we would get 2 bootstrappers.

src/radical/pilot/pmgr/launching/saga.py
andre-merzky: Ah, thanks - underscore was missing...

src/radical/pilot/pmgr/launching/base.py
andre-merzky: ack, done.

src/radical/pilot/pmgr/launching/base.py
andre-merzky: Thanks, fixed.

src/radical/pilot/pmgr/launching/saga.py
mtitov: I actually meant that there is no calls `launcher._cancel_pilots` (use an update method name here), where `launcher` is an object from `self._launchers` in base module

src/radical/pilot/pmgr/launching/base.py
mtitov: should it be outside of the block `with self._pilots_lock:`? since it has same lock inside itself...

src/radical/pilot/pmgr/launching/psi_j.py
mtitov: I also used a corresponding template to check how it is mapped - https://github.com/ExaWorks/psi-j-python/blob/main/src/psij/executors/batch/slurm/slurm.mustache

src/radical/pilot/pmgr/launching/base.py
```diff
@@ -1081,6 +1070,7 @@ def _prepare_pilot(self, resource, rcfg, pilot, expand, tar_name):
         jd_dict.wall_time_limit       = runtime
         jd_dict.queue                 = queue
         jd_dict.candidate_hosts       = candidate_hosts
+        jd_dict.file_transfer         = list()
```
mtitov, line 1073: just a note that this attribute is not used anywhere else - we keep it just for legacy?

andre-merzky: Yes.

src/radical/pilot/pmgr/launching/psi_j.py
andre-merzky: Not sure what you try to say, sorry :-)

src/radical/pilot/pmgr/launching/base.py
andre-merzky: Its an `RLock`, i.e. a reentrant lock which allows the same thread to lock more than once, for exactly this situation - that that should work ok.

src/radical/pilot/pmgr/launching/saga.py
andre-merzky: Ah, stupid me - cancel was indeed incomplete.  I fixed that now, but that needed a number of changes actually (need to remember what launcher submitted a pilot so that we can use the same launcher for the cancel call).  You may want to directly look at this commit: 71d3379d6

src/radical/pilot/pmgr/launching/base.py
```diff
@@ -595,26 +608,35 @@ def _start_pilot_bulk(self, resource, schema, pilots):
...
-        self._log.warn('rcfg: %s', pprint.pformat(rcfg))
-        self._log.warn('pilots: %s', pprint.pformat(pilots))
-        raise RuntimeError('no launcher found for pilots')
+        for pilot in pilots:
+            for lname,launcher in self._launchers.items():
+                if launcher.can_launch(rcfg, pilots):
+                    pilot['launcher'] = lname
+                    self._log.info('use launcher %s for pilot %s',
+                                   lname, pilot['uid'])
+                    return launcher
```
mtitov, line 635: - maybe be change the method name to `set_pilot_launchers`?
   - no need to return `launcher` since it is not used
- no need to provide `rcfg`, that information is already in each pilot as `pilot['cfg']['resource_cfg']` after `_prepare_pilot` is called

src/radical/pilot/pmgr/launching/base.py
```diff
@@ -595,26 +608,35 @@ def _start_pilot_bulk(self, resource, schema, pilots):
...
-        for pilot in pilots:
-            self._prof.prof('submission_stop', uid=pilot['uid'])
+        for lname,launcher in self._launchers.items():
+            bucket = list()
+            for pilot in pilots:
+                if pilot['launcher'] == lname:
+                    bucket.append(pilot)
+
+            if bucket:
+                launcher.launch_pilots(rcfg, bucket)
```
mtitov, line 620: `rcfg` also not needed here, pilots have all that

src/radical/pilot/pmgr/launching/base.py
```diff
@@ -265,14 +266,26 @@ def _cancel_pilots(self, pids):
...
+                pilot = self._pilots[pid]
+                lname = pilot['launcher']
+
+                if lname not in self._launchers:
+                    self._log.warn('invalid pilot launcher name: %s', lname)
+                    continue
+
+                launcher = self._launchers[lname]
+                try:
+                    launcher.cancel_pilots([pid])
```
mtitov, line 286: currently launcher classes have `_cancel_pilots` methods, better to keep them without starting `_`, since they are public methods

src/radical/pilot/pmgr/launching/base.py
```diff
@@ -25,34 +123,1105 @@ def __init__(self, cfg, session):
...
+
+
+        self._log.debug('killing pilots: kill! %s', pids)
+        try:
+            with self._pilots_lock:
+
+                pilots = list()
+                for pid in pids:
+
+                    pilot = self._pilots['uid']['pilot']
```
mtitov, line 362: `pilot = self._pilots[pid]['pilot']`

btw, I cannot figure out how `self._pilots` is filled out with pilots, did I miss it or it is actually missed?

src/radical/pilot/pmgr/launching/psi_j.py
andre-merzky: TODO AM: fix

src/radical/pilot/pmgr/launching/base.py
```diff
@@ -25,34 +123,1121 @@ def __init__(self, cfg, session):
...
+        jd_dict.arguments             = shlex.split('-l ./bootstrap_0.sh %s'
+                                                                      % bs_args)
+        jd_dict.working_directory     = pilot_sandbox
+        jd_dict.project               = project
+        jd_dict.output                = 'bootstrap_0.out'
+        jd_dict.error                 = 'bootstrap_0.err'
+        jd_dict.total_cpu_count       = total_cpu_count
+        jd_dict.total_gpu_count       = total_gpu_count
+        jd_dict.total_physical_memory = requested_memory
+        jd_dict.processes_per_host    = cores_per_node
```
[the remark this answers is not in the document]
andre-merzky: TODO AM: apply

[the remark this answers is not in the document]
andre-merzky: TODO AM: apply

src/radical/pilot/pmgr/launching/base.py
```diff
@@ -25,34 +123,1105 @@ def __init__(self, cfg, session):
...
+                return
+
+            # avoid busy poll)
+            time.sleep(1)
+
+        # we don't want the watcher checking for these pilot anymore
+        with self._check_lock:
+            for pid in pids:
+                if pid in self._checking:
+                    self._checking.remove(pid)
```
andre-merzky, line 352: clean up

andre-merzky, line 352: clean up

src/radical/pilot/pmgr/launching/base.py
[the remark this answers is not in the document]
andre-merzky: done.

src/radical/pilot/pmgr/launching/psi_j.py
andre-merzky: Fixed this now, but I am still uncertain about it really because we had to take out the GPU spec.  Assume a scheduler is clever enough to distinguish between GPU and CPU nodes - we could now suddenly scheduled on CPU-only nodes because we don't include GPU requests.

Anyway, that is more an issue for psij really than for RP to solve.  Let's see how it goes and if it breaks we should address it in psij.

src/radical/pilot/pmgr/launching/base.py
andre-merzky: > no need to return launcher since it is not used
Right, the return was wrong.

> maybe be change the method name to set_pilot_launchers?
Yeah, I did not like that either.  I moved that code now into the calling method as it is so small now anyway...

src/radical/pilot/pmgr/launching/base.py
andre-merzky: Hmm, the pilot does not actually have the complete resource config - or am I missing something?

src/radical/pilot/pmgr/launching/base.py
andre-merzky: done.

src/radical/pilot/pmgr/launching/base.py
andre-merzky: This is fixed now also - thanks for catching!

andre-merzky:
@mtitov : I would not mind merging this before the ExaScale call today, so if you happen to have a minute to have a final look, that would be great :-)

andre-merzky:
This should be ready for review again...

andre-merzky:
@mtitov : ping

src/radical/pilot/pmgr/launching/psi_j.py
mtitov: btw, right away we should add `project` and `queue` to psij job (usually cpu-only and gpu nodes are provided in different queues)

src/radical/pilot/pmgr/launching/base.py
mtitov: it is set within agent's cfg, which then is set to pilot
https://github.com/radical-cybertools/radical.pilot/blob/7f1cfee2b7ebfc2a4b5522159e5739ac18908c85/src/radical/pilot/pmgr/launching/base.py#L880-L884

src/radical/pilot/pmgr/launching/base.py
```diff
@@ -25,34 +118,995 @@ def __init__(self, cfg, session):
...
+    # --------------------------------------------------------------------------
+    #
+    def _pmgr_control_cb(self, topic, msg):
+
+        cmd = msg['cmd']
+        arg = msg['arg']
+
+        self._log.debug('launcher got %s', msg)
+
+        if cmd == 'kill_pilots':
```
mtitov, line 222: then the following should be updated as well?
https://github.com/radical-cybertools/radical.pilot/blob/7f1cfee2b7ebfc2a4b5522159e5739ac18908c85/src/radical/pilot/pilot_manager.py#L819-L822
should

andre-merzky: Thanks, done.

andre-merzky: Thanks, Done.

src/radical/pilot/pmgr/launching/base.py
```diff
@@ -25,34 +118,995 @@ def __init__(self, cfg, session):
...
+
+        total_cpu_count = (requested_nodes * cores_per_node) or requested_cores
+        total_gpu_count = (requested_nodes * gpus_per_node)  or requested_gpus
+
+        jd_dict = ru.TypedDict()
+
+        jd_dict.name                  = job_name
+        jd_dict.executable            = '/bin/bash'
+        jd_dict.arguments             = shlex.split('-l ./bootstrap_0.sh %s'
+                                                                      % bs_args)
```
mtitov, line 967: @andre-merzky this breaks bootstrapping... revert back with `['-l ./bootstrap_0.sh %s' % bs_args]`?

src/radical/pilot/pmgr/launching/psi_j.py
andre-merzky: Right, I added `project`, `queue` and the other job attributes.

`process_per_node` doesn't really fit (unless it means something different than the name suggests), as we want a single process over the whole allocation (several nodes).  I have a deja-vu so, I think we discussed this already?  Let me check again...

src/radical/pilot/pmgr/launching/base.py
andre-merzky: Fair - lets leave it anyway please.  I actually would opt for not adding the agent cfg to the pilot dict.  I think we only did that to get it into MongoDB, but I don't think it has any other purpose really and as such should be removed when we get rid of MongoDB.

andre-merzky: Also, that contradicts the bulk launching somewhat: we potentially launch several pilots with the same resource config.  If each pilot has it's own we need to assume or ensure that they are the same which seems a bit backwards.

src/radical/pilot/pmgr/launching/base.py
andre-merzky: This is fixed now and handles spaced args (the split now happens on the `bs_args` and not on the full arg string).

andre-merzky:
radical-cybertools/radical.analytics/pull/162 should fix the failing analytics test.

src/radical/pilot/pmgr/launching/base.py
mtitov: ack on leave it for now!

> we potentially launch several pilots with the same resource config.

pilots are split among buckets, thus pilots with the same resource and schema_access are grouped together, that seems legit

src/radical/pilot/pmgr/launching/psi_j.py
mtitov: this is part of `spec.resources` - https://github.com/ExaWorks/psi-j-python/blob/main/src/psij/resource_spec.py#L50

src/radical/pilot/pmgr/launching/base.py
```diff
@@ -25,34 +118,995 @@ def __init__(self, cfg, session):
...
+
+        total_cpu_count = (requested_nodes * cores_per_node) or requested_cores
+        total_gpu_count = (requested_nodes * gpus_per_node)  or requested_gpus
+
+        jd_dict = ru.TypedDict()
+
+        jd_dict.name                  = job_name
+        jd_dict.executable            = '/bin/bash'
+        jd_dict.arguments             = ['-l', './bootstrap_0.sh'] \
+                                      + shlex.split(bs_args)
```
mtitov, line 967: sorry, still not resolved, it removes quotations
```
>>> import shlex
>>> bs_args = " -p 'abc.123' -e 'module load a'"
>>> shlex.split(bs_args)
['-p', 'abc.123', '-e', 'module load a']
```
and then option `-e` has value `module`

should we collect this `bs_args` as list and while adding a value apply a corresponding method from RU?
```python
bs_args += ['-e', ru.sh_quote(arg)]
```

src/radical/pilot/utils/prof_utils.py
```diff
@@ -961,6 +961,22 @@ def get_consumed_resources(session, rtype='cpu', tdurations=None):
...
+
+        # fallback for missing bootstrap events
+        if p_min is None: p_min = pilot.timestamps(state='PMGR_ACTIVE')
+        if p_max is None: p_max = pilot.events[-1][ru.TIME]
+
+        assert(p_min is not None)
+        assert(p_max is not None)
+
         p_min = pt(event=PILOT_DURATIONS['consume']['idle'][0]) [0]
         p_max = pt(event=PILOT_DURATIONS['consume']['idle'][1])[-1]
```
mtitov, line 981: section with analytics having failure with `p_max` while trying to get event
```
{ru.EVENT: 'cmd'              ,
 ru.MSG  : 'cancel_pilot'     }
```

Shouldn't that last two lines (L980,981) be part of the code block above?

p.s. also still cannot figure out why `cancel_pilot` could be missed, as an example on Summit I had that event
```
./pilot.0000/agent.0.prof:1649780740.4137278,cmd,agent.0,Thread-9,pilot.0000,,cancel_pilot : []
```

andre-merzky: > also still cannot figure out why cancel_pilot could be missed,

the event will be missing if the pilot does not cancel on it's own fast enough and the pmgr kills the pilot job instead.

src/radical/pilot/pmgr/launching/base.py
andre-merzky: This is fixed now - we assemble the args as proper list now and leave all quoting to the SAGA  and PSIJ launchers.

mtitov (approved):
LGTM! Thanks!

Issue #2573, opened 2022-04-04
Feature Request: run in an interactive job


andre-merzky:
The agent will always assume it's running with the local resource manager - but we should be able to configure other resource managers.  However, those will not see the batch system's environment - that would need to be communicated from client to agent.

See radical-collaboration/hpc-workflows/issues/154

andre-merzky:
@AymenFJA : apply for ExaWorks alloc on Summit (PI: Dan Laney)

mtitov:
Project: CSC449 ("2.3.5.10 ExaWorks")

AymenFJA:
@mtitov thanks :).

AymenFJA:
Currently, I am facing this problem: 
```shell
  radical.saga.log BadParameter: 'JobDescription.Project' (csc449) not supported by radical.saga.adaptors.shell_job

 radical.saga.pty.log (ERROR\nNOK - command 'PING' failed\n/ccs/home/aymen/.radical/saga/adaptors/shell_job/wrapper.sh: line 990: PING: command not found\n)
```

AymenFJA:
> Currently, I am facing this problem:
> 
> ```shell
>   radical.saga.log BadParameter: 'JobDescription.Project' (csc449) not supported by radical.saga.adaptors.shell_job
> 
>  radical.saga.pty.log (ERROR\nNOK - command 'PING' failed\n/ccs/home/aymen/.radical/saga/adaptors/shell_job/wrapper.sh: line 990: PING: command not found\n)
> ```

Update: Based on Andre's advice (today's devel meeting) we decided to: 

1- The shell_adaptor does not support `PROJECT` attributes. So we need to find an environment variable that can expose `summit` project number to construct the pilot_sandbox path.
2- Create an individual resource_config `summit_interactive`.
3- for the `command 'PING' failed` delete `ccs/home/aymen/.radical`.

AymenFJA:
Update:

1.  I couldn't find an environment variable that exposes the `Project number of summit`. In my understanding, we have to find a way to ask the user to provide it.

andre-merzky:
Dammit.  Maybe we can run `bjobs` and grep for the job id (which we know).  There might be a more efficient command to get the job info?

AymenFJA:
> Dammit.  Maybe we can run `bjobs` and grep for the job id (which we know).  There might be a more efficient command to get the job info?

@andre-merzky figured it out. `
(base) bash-4.4$ env | grep CS449
[...]

PR #2575, opened 2022-04-11
feature interactive job

AymenFJA:
This PR is based on #2573 (the next step is summit) provides the resource config file to run RP within an interactive job on multiple compute nodes. Example on TACC Frontera: 

```shell
srun --pty -A xxxx -N 2 -n 56 -t 01:00:00 -p development /bin/bash -l #  ask for 2 nodes interactively on Frontera
python 00_getting_started.py # run RP within these requested resources.
```

Note this config can be used to submit RP within `sbatch job`. Example from Frontera login node: 

1. Create an sbatch shell (rp_example.sh): 

```shell

#!/bin/bash
#----------------------------------------------------

#SBATCH -J 00_getting_started           # Job name
#SBATCH -o 00_getting_started.o%j       # Name of stdout output file
#SBATCH -e 00_getting_started.e%j       # Name of stderr error file
#SBATCH -p development     # Queue (partition) name
#SBATCH -N 2               # Total # of nodes
#SBATCH -n 56              # Total # of mpi tasks
#SBATCH -t 01:00:00        # Run time (hh:mm:ss)
#SBATCH -A XXXX            # Project/Allocation name (req'd if you have more than 1)

export RADICAL_LOG_LVL="DEBUG"
export RADICAL_PROFILE="TRUE"
export RADICAL_PILOT_DBURL=mongodb://user:password@ip/db_name

python 00_getting_started.py tacc.frontera_interactive

```

2. `sbatch rp_example.sh`


src/radical/pilot/configs/resource_tacc.json
```diff
@@ -1,5 +1,42 @@
...
 {
+    "frontera_interactive": {
+        "description"                 : "Petascale computing system at the Texas Advanced Computing Center (TACC)",
+        "notes"                       : "Specify this resource in the pilot description and run RP within a compute node(s)",
+        "schemas"                     : ["local", "ssh"],
+        "ssh"                         :
+        {
+            "job_manager_endpoint"    : "ssh://localhost/",
+            "filesystem_endpoint"     : "sftp://localhost/"
+        },
```
mtitov, line 11: could be removed (and a corresponding value from `schemas`), since we do local runs anyway

src/radical/pilot/configs/resource_tacc.json
```diff
@@ -1,5 +1,42 @@
...
+        "agent_scheduler"             : "CONTINUOUS",
+        "agent_spawner"               : "POPEN",
+        "launch_methods"              : {
+                                         "order": ["SRUN"],
+                                         "SRUN" : {
+                                             "pre_exec_cached": []
+                                         }
+                                        },
+        "pre_bootstrap_1"             :["export RP_APP_TUNNEL_ADDR=144.76.72.175:27017",
+                                        "echo  $RP_APP_TUNNEL_ADDR"],
```
mtitov, line 30: is this address valid?
but anyway, this could be removed as well, tunneling is not needed here

src/radical/pilot/configs/resource_tacc.json
```diff
@@ -1,5 +1,42 @@
...
+        "resource_manager"            : "SLURM",
+        "cores_per_node"              : 56,
+        "gpus_per_node"               : 0,
+        "agent_config"                : "default",
+        "agent_scheduler"             : "CONTINUOUS",
+        "agent_spawner"               : "POPEN",
+        "launch_methods"              : {
+                                         "order": ["SRUN"],
+                                         "SRUN" : {
+                                             "pre_exec_cached": []
```
mtitov, line 26: do we need to add corresponding modules load here? (as for `frontera`)

AymenFJA: loading modules like the one we have in the `non-interactive` resource config ones lead to an error in MPI binding, as it overrides the preloaded module by the interactive job.

mtitov: I would think yes and no - (1) I see that user could be completely responsible which modules should be loaded, since it is an interactive mode and user is able to control the environment; (2) RP is responsible to set a correct environment (load modules) to run a corresponding LM. I would follow the 2nd case, and extra modules, which are required for user, should go into task's `pre_exec`

Let's also see what's @andre-merzky 's opinion on this

AymenFJA: Good points, thanks Mikhail, happy to accommodate any.

andre-merzky: I don't even understand the problem :-P  Why would the environment of the client interact with the environment of the LM?  Or, in other words, why does it matter if the client lives on a login node or a compute node?  After the bootstrapper run, the agent env should be basically identical (assuming the bootstrapper loads the same modules)...

andre-merzky:
@AymenFJA : additional to the points made by @mtitov , could you please also add resource configs for summit (both for jsrun and prrte), please?  That would useful in the context of radical-collaboration/hpc-workflows/issues/154 .

src/radical/pilot/configs/resource_tacc.json
AymenFJA: fixed

src/radical/pilot/configs/resource_tacc.json
AymenFJA: fixed

andre-merzky:
> @AymenFJA : additional to the points made by @mtitov , could you please also add resource configs for summit (both for jsrun and prrte), please? That would useful in the context of [radical-collaboration/hpc-workflows/issues/154](https://github.com/radical-collaboration/hpc-workflows/issues/154) .

It's also fine to do that in a separate PR, no need to hold up this one...

src/radical/pilot/configs/resource_ornl.json
```diff
@@ -181,6 +181,56 @@
...
+                                         "module load   libzmq/4.3.3",
+                                         "module load   python/3.7-anaconda3",
+                                         "ulimit -u 65536"
+                                        ],
+        "valid_roots"                 : ["$MEMBERWORK/"],
+        "default_remote_workdir"      : "$MEMBERWORK/${LSB_PROJECT_NAME,,}",
+        "python_dist"                 : "default",
+        "virtenv_dist"                : "default",
+        "virtenv_mode"                : "local",
+        "rp_version"                  : "local",
```
mtitov, line 225: just keep only one parameter: `"virtenv_mode"                : "local"` (other parameters will be set by RP)

mtitov:
@andre-merzky in case of Summit, we need to think how to determine the value of SMT. Currently RP will assume that SMT is equal `1`, but if user did set `smt4` and will try to use all hwthreads (168) per node, RP would think that more resources requested than provided (during the scheduling step?)
https://github.com/radical-cybertools/radical.pilot/blob/b6d3f04ac8332b14f7604e561ddf5bbabfc8d9b8/src/radical/pilot/pmgr/launching/base.py#L766-L770
Should Aymen make a test run with `smt4` set and couple tasks each with 64 threads?

andre-merzky:
Did we check the pilot env to see if the SMT setting is (directly or indirectly) available somewhere?  If not, can we let the RM run a small test on a compute node to count the number of available cores maybe? (`jsrun -n 1 /bin/sh -c "cat /proc/cpuinfo  | grep -c 'core id'"`)?

mtitov:
@andre-merzky looks like with some hacks it is possible to get that info:

> run a small test on a compute node to count the number of available cores maybe? (jsrun -n 1 /bin/sh -c "cat /proc/cpuinfo | grep -c 'core id'")?

Was able to get info with `jsrun -n 1 /bin/sh -c "cat /proc/cpuinfo | grep -c 'processor'"`, BUT it counts available cores plus blocked/system_reserved cores, e.g., for `smt4` it gives `176`, but for `smt2` it gives `92` (`(42 * 2) + (2 * 4)`)

Found a proposed solution, but still would need to do some parsing
```
# smt4
$ jsrun -n1 -bpacked:1 /bin/bash -c 'echo $OMP_PLACES'
{0:4}

# smt2
$ jsrun -n1 -bpacked:1 /bin/bash -c 'echo $OMP_PLACES'
{0:2}

# smt1
$ jsrun -n1 -bpacked:1 /bin/bash -c 'echo $OMP_PLACES'
{0}
```

Thus during the bootstrap we can execute the command above (+some parsing) and put that info into some environment variable (`RADICAL_SMT`? and keep the same env variable in SAGA instead of the [current one](https://github.com/radical-cybertools/radical.saga/blob/devel/src/radical/saga/adaptors/lsf/lsfjob.py#L255))

and then use that in PMGRLaunching as `smt = os.environ.get('RADICAL_SMT') or system_architecture.get('smt', 1)`

andre-merzky:
> Found a proposed solution, but still would need to do some parsing

Nice!  I mean, it is somewhat clumsy what we try to do here - but I think that node inspection is something we will need to think about anyway (for example for heterogeneous nodes), and it's good to have some precedence to play with.

> but still would need to do some parsing

```sh
$ for X in "{0:4}" "{0:2}" "{0}"; do echo -n "$X : "; expr "$X" : '.*\(.\)}' | tr 0 1; done
{0:4} : 4
{0:2} : 2
{0} : 1
```


> UPDATE: I was wrong with speaking about "during the bootstrap" - SMT should be in sync between the client (PMGRLaunching) and the agent (ResourceManager) parts, and for interactive mode, env variable above (RADICAL_SMT) should be set ahead before client starts (?). And we want to have SMT-determing command within the config? Just thinking out loud..

Yeah, it won't help SAGA really, let's leave that as it is for now...

src/radical/pilot/pmgr/launching/base.py
```diff
@@ -286,6 +286,13 @@ def work(self, pilots):
         for pilot in pilots:
             resource = pilot['description']['resource']
             schema   = pilot['description']['access_schema']
+            if schema == 'interactive' and 'ornl' in resource:
+                smt_cmd = r""
+                smt_cmd += r"expr $(jsrun -n1 -bpacked:1 /bin/bash -c 'echo $OMP_PLACES') :"
+                smt_cmd += r" '.*\(.\)}' | tr 0 1;"
+                smt = ru.sh_callout(smt_cmd, shell=True)[0].strip()
+                os.environ["RADICAL_SMT"] = smt
```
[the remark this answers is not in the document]
andre-merzky: Sorry for late response!  I would prefer if we do not need to rely on the user to provide the export.  `RADICAL_SMT` is great if it exists, but otherwise I agree we should either inspect the system for the setting (as we discussed in some place) or fall back to the configs.  A prominent log message (`logger.info('used SMT setting: %d' % smt`) would be useful.

src/radical/pilot/session.py
```diff
@@ -714,6 +714,8 @@ def _get_resource_sandbox(self, pilot):
                     for k,v in pilot['description'].items():
                         if v is None:
                             v = ''
+                        if k == 'project' and '_' in v and 'ornl' in resource:
+                            v = v.split('_')[0]
```
andre-merzky, line 718: What is the purpose of this change?

src/radical/pilot/pmgr/launching/base.py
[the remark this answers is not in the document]
mtitov: Andre, but such code with `jsrun` command we shouldn't have in PMGRLaunching, right? It should get such info from the config (that what I would propose to have -> `cfg.pmgr_pre_launch`) and then execute it, thus it will set necessary env variable

src/radical/pilot/session.py
mtitov: this part was moved from outer scope, since it is related to PD attributes it is under `if '%' in sandbox_raw:` statement now, and it doesn't interfere with such case as having `_` in env variable name (`$MEMBERWORK/$LSB_PROJECT_NAME`)

src/radical/pilot/pmgr/launching/base.py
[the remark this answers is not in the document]
andre-merzky: @mtitov and @andre-merzky discussed this again.  We decided that both on pmgr and on agent side we check `RADICAL_SMT`, and if not set fall back to the config value.  It is the user's responsibility to ensure either.

This also implies that `RADICAL_SMT` should be consistently set on the SAGA level, including in the SHELL adaptor.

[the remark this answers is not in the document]
andre-merzky: @mtitov : I pushed a commit for the above, can you please check if that matches our discussion?
This should actually work w/o changes to the SAGA adaptors as `RADICAL_SMT` ends up in the pilot environment anyway and the bootstrapper picks it up from there.  But supporting in in the shell adaptor is cleaner and makes sense either way...

src/radical/pilot/pmgr/launching/base.py
```diff
@@ -987,7 +980,7 @@ def _prepare_pilot(self, resource, rcfg, pilot, expand, tar_name):
         jd_dict.queue                 = queue
         jd_dict.candidate_hosts       = candidate_hosts
         jd_dict.file_transfer         = list()
-        jd_dict.environment           = dict()
+        jd_dict.environment           = {'RADICAL_SMT': smt}
```
mtitov, line 983: can we move adding this env into line 991? - to keep env setup together

```python
jd_dict.environment['RADICAL_BASE'] = resource_sandbox
jd_dict.environment['RADICAL_SMT']  = smt
```

docs/source/faq.rst
```diff
@@ -183,4 +183,40 @@ Number of concurrent RADICAL-Pilot scripts that can can be executed
...
+run an RP application in an interactive job allocation.  For example, the
+`ornl.summit` resource configuration defines an `interactive` access schema
+which can be used for that purpose.  A respective pilot description would look
+like this:
+
+    import radical.pilot as rp
+
+    pd = rp.PilotDescription()
+    pd.resource = 'ornl.summit'
+    pd.schema   = 'interactive'
```
mtitov, line 202: With this PR we have a separate config for such case - `ornl.summit_interactive`.. there are two limitations to have it within `ornl.summit` - because of SAGA's Shell adaptor, which doesn't allow the following attributes ([link](https://github.com/radical-cybertools/radical.saga/blob/671dd9563228cad58580017963e573f83b91cf54/src/radical/saga/adaptors/shell/shell_job.py#L158-L175))
(1) use of `project` for `default_remote_workdir` (thus, for interactive mode it is `$MEMBERWORK/${LSB_PROJECT_NAME,,}`)
(2) `system_architecture`

Kinda, if we want to have a new schema in the current config for Summit, then these attributes could be added just to pass validation for SAGA Shell adaptor.

src/radical/pilot/pmgr/launching/base.py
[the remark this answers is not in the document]
mtitov: added couple comments on that updates (mostly related to the provided example for Summit, and it is not the case for Frontera)

src/radical/pilot/configs/resource_ornl.json
```diff
@@ -179,6 +179,53 @@
...
+                                         "module load   libzmq/4.3.3",
+                                         "module load   python/3.7-anaconda3"
+                                        ],
+        "pre_bootstrap_1"             : ["module load   gcc/9.1.0",
+                                         "module load   libzmq/4.3.3",
+                                         "module load   python/3.7-anaconda3",
+                                         "ulimit -u 65536"
+                                        ],
+        "valid_roots"                 : ["$MEMBERWORK/"],
+        "default_remote_workdir"      : "$MEMBERWORK/${LSB_PROJECT_NAME,,}",
```
andre-merzky, line 219: I don't understand the `${NAME,,}` syntax - what am I missing about the commas?  The `ru.expand_env` uses a single `:` to delimit default values - I don't think that method is used here though...

src/radical/pilot/pmgr/launching/base.py
andre-merzky: Done.

docs/source/faq.rst
andre-merzky: Thanks,   I think this works for the time being.

src/radical/pilot/configs/resource_ornl.json
mtitov: that env variable contains a project name, but it is with capital letters, such format `${env_name,,}` gives a value in lower case (to be used for path)

docs/source/faq.rst
mtitov: should example be updates accordingly?
```
    pd.resource = 'ornl.summit_interactive'
    pd.schema   = 'interactive'
```
and/or
```
    pd.resource = 'tacc.frontera'
    pd.schema   = 'interactive'
```

examples/config.json
```diff
@@ -110,6 +110,14 @@
         "cores"    : 56
     },
 
+    "tacc.frontera_interactive":{
```
mtitov, line 113: @AymenFJA we don't have such config anymore

src/radical/pilot/pmgr/launching/base.py
```diff
@@ -764,7 +771,7 @@ def _prepare_pilot(self, resource, rcfg, pilot, expand, tar_name):
                 cleanup = cleanup.replace('v', '')
 
         # estimate requested resources
-        smt = system_architecture.get('smt', 1)
+        smt = os.environ.get('RADICAL_SMT') or system_architecture.get('smt', 1)
```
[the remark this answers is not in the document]
mtitov: @AymenFJA ping

docs/source/faq.rst
mtitov: updated description

andre-merzky (approved):
LGTM - thanks all!

PR #2576, opened 2022-04-13
Update Spock endpoints in resource config

mtitov:
Endpoint full name should be set, so SAGA would be able to distinguish it

https://github.com/radical-cybertools/radical.saga/blob/ab9b9b703934385d939c8dbc4822128f131d23fc/src/radical/saga/adaptors/slurm/slurm_job.py#L665


andre-merzky (approved):
LGTM, thanks

Issue #2577, opened 2022-04-14
Cannot build docs


andre-merzky:
The doc build process seems to use the wrong python:
```sh
$ sphinx-build -b html -d build/doctrees docs/source/ build/html
Running Sphinx v4.5.0
loading translations [python]... not available for built-in messages
WARNING: html_static_path entry '_static' does not exist
building [mo]: targets for 0 po files that are out of date
building [html]: targets for 17 source files that are out of date
updating environment: [new config] 17 added, 0 changed, 0 removed
/mnt/home/merzky/radical/radical.pilot.work/docs/source/3/3.1-getting-started.ipynb:2: WARNING: No smart quotes defined for language "python".                                              

Notebook error:
FileNotFoundError in 3/3.11-using-pre-and-post-exec-commands.ipynb:
[Errno 2] No such file or directory: '/mnt/home/merzky/radical/radical.pilot.work/ve3/bin/python3.8': '/mnt/home/merzky/radical/radical.pilot.work/ve3/bin/python3.8'

$ which python3
/mnt/home/merzky/radical/radical.pilot.work/ve3/bin/python3

$ python3 -V
Python 3.6.9
```
I don't see where it is taking the `3.8` from.   I used the following steps to set up the environment:
```sh
$ virtualenv -p python3 ve3
$ . ve3/bin/activate
$ pip install -r docs/source/requirements_docs.txt 
$ pip install .
$ ipython kernel install --name "python3"
# failed

$ pip install ipython
$ pip install ipykernel
$ ipython kernel install --name "python3"
# works, but why does it install stuff into /usr/share???  I don't want that.
# Is that needed?

$ jupyter kernelspec list
$ dotenv set RADICAL_PILOT_DBURL mongodb://localhost:27017/rpdb
$ sphinx-build -b html -d build/doctrees docs/source/ build/html
```

eirrgang:
The particular problem described looks like it probably depends in part on the user's environment, but some additional packages are definitely needed in the requirements file. At a minimum, requirements_docs.txt needs the following:

```
dill
ipython
ipykernel
nbsphinx
python-dotenv[cli]
sphinx_copybutton
sphinx_rtd_theme
```

I don't find that I need to manually install the python3 ipython kernel, either.

```
virtualenv -p python3 ve3
. ve3/bin/activate
pip install -r docs/source/requirements_docs.txt
pip install .
dotenv set RADICAL_PILOT_DBURL mongodb://<redacted>
[...]

PR #2579, opened 2022-04-14
Updates for Sphinx builds.

eirrgang:
* Update package requirements.
* Adjust conf.py to remove warnings.

Ref #2577

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -81,7 +81,7 @@ master_doc = 'index'
 #
 # This is also used if you do content translation via gettext catalogs.
 # Usually you set "language" from the command line for these cases.
-language = 'python'
+language = 'en'
 
 # There are two options for replacing |today|: either, you set today to some
 # non-false value, then it is used:
@@ -173,7 +173,7 @@ html_theme_path = ["_themes", ]
 # Add any paths that contain custom static files (such as style sheets) here,
 # relative to this directory. They are copied after the builtin static files,
 # so a file named "default.css" will overwrite the builtin "default.css".
-html_static_path = ['_static']
+# html_static_path = ['_static']
 
 # Add any extra paths that contain custom files (such as robots.txt or
 # .htaccess) here, relative to this directory. These files are copied
@@ -470,5 +470,3 @@ html_sourcelink_suffix = ''
 # Output prompt for code cells. %s is replaced by the execution count. To get a
 # prompt similar to the Classic Notebook, use
 # nbsphinx_output_prompt = 'Out[%s]:'
-
-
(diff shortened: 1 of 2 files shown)


docs/source/conf.py
```diff
@@ -81,7 +81,7 @@
 #
 # This is also used if you do content translation via gettext catalogs.
 # Usually you set "language" from the command line for these cases.
-language = 'python'
```
eirrgang, line 84: Maybe this was intended for `highlight_language`. However, the default value of `highlight_language` should be sufficient and, if not, `python3` would be a better choice that `python`.

kartikmodi: Yes, `python3` would be better. There is also `nbsphinx_kernel_name` worth noting

eirrgang: Okay. Well, I think adding a `highligh_language` setting may be beyond the scope of the present PR, but is certainly unrelated to fixing the `language` setting.

andre-merzky:
@kartikmodi : Kartik, can you please review this PR?  Thanks!

docs/source/requirements_docs.txt
```diff
@@ -1,6 +1,8 @@
+dill
+ipython
+ipykernel
 nbsphinx
+python-dotenv[cli]
 sphinx_copybutton
 sphinx_rtd_theme
-python-dotenv[cli]
```
kartikmodi, line 4: The dotenv is used for loading env variables in notebooks. We have  not found any other way to bulk load env vars in NB

docs/source/requirements_docs.txt
```diff
@@ -1,6 +1,8 @@
...
-ipykernel
```
kartikmodi, line 5: This is required for notebooks to actually run while building on readthedocs

eirrgang: I alphabetized for maintainability. The original commit just adds `dill`, `ipython`, and `ipykernel`.

mtitov: hi Eric, actually `dill` can be dropped as an explicit requirement, since RP is in requirements and `dill` is in its dependencies (since 1.14 - https://github.com/radical-cybertools/radical.pilot/blob/devel/setup.py#L270)

eirrgang: Done.

I just tried `make html` in a fresh environment (after setting the DBURL) and it got as far as `reading sources... [ 21%] 3/3.2-obtaining-task-details `. It's been there for a couple of minutes. I haven't bothered to check what it's doing, but I'm assuming it's got at least far enough to confirm the Python package set.

Does it have to do actual RP work to build docs every time? Does nbsphinx have a way to mock imports, the way autodoc can?

kartikmodi (approved):

mtitov (approved):
Thanks!

PR #2587, opened 2022-04-21
Misspelled doc name.

eirrgang:
An incorrectly identified RST file caused a documentation section to be missing from the Sphinx document tree.


andre-merzky (approved):
Thank you!

PR #2596, opened 2022-04-29
Raptor worker pre_exec

AymenFJA:
This PR adds the option to specify `pre_exec` to the worker description when it gets submitted via a Raptor master.


andre-merzky (approved):
LGTM

mtitov:
should be updated this part as well
https://github.com/radical-cybertools/radical.pilot/blob/4c3a2a9301c7c7df4d79b2c993d65bc767c048ab/src/radical/pilot/raptor/master.py#L302
```td = dict(descr)```
otherwise any added info will not be reflected in Master

PR #2599, opened 2022-05-03
anaconda support for `prepare_env`

andre-merzky:
Hi @AymenFJA - this PR should have full support for anaconda under `pilot.prepare_env`.  Would you mind giving it a try and to review?  Thanks!


src/radical/pilot/raptor/worker_mpi.py
```diff
@@ -554,47 +612,68 @@ def _dispatch_function(self, task, env):
...
+        comm = task.get('mpi_comm')
+        if comm:
+            # we have an MPI communicator we need to inject into the function's
+            # arguments.
+            if py_func:
+                # For a `py_func` we add the communicator as `comm` kwarg if
+                # that is set to None, and otherwise as first `arg` if that is
+                # None.  If neither is true we'll error out.
+                # NOTE that we don't change the number of arguments either way.
+                if 'comm' in kwargs and kwargs['comm'] is None:
```
andre-merzky, line 656: TODO AM: make sure this still works

andre-merzky:
From @mtitov : prefer short name over long names (conda vs. anaconda, venv vs. virtualenv)

src/radical/pilot/raptor/worker_mpi.py
AymenFJA: This does work, it was a silly mistake from my side, sorry for the noise.

src/radical/pilot/raptor/worker_mpi.py
```diff
@@ -554,47 +612,68 @@ def _dispatch_function(self, task, env):
...
         kwargs  = task['description'].get('kwargs', {})
+        py_func = False
+
+        self._log.debug('=== orig args: %s : %s', args, kwargs)
 
         # check if we have a serialized object
+        self._log.debug('func serialized: %d: %s', len(func), func)
         try:
-
+            # FIXME: can we have a better test than try/except?  This hides
```
AymenFJA, line 625: I totally agree with this example: if we have an issue on the deserialization level, it's totally hidden, and won't know the type of error because we do not have logs from the sterilizer side. So it is an issue (I faced it recently).

AymenFJA (approved):

examples/misc/raptor_master.py
```diff
@@ -224,12 +229,15 @@ def request_cb(self, tasks):
     #
     def result_cb(self, tasks):
 
+        raise RuntimeError('oops')
```
AymenFJA, line 232: @andre-merzky  Why are we raising here? Also, it raises the "unreachable code" error in pylint/flake8

src/radical/pilot/raptor/worker_mpi.py
andre-merzky: Could you let `get_func_attr` return a reasonable error for example?

andre-merzky: Anyway, I think tht is out of scope for this specific PR, so will merge now...

Issue #2602, opened 2022-05-05
setting environment variable fails in RAPTOR worker_mpi


AymenFJA:
```python
Traceback (most recent call last):
  File "/home/aymen/radical.pilot.sandbox/parsl.radical.session.radical.3.aymen.019117.0007/pilot.0000/master.000000/./raptor_worker.py", line 96, in run
    out, err, ret, val, exc = self._dispatch(task)
  File "/home/aymen/ve/raptor/lib/python3.9/site-packages/radical/pilot/raptor/worker_mpi.py", line 533, in _dispatch
    return self._dispatch_non_mpi(task)
  File "/home/aymen/ve/raptor/lib/python3.9/site-packages/radical/pilot/raptor/worker_mpi.py", line 593, in _dispatch_non_mpi
    return dispatcher(task)
  File "/home/aymen/ve/raptor/lib/python3.9/site-packages/radical/pilot/raptor/worker_mpi.py", line 675, in _dispatch_function
    os.environ[k] = v
  File "/home/aymen/ve/raptor/lib/python3.9/os.py", line 684, in __setitem__
    value = self.encodevalue(value)
  File "/home/aymen/ve/raptor/lib/python3.9/os.py", line 756, in encode
    raise TypeError("str expected, not %s" % type(value).__name__)
TypeError: str expected, not int
```
```
                 'environment': {'RP_GTOD': '/home/aymen/radical.pilot.sandbox/parsl.radical.session.radical.3.aymen.019117.0007//pilot.0000//gtod',
                                 'RP_PILOT_ID': 'pilot.0000',
                                 'RP_PILOT_SANDBOX': '/home/aymen/radical.pilot.sandbox/parsl.radical.session.radical.3.aymen.019117.0007//pilot.0000/',
                                 'RP_PROF': '/home/aymen/radical.pilot.sandbox/parsl.radical.session.radical.3.aymen.019117.0007//pilot.0000//prof',
                                 'RP_PROF_TGT': '/home/aymen/radical.pilot.sandbox/parsl.radical.session.radical.3.aymen.019117.0007//pilot.0000//master.000000/master.000000.worker.0000.prof',
                                 'RP_RANK': 0,
                                 'RP_RANKS': 1,
                                 'RP_RESOURCE': 'local.localhost',
                                 'RP_RESOURCE_SANDBOX': '/home/aymen/radical.pilot.sandbox',
[...]

PR #2603, opened 2022-05-05
fix raptor env vals

AymenFJA:
This will fix issue #2602

### DIFF
--- a/src/radical/pilot/raptor/worker_mpi.py
+++ b/src/radical/pilot/raptor/worker_mpi.py
@@ -519,8 +519,8 @@ class _Worker(mt.Thread):
                'RP_GTOD'            : os.environ['RP_GTOD'],
                'RP_PROF'            : os.environ['RP_PROF'],
                'RP_PROF_TGT'        : os.environ['RP_PROF_TGT'],
-               'RP_RANKS'           : 1,  # dispatch_mpi will oveerwrite this
-               'RP_RANK'            : 0,  # dispatch_mpi will oveerwrite this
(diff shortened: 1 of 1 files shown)


src/radical/pilot/raptor/worker_mpi.py
```diff
@@ -519,8 +519,8 @@ def _dispatch(self, task):
                'RP_GTOD'            : os.environ['RP_GTOD'],
                'RP_PROF'            : os.environ['RP_PROF'],
                'RP_PROF_TGT'        : os.environ['RP_PROF_TGT'],
-               'RP_RANKS'           : 1,  # dispatch_mpi will oveerwrite this
-               'RP_RANK'            : 0,  # dispatch_mpi will oveerwrite this
+               'RP_RANKS'           : str(1),  # dispatch_mpi will oveerwrite this
+               'RP_RANK'            : str(0),  # dispatch_mpi will oveerwrite this
                })
```
andre-merzky, line 524: Another option would be to allow other types as env values and to explicitly convert them when setting the env variables.  That would provide some additional safety against user level errors.  What do you think?

AymenFJA: @andre-merzky : so something like this: 

```python
        for k, v in task['description'].get('environment', {}).items():
            os.environ[k] = str(v)
```

andre-merzky: Yes, indeed.

AymenFJA:
@andre-merzky Done, comment addressed in the last commit.

andre-merzky (approved):
Thanks @AymenFJA , that LGTM!

PR #2607, opened 2022-05-06
Status callback for SAGA Launcher

mtitov:
TBD: `_job_state_cb` is not triggered when SAGA jobs changes its state - @andre-merzky can you please have a look


andre-merzky:
Ack, the missing state cb is the problem.  RP only works because of the implicit timeout for cancel during termination...

andre-merzky (approved):
LGTM.

PR #2613, opened 2022-05-09
fix which conda call

AymenFJA:
Trying to deploy RP on `xsede expanse` ended up with this error: 
```shell
./bootstrap_2.sh: line 14: syntax error near unexpected token `('
./bootstrap_2.sh: line 14: `if test "default" = "anaconda" && ! test -z conda ()'
```

Calling `which conda` actually returns the body of the function instead of executing it here https://github.com/radical-cybertools/radical.pilot/blob/14c663f927b2908cf317954d9e0ab01195a812b9/src/radical/pilot/agent/bootstrap_0.sh#L1807: 
```shell
conda ()
{
    \local cmd="${1-__missing__}";
    case "$cmd" in

        activate | deactivate)
            __conda_activate "$@"
        ;;
        install | update | upgrade | remove | uninstall)
            __conda_exe "$@" || \return;
            __conda_reactivate
        ;;
        *)
            __conda_exe "$@"
        ;;
    esac
}
```

### DIFF
--- a/src/radical/pilot/agent/bootstrap_0.sh
+++ b/src/radical/pilot/agent/bootstrap_0.sh
@@ -1804,7 +1804,7 @@ export PATH="$PB1_PATH"
 export LD_LIBRARY_PATH="$PB1_LDLB"
 
 # activate virtenv
-if test "$PYTHON_DIST" = "anaconda" && ! test -z $(which conda)
+if test "$PYTHON_DIST" = "anaconda" && ! test -z "\$(which conda)"
 then
     eval "\$(conda shell.posix hook)"
     conda activate $VIRTENV


mtitov (approved):
LGTM!

src/radical/pilot/agent/bootstrap_0.sh
```diff
@@ -1804,7 +1804,7 @@ export PATH="$PB1_PATH"
 export LD_LIBRARY_PATH="$PB1_LDLB"
 
 # activate virtenv
-if test "$PYTHON_DIST" = "anaconda" && ! test -z $(which conda)
+if test "$PYTHON_DIST" = "anaconda" && ! test -z "\$(which conda)"
```
andre-merzky, line 1807: I think the slash before `$` is incorrect - this will always be true:
```sh
$ which ls && echo "yes" || echo "no"
/bin/ls
yes
$ test -z $(which ls) && echo "yes" || echo "no"
no

$ which foo && echo "yes" || echo "no"
no
$ test -z $(which foo) && echo "yes" || echo "no"
yes
$ test -z "$(which foo)" && echo "yes" || echo "no"
yes
$ test -z "\$(which foo)" && echo "yes" || echo "no"
no
```

andre-merzky (changes requested):

src/radical/pilot/agent/bootstrap_0.sh
mtitov: @andre-merzky this part is what will be written into `bootstrap_2.sh`, thus to keep `$` and not execute it before to write into the script we put `\` -> `bootstrap_2.sh` will have `... test -z "$(which conda)"`

andre-merzky (approved):

PR #2615, opened 2022-05-11
Update error handling and logging.

eirrgang:
Apply more granular logging and error handling with respect to
PythonTask serialized functions in TaskDescriptions.


src/radical/pilot/pytask.py
```diff
@@ -45,10 +45,15 @@ def get_func_attr(bson_obj):
...
 
-        pytask = deserialize_bson(bson_obj)
-        func   = deserialize_obj(pytask['func'])
+        try:
+            pytask = deserialize_bson(bson_obj)
+            func   = deserialize_obj(pytask['func'])
+            args   = list(pytask['args'])
+            kwargs = pytask['kwargs']
+        except Exception as e:
+            raise ValueError('bson_obj is not a PythonTask') from e
```
andre-merzky, line 54: Aymen is looking into this already.  Since this is a performance critical code path I would like to avoid the additional try/except/raise, but you are also right, better error handling is needed also.  See later comment below.

src/radical/pilot/utils/serializer.py
```diff
@@ -129,17 +129,25 @@ def deserialize_file(fname):
...
+        if no result is found.
+
     '''
     result = None
 
     try:
         result = dill.loads(obj)
     except Exception as e:
-        raise Exception ("failed to deserialize from object") from e
+        raise TypeError("failed to deserialize from object") from e
```
andre-merzky, line 147: Do we really know that this is caused by an type error?
```py
>>> dill.loads(b'x')
  File "/mnt/home/merzky/radical/radical.pilot.devel/ve3/lib/python3.6/site-packages/dill/_dill.py", line 327, in loads
    return load(file, ignore, **kwds)
  File "/mnt/home/merzky/radical/radical.pilot.devel/ve3/lib/python3.6/site-packages/dill/_dill.py", line 313, in load
    return Unpickler(file, ignore=ignore, **kwds).load()
  File "/mnt/home/merzky/radical/radical.pilot.devel/ve3/lib/python3.6/site-packages/dill/_dill.py", line 525, in load
    obj = StockUnpickler.load(self)
_pickle.UnpicklingError: invalid load key, 'x'.
```

src/radical/pilot/utils/serializer.py
```diff
@@ -129,17 +129,25 @@ def deserialize_file(fname):
...
 
     if result is None:
```
andre-merzky, line 149: @AymenFJA : can this actually happen, that `dill.loads` will not raise any exception but still returns no data?

andre-merzky (commented):
Left some comments...

src/radical/pilot/utils/serializer.py
AymenFJA: No, `dill loads/load` will always raise an error/exception if there is data corruption or unserialized object fed, etc... We do not need to reraise since the exception will fall normally.

src/radical/pilot/pytask.py
AymenFJA: In principle, this should not raise any exceptions since the exception is expensive if it was raised [(python 3.9 docs),] and since the serializer will do that for us (i.e. raise an exception) then the `try/except` is not needed indeed. If it happens then the sterilizer (dill) will raise an exception that can fall naturally to the invoker.

src/radical/pilot/utils/serializer.py
AymenFJA: same as above.

src/radical/pilot/utils/serializer.py
andre-merzky: Right - then let's take this part out.

src/radical/pilot/raptor/worker_mpi.py
```diff
@@ -491,15 +491,15 @@ def run(self):
                         task['exit_code']    = -1
                         task['return_value'] = None
                         task['exception']    = [e.__class__.__name__, str(e)]
-                        self._log.exception('recv err  %s  to  0' % (task['uid']))
+                        self._log.exception('recv err  %s  to  0' % (task['uid']), exc_info=e)
```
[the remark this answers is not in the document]
eirrgang: Oops. I was going back and forth between 1.14 and devel and didn't notice that `exception` was being used here instead of error. 👍

src/radical/pilot/pytask.py
eirrgang: I suppose the question is whether you want to distinguish between errors from dill versus errors in the expectations on the object deserialized. It looked to me like I couldn't assume that KeyError would definitely _not_ come from `deserialize_bson` or `deserialize_obj`.

src/radical/pilot/raptor/worker_mpi.py
```diff
@@ -620,13 +620,21 @@ def _dispatch_function(self, task):
...
-            # FIXME: can we have a better test than try/except?  This hides
-            #        potential errors...
-            # FIXME: ensure we did not get args and kwargs from above
-            to_call, args, kwargs = PythonTask.get_func_attr(func)
+            to_call, _args, _kwargs = PythonTask.get_func_attr(func)
+            assert callable(to_call)
             py_func = True
-        except:
-            pass
+        except (ValueError, TypeError) as e:
```
eirrgang, line 626: It is extremely important for debugging to know why `get_func_attr` was not successful. It is not clear to me under what circumstances, if any, you want `get_func_attr` to throw versus returning `None`, and what the raised exceptions should represent.

It is much less necessary to distinguish the reason for failure in this particular function, though, if a pickled function were the _last_ option checked for instead of the first. I would think this would also be the most expensive step. Shouldn't this be moved to the end? Then, there is no need to distinghish one type of failure from another. We can just log the exception, and raise the `ValueError` for "callable not found".

src/radical/pilot/utils/serializer.py
eirrgang: TypeError or ValueError both seem reasonably appropriate, since either could mean that Python objects with incompatible interfaces were asked to interact.

In this case, it is a recoverable condition, so RuntimeError seems a bit strong. Also, RuntimeError can be triggered by other parts of this use case, and I really don't want to make it harder to catch and discriminate errors specifically related to the deserialization attempt.

Here, we already see that dill may raise exceptions that don't inherit from `dill.PickleError`, and I don't think we can assume that they will otherwise derive from `pickle.PickleError`.

eirrgang (commented):
I would propose to refresh this with
* fewer try/except frames
* the pythontask check moved to the end, so that it is a final failure condition of the dispatcher

If there is a parallel effort in progress, though, please direct me to an issue or PR to follow, and I can cancel this one instead.

andre-merzky:
@AymenFJA : can you please follow up on this PR if time permits?

eirrgang:
> @AymenFJA : can you please follow up on this PR if time permits?

Would you like me to post an update?

I see that @andre-merzky is suggesting that the handling for `None` deserialization result should be removed.

Other questions surrounding performance trade-offs require some clarification of which sorts of errors need to be distinguishable, and whether distinguishing them should require knowledge of implementation details, such as exceptions from particular serializer libraries.

Could I at least ask for feedback on whether there is any reason not to move the pythontask check to the end, so that it is a final failure condition of the dispatcher?

src/radical/pilot/utils/serializer.py
eirrgang: I'm looking at this thread and I'm not sure @andre-merzky and @AymenFJA are talking about the same line. @andre-merzky  is asking whether it is necessary to check `if result is None`. I can't tell whether @AymenFJA is saying that `dill.loads` will not return `None`.

Can you please clarify whether you agree that the `if result is None` conditional should be removed?

andre-merzky:
@AymenFJA : any update by any chance?

src/radical/pilot/raptor/worker_mpi.py
eirrgang: done

andre-merzky (approved):
LGTM - thanks all

PR #2619, opened 2022-05-20
Remove node names duplication in SRun LM command

mtitov:
Eliminates the following error:
```
srun: error: Required nodelist includes more nodes than permitted by max-node count (2 > 1). 
Eliminating nodes from the nodelist.
```
example of the command that caused the error message above: `srun --nodes 1 --ntasks 2 --cpus-per-task 16 --nodelist=crusher129,crusher129 $RP_TASK_SANDBOX/task.000000.exec.sh`

p.s. tasks still were executed successfully

### DIFF
--- a/src/radical/pilot/agent/launch_method/srun.py
+++ b/src/radical/pilot/agent/launch_method/srun.py
@@ -131,8 +131,8 @@ class Srun(LaunchMethod):
         else:
             # the scheduler did place tasks - we can't honor the core and gpu
             # mapping (see above), but we at least honor the nodelist.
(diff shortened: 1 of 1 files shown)


andre-merzky (approved):
I am a bit surprised that `rank['node_name']` needs to be cast to a string (shouldn't this always be a string, given that it's a *_name_*?) - but apart from that this LGTM.  Thanks.

mtitov:
@andre-merzky agree that it always should be a string, I would think it is just "re-assurance" of a value type (and I kept it from the updated part of the code, which was `mapping += ' --nodelist=%s' % ','.join(str(n) for n in nodelist)`)

PR #2632, opened 2022-06-26
Checking input staging exists before tar-ing Fixes #2483

### LINKED ISSUE #2483: Confirm file existence before data staging in
Should RP ensure that files (and directories) exist before trying to stage them?
(*) with focus on staging for `pilot` and not for `task`

example of logging record about failed file
```
1642537009.288 : pmgr_launching.0000  : 29972 : 140520324400896 : DEBUG    : err: tar: pilot.0000/ascent_actions.yaml: File removed before we read it
```

kartikmodi:
The behaviour - 
This fix will stop ALL pilots on agent end, if ANY of the pilot has input staging path missing.
I understand that is the expectation, and NOT we skip that particular pilot and  ERROR log. 

Also, currently checking at pilot level only, not at task level

### DIFF
--- a/src/radical/pilot/pmgr/launching/base.py
+++ b/src/radical/pilot/pmgr/launching/base.py
@@ -280,7 +280,7 @@ class PMGRLaunchingComponent(rpu.Component):
         self.advance(pilots, rps.PMGR_LAUNCHING, publish=True, push=False)
 
         # We can only use bulk submission for pilots which go to the same
-        # target, thus we sort them into buckets and lunch the buckets
+        # target, thus we sort them into buckets and launch the buckets
         # individually
         buckets = defaultdict(lambda: defaultdict(list))
         for pilot in pilots:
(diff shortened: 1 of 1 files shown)


src/radical/pilot/pmgr/launching/base.py
```diff
@@ -421,6 +421,13 @@ def _start_pilot_bulk(self, resource, schema, pilots):
```
kartikmodi, line 421: I did not used - 
1. pathlib - because I did not find in nearby code, probably we don't use this 
2. sh_callout - I can use this if you say so, but didn't find a good reason to do so. 
3. ru - I did not find any particular method for checking file exists in utils 
@andre-merzky

src/radical/pilot/pmgr/launching/base.py
```diff
@@ -421,6 +421,13 @@ def _start_pilot_bulk(self, resource, schema, pilots):
 
             for fname in ru.as_list(pilot['description'].get('input_staging')):
                 base = os.path.basename(fname)
+                # checking if input staging file exists
+
+                assert os.path.exists(fname) == False
```
kartikmodi, line 426: If am am sure if I should use assert here, or a check like just below, raising a RuntimeError 
Please let me know, so I'll remove one of them.

src/radical/pilot/pmgr/launching/base.py
andre-merzky: `os.path` is perfectly acceptable for local files.

src/radical/pilot/pmgr/launching/base.py
andre-merzky: An exception will leave more detailed traces in the logs and will make it easier to understand what went wrong.

andre-merzky (changes requested):
The assert can be removed - otherwise looks good to me.
Thanks! :-)
Can you think of a way to add a unit test for this?  That might not be easy to do as we don't test the bulk lunching, yet.  Let's discuss this on Monday's call...

PR #2634, opened 2022-06-27
ensure pip in venv mode

### LINKED ISSUE #2622: radical-pilot-create-static-ve should use explicit executable paths
In RP 1.14, the radical-pilot-create-static-ve script sets `VIRTENV_CMD="$PYTHON -m venv"`, then creates and activates a virtual environment.

Before guaranteeing the existence of `pip` in the virtual environment, though, the script calls `pip install --no-cache-dir --upgrade setuptools pip wheel| progress || exit 1`.

If `python3 -m venv` does not guarantee the installation of `pip` for whichever python + venv package are resolved, the `pip` executable may correspond to a different Python installation.

This causes two problems.

* The packages may or may not be upgraded/installed to a usable location (depending on behavior that may be installation-specific or version-specific, and which may change in future versions of `pip`).
* The Python installation seen by the build system for `$MODULES` may not be consistent with that of the virtual environment.

Suggestion:
* Use `"$PREFIX"/bin/python -m pip` or `"$PREFIX"/bin/pip3` instead of a bare `pip`
* Ensure that the venv has access to pip using a mechanism appropriate to the virtual environment type. (E.g. `conda install pip`, `python -m ensurepip`, or other)
* Make sure to fail if `pip` is not available.

andre-merzky:
This fixes #2622

### DIFF
--- a/bin/radical-pilot-create-static-ve
+++ b/bin/radical-pilot-create-static-ve
@@ -191,6 +191,9 @@ then
 
         . "$PREFIX"/bin/activate
 
+        # ensure we have pip
+        python3 -m ensurepip
+
         echo -n "update  setuptools pip wheel"
         pip install --no-cache-dir --upgrade setuptools pip wheel || exit 1
 
--- a/src/radical/pilot/agent/agent_0.py
+++ b/src/radical/pilot/agent/agent_0.py
@@ -791,8 +791,8 @@ class Agent_0(rpu.Worker):
 
         self._log.debug('env_spec: %s', env_spec)
 
-        etype = env_spec['type']
-        evers = env_spec['version']
+        etype = env_spec.get('type', 'venv')
+        evers = env_spec.get('version', '')
         path  = env_spec.get('path')
         emods = env_spec.get('setup')    or []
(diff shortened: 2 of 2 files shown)


bin/radical-pilot-create-static-ve
```diff
@@ -191,6 +191,9 @@ then
 
         . "$PREFIX"/bin/activate
 
+        # ensure we have pip
+        python3 -m ensurepip
+
         echo -n "update  setuptools pip wheel"
         pip install --no-cache-dir --upgrade setuptools pip wheel || exit 1
```
eirrgang, line 198: If `ensurepip` installs pip, are you sure it will be on the path? And will it be available as `pip` and not (only) `pip3`? It seems like it would be best to also update this line.
```suggestion
        "$PREFIX"/bin/pip install --no-cache-dir --upgrade setuptools pip wheel || exit 1
```

andre-merzky: Yes, `ensurepip` will put it in `bin/` which is on the path in the activated env.

PR #2638, opened 2022-07-06
Remove redis_link attributes from raptor example

AymenFJA:
Removing `redis_link` from a devel example (redis_link attribute is a use case specific)


andre-merzky (approved):
LGTM

PR #2641, opened 2022-07-06
Updated mpi-info extraction

mtitov:


andre-merzky (approved):
Another day, another MPI version string... :-P

PR #2644, opened 2022-07-07
fix rp serializer unittest

### LINKED ISSUE #2642: Failed unit test for `serializer`
Running tests on Ruby@LLNL, RPv.1.14

```
=================================== FAILURES ===================================
_________________________ TestSerializer.test_ser_path _________________________
self = <tests.unit_tests.test_serializer.TestSerializer testMethod=test_ser_path>
    def test_ser_path(self):
    
        def hello_test():
            return 1
    
        obj = hello_test
        ser_obj_file_path = serializer.serialize_file(obj)
        expected_path = '/tmp/rp_obj.pkl'
    
>       self.assertEqual(ser_obj_file_path, expected_path)
E       AssertionError: '/var/tmp/titov1/rp_obj.pkl' != '/tmp/rp_obj.pkl'
E       - /var/tmp/titov1/rp_obj.pkl
E       + /tmp/rp_obj.pkl
tests/unit_tests/test_serializer.py:23: AssertionError
```

AymenFJA:
This will fix #2642


mtitov:
Sorry, but since we are on this test, need to change it more. The point of every test is to check that a corresponding function/method performs expected actions and generates an expected outcome.

This path of a temporary file is not generated within tested function, thus either (1) move corresponding [lines](https://github.com/radical-cybertools/radical.pilot/blob/38ca2f0603649550b021b364f9b9b0e95e491852/src/radical/pilot/utils/serializer.py#L22-L24) into the function; or (2) just get that path from the module (`from radical.pilot.utils import _obj_file_path`, and in this case a name of variable should be more representative)

What is more important to test in this function - (1) file with such name didn't exist before func execution, but (2) after execution, file was created.

p.s. the goal of tests is to cover every line of code - that's kinda our direction :)

mtitov (approved):

PR #2648, opened 2022-07-15
Updated exec profile events

mtitov:
Reworked events related to the execution layer
(resolved an issue with `exec_start/stop` duplication)


andre-merzky (approved):
Thanks for resolving the issue!

This reminds me though (unrelated to this PR) that we should provide backward-compatibility with the event names used before the env isolation came in - I opened #2650 for this.

PR #2654, opened 2022-07-26
bulk submission for parsl executor

andre-merzky:
This PR improves submission throughput by collecting task bulks on the PARSL layer before pushing to RP.  The max task rate I see is still not great though, about 20/sec. (improved from 5/sec though).


andre-merzky:
We follow up with that once Aymen's experiments are done

andre-merzky:
TODO: make bulk size configurable

AymenFJA:
To be compatible with other use cases that require `task stream submission mode` like `Colmena`, `bulk mode` is now optional.

src/radical/pilot/agent/executing/parsl_rp.py
```diff
@@ -351,16 +352,9 @@ def submit(self, func, *args, **kwargs):
...
-        # FIXME: use uid generator
-        task_id = 'task.parsl.%06d' % self._task_counter
-
-      # # ----------------------------------------------------------------------
-      # # test code: this is the fastest possible executor implementation
-      # self.future_tasks[task_id] = Future()
-      # self.future_tasks[task_id].set_result(3)
-      # return self.future_tasks[task_id]
-      # # ----------------------------------------------------------------------
+        self._task_counter = ru.generate_id('task.%(item_counter)06d',
```
andre-merzky, line 355: Why do we need the `self._task_counter variable?

examples/parsl/map_reduce.py
```diff
@@ -12,6 +12,7 @@
 config = Config(
          executors=[RADICALExecutor(
                         label        = 'RADICALExecutor',
+                        bulk_mode    = False,
```
andre-merzky, line 15: Should we turn on the bulk mode for this example, or is there any drawback?

andre-merzky:
The optional bulk mode makes sense IMHO.

src/radical/pilot/agent/executing/parsl_rp.py
AymenFJA: For two reasons: 

1. Maintain consistency in terms of task numbers between RP and Parsl (we create a future that the DFK needs to look for when fetching the task info if it's finished/pending/running).
2. I followed the class code template provided by `HTEX` for `RP-Parsl` executor which used `self._task_counter`.

examples/parsl/map_reduce.py
AymenFJA: @andre-merzky sure, that makes sense.

src/radical/pilot/agent/executing/parsl_rp.py
AymenFJA: Ohh, I got it and makes sense. I addressed your proposal in the latest commit.

examples/parsl/map_reduce.py
AymenFJA: Addressed.

andre-merzky:
I can't approve the PR as I opened it, but that looks good to merge from my perspective.

mtitov (approved):
LGTM!

AymenFJA:
Thank you all. I will merge now!

PR #2656, opened 2022-08-03
Fixed `task['resources']` value

mtitov:
@andre-merzky for raptor-worker-tasks, should the code below be updated to have `cpu` and `gpu` in `resources` as for regular tasks?

https://github.com/radical-cybertools/radical.pilot/blob/8847925be5dd53accdb8d3a5c62a118d53f2db07/src/radical/pilot/raptor/worker_default.py#L444-L445


andre-merzky (approved):
LGTM - thanks!

andre-merzky:
> @andre-merzky for raptor-worker-tasks, should the code below be updated to have `cpu` and `gpu` in `resources` as for regular tasks?
> 
> https://github.com/radical-cybertools/radical.pilot/blob/8847925be5dd53accdb8d3a5c62a118d53f2db07/src/radical/pilot/raptor/worker_default.py#L444-L445

Yes, indeed they should.

andre-merzky:
Dammit - the `resources` files for the task dict means something different here - it's more like the `slots` we use for the cont scheduler.  I pushed a fix for that to this branch but will add proper resources in another branch (working on fixes for the profile events in raptor right now, the `resources` can be added there).

PR #2657, opened 2022-08-03
Fixed `cpu_threads` for special tasks (service, sub-agent)

mtitov:
Resource Manager gets "expanded" value for cores per node

https://github.com/radical-cybertools/radical.pilot/blob/48d9414af93822d1799b404a92bbe6e28c024e87/src/radical/pilot/pmgr/launching/base.py#L767-L770
https://github.com/radical-cybertools/radical.pilot/blob/48d9414af93822d1799b404a92bbe6e28c024e87/src/radical/pilot/pmgr/launching/base.py#L868
https://github.com/radical-cybertools/radical.pilot/blob/48d9414af93822d1799b404a92bbe6e28c024e87/src/radical/pilot/agent/resource_manager/base.py#L215-L223


andre-merzky (approved):
Changes make sense - but by now we don't seem to use `threads_per_core` anywhere in the code anymore.  Should we remove it?

mtitov:
> but by now we don't seem to use threads_per_core anywhere in the code anymore. Should we remove it?

we kinda do - in RM LSF we use it to confirm `cores_per_node` (could be considered as excessive)
https://github.com/radical-cybertools/radical.pilot/blob/48d9414af93822d1799b404a92bbe6e28c024e87/src/radical/pilot/agent/resource_manager/lsf.py#L34-L35
https://github.com/radical-cybertools/radical.pilot/blob/48d9414af93822d1799b404a92bbe6e28c024e87/src/radical/pilot/agent/resource_manager/lsf.py#L53-L57

But in general, I'd like to keep it - since it could be used in LM APRun and JSRun, where threads per core could be provided explicitly.

p.s. And maybe at some point we would want to update a notation (?) and to have `cores_per_node` and `threads_per_core` as well as `cpus_per_node`, which will be equal to `cores_per_node x threads_per_core` - just thinking aloud :)

PR #2662, opened 2022-08-10
better raptor profiling 

andre-merzky:
- move raptor profiles and logfiles into sandboxes
 - consistent use of task modes
 - derive etypes from task modes

rp session rivendell merzky 019213 0012 util2


src/radical/pilot/agent/scheduler/base.py
```diff
@@ -450,11 +453,14 @@ def _change_slot_states(self, slots, new_state):
...
+    def slot_status(self, msg=None, uid=None):
         '''
         Returns a multi-line string corresponding to the status of the node list
         '''
 
+        if not _debug:
+            return
+
         if not self._log.isEnabledFor(logging.DEBUG):
             return
```
mtitov, line 465: we shouldn't have such internal flag `_debug`, and we do have a check for debug mode (`if not self._log.isEnabledFor(logging.DEBUG):`)

src/radical/pilot/agent/scheduler/base.py
```diff
@@ -450,11 +453,14 @@ def _change_slot_states(self, slots, new_state):
     #
     # NOTE: any scheduler implementation which uses a different nodelist
     #       structure MUST overload this method.
-    def slot_status(self, msg=None):
+    def slot_status(self, msg=None, uid=None):
```
mtitov, line 456: `uid` should be a part of `msg` - no need to add extra inputs

src/radical/pilot/pmgr/launching/base.py
```diff
@@ -801,6 +801,11 @@ def _prepare_pilot(self, resource, rcfg, pilot, expand, tar_name):
 
             requested_nodes = math.ceil(requested_nodes)
 
+        # now that we know the number of nodes to request, calculate back the
+        # actual number of requested cores and gpus
+        requested_cores = requested_nodes * cores_per_node
+        requested_gpus  = requested_nodes * gpus_per_node
```
mtitov, line 807: should be removed, since duplicates lines above (which also uses `avail_(cores|gpus)_per_node`)

src/radical/pilot/raptor/master.py
```diff
@@ -163,11 +166,9 @@ def _get_config(self, cfg=None):
         cfg['log_lvl'] = 'warn'
         cfg['kind']    = 'master'
         cfg['base']    = pwd
-        cfg['sid']     = os.environ['RP_SESSION_ID']
-        cfg['base']    = os.environ['RP_PILOT_SANDBOX']
-
-        # FIXME: use registry for comm EP info exchange, not cfg files
-      # cfg['path'] = os.environ['RP_TASK_SANDBOX']
+        cfg['sid']     = self._sid
+        cfg['base']    = self._sbox
```
mtitov, line 170: then line above `cfg['base']    = pwd` should be removed?
and line above that should be updated
```python
ru.dict_merge(cfg, ru.read_json('%s/control_pubsub.json' % self._psbox))
```
and `pwd` is not needed

src/radical/pilot/raptor/master.py
```diff
@@ -303,6 +304,7 @@ def submit_workers(self, descr: Dict[str, Union[str, int]],
                 ru.write_json(cfg, fname)
 
                 td = dict()
```
mtitov, line 306: this line could be updated as well (as we discussed earlier): `td = dict(descr)`
and only several lines below would be needed
```python
                td['mode']             = RAPTOR_WORKER
                td['cpu_process_type'] = rpc.MPI
                td['cpu_thread_type']  = rpc.POSIX

                td['executable'] = 'python3'
                td['arguments']  = [...
```

src/radical/pilot/task.py
```diff
@@ -185,6 +185,7 @@ def as_dict(self):
             'tmgr':             self.tmgr.uid,
             'uid':              self.uid,
             'name':             self.name,
+            'mode':             self.mode,
```
mtitov, line 188: why do we need duplicate it? (it is in `description`)

src/radical/pilot/utils/prof_utils.py
```diff
@@ -496,8 +497,9 @@ def get_session_profile(sid, src=None):
 
     if os.path.exists(src):
         # we have profiles locally
-        profiles  = glob.glob("%s/*.prof"   % src)
-        profiles += glob.glob("%s/*/*.prof" % src)
+        profiles  = glob.glob("%s/*.prof"     % src)
+        profiles += glob.glob("%s/*/*.prof"   % src)
+        profiles += glob.glob("%s/*/*/*.prof" % src)
```
mtitov, line 502: would it be ok not to limit the number of levels to check?
`profiles  = glob.glob("%s/**/*.prof"     % src,  recursive=True)`

mtitov (commented):
several follow up comments

mtitov:
There are couple files slipped in, which shouldn't be in repo
```
src/radical/pilot/agent/scheduler/.base.py.swp
src/radical/pilot/raptor/.worker_mpi.py.swp
```

src/radical/pilot/agent/scheduler/base.py
andre-merzky: `slot_status` is called with high frequency, and I wanted to avoid (somewhat costly) string expansion when the string might not be needed.  Also avoids code clutter to do string expansion in a single place.

src/radical/pilot/agent/scheduler/base.py
```diff
@@ -450,11 +453,14 @@ def _change_slot_states(self, slots, new_state):
...
         '''
         Returns a multi-line string corresponding to the status of the node list
         '''
 
+        if not _debug:
+            return
+
```
andre-merzky, line 463: Yeah, we should use something like `debug_10` or so.  Let me try if I get that to work.  For now though this is rather useful :-)

src/radical/pilot/agent/scheduler/base.py
andre-merzky: adding to the above reply: even if we have `debug` enabled, the slot status is usually too costly to include, thus the separate flag.

src/radical/pilot/pmgr/launching/base.py
andre-merzky: I see - well, it should be indented then I guess, because w/o that code the pilot will report incorrect amount of resources and analytics will be thrown off (percentages don't work out).

src/radical/pilot/raptor/master.py
andre-merzky: Ah, of course, I removed the duplication.  Ack on the merge path, fixed.

src/radical/pilot/task.py
andre-merzky: indeed, removed

src/radical/pilot/utils/prof_utils.py
andre-merzky: I didn't know `glob` supports that - thanks! :-)
Having said that, I would like to be carful with that.  if `rp_install` exists in the pilot sandbox, then the above can become costly.  
I am not too happy with the above either though - this now fetches more than it should (which takes time) - but otherwise we miss out on the raptor profiles which live in the master sandboxes. With the termination nowadays more relying on a hard kill, the bootstrapper usually has no chance to tar up the sandboxes and the client can't pull that tarball anymore.  We probably should initiate that tar from the client side as a separate saga job - but that would be out of scope for this PR.
I'll open an issue for this - #2663

src/radical/pilot/raptor/master.py
andre-merzky: Probably right, but out of scope for this PR?

src/radical/pilot/agent/scheduler/base.py
andre-merzky: This has been addressed - the `debug_level` of the logger is checked now.

src/radical/pilot/pmgr/launching/base.py
andre-merzky: The code now makes a difference between `requested_cores/gpus` and `allocated_cores/gpus`, and the latter is what is reported as pilot resources for analytics.

src/radical/pilot/utils/prof_utils.py
andre-merzky: I applied the `**` syntax now - but also, the fetching was fixed in `rpu.session` lately.

src/radical/pilot/agent/scheduler/base.py
andre-merzky: Done

andre-merzky:
@AymenFJA , @mtitov : this is ready for another review now.

andre-merzky:
> There are couple files slipped in, which shouldn't be in repo
> 
> ```
> src/radical/pilot/agent/scheduler/.base.py.swp
> src/radical/pilot/raptor/.worker_mpi.py.swp
> ```

Thanks, removed.

PR #2666, opened 2022-08-16
Fix/issue 2590

andre-merzky:
This builds upon @kartikmodi 's fixes for abs pathes and makes the staging behavior consistent for task and pilot staging.  An example is added to show several combinations for the staging directives and the various supported schemas.

PS.: apologies for adding `client.py` in this PR - that should have been a separate PR.  Please ignore that file, it is not part of the core code base...


src/radical/pilot/staging_directives.py
```diff
@@ -141,17 +141,21 @@ def complete_url(path, context, log=None):
...
 
     The `context` parameter is expected to be a dict which provides a set of
-    URLs to be used to expand the path.
+    URLs to be used to expand the path for those schemas
 
-    Other URL schemas are left alone, any other strings are interpreted as
+    Other URL schemas are left alone, any non-URL strings are interpreted as
     path in the context of `pwd`.
 
     The method returns an instance of ru.Url.  Note that URL parsing is not
```
eirrgang, line 161:
```suggestion
    The method returns an instance of :py:class:`radical.utils.Url`.  Note that URL parsing is not
```

src/radical/pilot/staging_directives.py
```diff
@@ -141,17 +141,21 @@ def complete_url(path, context, log=None):
     certain locations, namely relative to
 
         * `client://`  : the client's working directory
-        * `resource://`: the RP    sandbox on the target resource
-        * `pilot://`   : the pilot sandbox on the target resource
-        * `task://`    : the task  sandbox on the target resource
+        * `endpoint://`: the root file system on the target resource
+        * `resource://`: the sandbox base dir on the target resource
```
eirrgang, line 145: `file:///` and `pwd:///` are not described here.

src/radical/pilot/staging_directives.py
```diff
@@ -141,17 +141,21 @@ def complete_url(path, context, log=None):
...
     For the above schemas, we interpret `schema://` the same as `schema:///`,
     ie. we treat this as a namespace, not as location qualified by a hostname.
 
     The `context` parameter is expected to be a dict which provides a set of
-    URLs to be used to expand the path.
+    URLs to be used to expand the path for those schemas
 
-    Other URL schemas are left alone, any other strings are interpreted as
+    Other URL schemas are left alone, any non-URL strings are interpreted as
     path in the context of `pwd`.
```
eirrgang, line 159: This is incomplete, and it is not clear what `pwd` refers to. I suggest making a change similar to the following, depending on additional documentation expansion.
```suggestion
    Other URL schemes are left alone. Any non-URL strings are interpreted as
    paths. Absolute paths are interpreted as ``file:///path``. Relative paths are
    interpreted as ``pwd:///path``.
```

eirrgang (commented):
The docstrings in this file would be useful for https://github.com/radical-cybertools/radical.pilot/blob/devel/docs/source/datastaging.rst#staging-directives

If it is inappropriate to explicitly document the functions, then some of their important details could be moved to a module docstring, consolidating the text at https://github.com/radical-cybertools/radical.pilot/blob/devel/docs/source/datastaging.rst#staging-directives and replacing that section of documentation with a `.. automodule::`.

If that is beyond the scope of the present pull request, I think it is still within the scope of the issue. I can offer to help with a PR, if desired.

src/radical/pilot/staging_directives.py
andre-merzky: Thanks, applied.

src/radical/pilot/staging_directives.py
andre-merzky: This has been applied - thanks.

src/radical/pilot/staging_directives.py
andre-merzky: The semantics for those two schemas is clarified now.

PR #2669, opened 2022-08-23
add rcfg inspection utilities

andre-merzky:
@eirrgang : Eric, does that do the trick for you?

```py
#!/usr/bin/env python3

import radical.pilot as rp

cfgs = rp.utils.get_resource_configs()
sites = cfgs.keys()
for site in sites:
    print('\n --- %s' % site)
    resource_names = cfgs[site].keys()
    for name in resource_names:
        print('  -- %s' % name)
        label = site + '.' + name
        rcfg = rp.utils.get_resource_config(label)
        print('   - %s' % 'default')
        print('     %s' % rp.utils.get_resource_fs_url(label))
        print('     %s' % rp.utils.get_resource_job_url(label))
        for schema in rcfg['schemas']:
            print('   - %s' % schema)
            print('     %s' % rp.utils.get_resource_fs_url(label, schema))
            print('     %s' % rp.utils.get_resource_job_url(label, schema))
```

```
$ ./test.py
[...]
 --- xsede
  -- expanse
   - default
     file://expanse.sdsc.xsede.org
     slurm://expanse.sdsc.xsede.org
   - local
     file://expanse.sdsc.xsede.org
     slurm://expanse.sdsc.xsede.org
   - ssh
     sftp://expanse.sdsc.xsede.org
     slurm+ssh://expanse.sdsc.xsede.org
[...]
```


src/radical/pilot/utils/misc.py
```diff
@@ -79,5 +84,152 @@ def write_dir(path):
...
+               resource_names = cfgs[site].keys()
+    '''
+
+    global _rcfgs
+
+    if not _rcfgs:
+        _rcfgs = ru.Config('radical.pilot.resource', name='*', expand=False)
+
+    # return a deep copy
+    return ru.Config(cfg=_rcfgs)
```
mtitov, line 112: I would propose to use `ru.Config(from_dict=_rcfgs)`, since config is already formed, thus no need to go through all configs again

src/radical/pilot/utils/misc.py
```diff
@@ -79,5 +84,152 @@ def write_dir(path):
...
+    if not _rcfgs:
+        get_resource_configs()
+
+    site, host = resource.split('.', 1)
+
+    if site not in _rcfgs:
+        return None
+
+    if host not in _rcfgs[site]:
+        return None
```
mtitov, line 183: maybe reuse the method above?
```python
rcfg = get_resource_config(resource)
if rcfg is not None:
    ...
    return ru.Url(rcfg[schema].filesystem_endpoint)
```

src/radical/pilot/utils/misc.py
```diff
@@ -79,5 +84,152 @@ def write_dir(path):
...
```
mtitov, line 225: same as for `get_resource_fs_url`

src/radical/pilot/utils/misc.py
andre-merzky: Done, thanks

src/radical/pilot/utils/misc.py
andre-merzky: of course - done

eirrgang:
> @eirrgang : Eric, does that do the trick for you?

Oh! I'm sorry, I never saw the notification for this.

Should this be considered more stable than the `ru.Config` constructor, then?

What would be a good section to insert documentation in https://radicalpilot.readthedocs.io/en/stable/?

eirrgang:
> What would be a good section to insert documentation in https://radicalpilot.readthedocs.io/en/stable/?

Maybe I should infer from the deep nesting into `rp.utils.misc` that you prefer not to draw attention to these tools?

andre-merzky:
> Should this be considered more stable than the `ru.Config` constructor, then?

It gives the same, really, as that is what is being used here: [code](https://github.com/radical-cybertools/radical.pilot/blob/devel/src/radical/pilot/utils/misc.py#L111)

> What would be a good section to insert documentation in https://radicalpilot.readthedocs.io/en/stable/?
> Maybe I should infer from the deep nesting into rp.utils.misc that you prefer not to draw attention to these tools?

No, that is not really intended t hide it away, just the way we organize code.  We don't have any use cases to integrate that tighter into the API (the `rp.Session` has the related `session.get_resource_config(resource, schema=None)` call, but this PR is specifically catered to the request to inspect the configs outside of an `rp.Session`).  

I guess section 7 would be a good point to mention this, maybe in 7.4?

PR #2670, opened 2022-08-24
Fixed session cache

mtitov:
Issue related to the method `rp.utils.db_utils.get_session_docs` - if `cachedir` is not provided, then `None` directory is created (within the directory where RP/RE is launched) with session json-data


andre-merzky (approved):
Thanks for tracking this down - that was annoying! :-)

PR #2678, opened 2022-08-31
Sync term signals

mtitov:


src/radical/pilot/agent/scheduler/base.py
```diff
@@ -255,7 +255,7 @@ def initialize(self):
         self.register_subscriber(rpc.AGENT_UNSCHEDULE_PUBSUB, self.unschedule_cb)
 
         # start a process to host the actual scheduling algorithm
-        self._p = mp.Process(target=self._schedule_tasks)
+        self._p = mp.Process(target=self._schedule_tasks, args=(self._term,))
```
andre-merzky, line 258: I don't think we need to pass `self._term` - `self` is passed implicitly (as the target is a member function) and that caries the `self._term` with it?

andre-merzky (commented):
See comment.  Thanks!

src/radical/pilot/agent/scheduler/base.py
mtitov: yeah, this one is to pass term-signal explicitly, and thought to follow that `_subscribers` and `_publishers` are set within this `_schedule_tasks` method. Reassure environment setup for child process (?)

andre-merzky: I think this is more confusing than helpful, as we then again overwrite `self._term` in the child and need to track that on debugging.

mtitov: ok, reassignment removed

andre-merzky (approved):
Thanks!

PR #2679, opened 2022-08-31
Removed duplication call for `stop` method (`Agent_0`)

mtitov:
When command `cancel_pilot` is received (from client side), it triggers command `terminate` for agent's components (including `Agent_0`), which is received by method `rpu.Component._cancel_monitor_cb` and which calls method `rpu.Component.stop`


mtitov:
Double call of `ResourceManager.stop()` (`self._rm.stop()` inside `Agent_0.finalize()`) is not confirmed

andre-merzky (approved):
LGTM

PR #2680, opened 2022-08-31
pilot profiles should be transferred as tarball Fixes #2663

### LINKED ISSUE #2663: pilot profiles should be transferred as tarball
See https://github.com/radical-cybertools/radical.pilot/pull/2662/files#r942613889

kartikmodi:


src/radical/pilot/utils/session.py
```diff
@@ -114,32 +113,44 @@ def fetch_profiles (sid, dburl=None, src=None, tgt=None, access=None,
...
-                        prof_file.close()
+                else:
+                    log.warn("profiles tarball doesnt exists! creating now.")
+                    # so lets create a tarball right now with SAGA JobService
+                    js_url = pilot['js_hop']
+                    log.debug('js  : %s', js_url)
+                    js = rs.job.Service(js_url, session=session)
+
+                    # find <pilot_sandbox> -name \*.prof > profiles.lst; tar jcf profiles.tbz -T profiles.lst
+                    cmd = "find %s -name \\*.prof  -exec tar -czf %s/%s {} +" % (sandbox.url.path, sandbox.url.path,
```
kartikmodi, line 126: I tried using profiles.lst method, but was facing absolute path problems then. Later figured out it was because of something else and profiles.lst path itself. 
I can change the logic to > profiles.lst path way. NP

kartikmodi:
There is a gap between tarball fetching way and the individual profiles way. 
The individual profiles way is not getting `task.xxxxx.prof` files which are nested under task.xxx dir. 

Should I try to make the tarball also work the same way ignoring these `task.xxxx.prof` ??
Or should I create a new ticket for fixing individual profiles to include these files?

src/radical/pilot/utils/session.py
andre-merzky: The command is likely going to fail (or to yield unexpected results such as incomplete tarballs) for large numbers of files: the total length of the command line `find` can build up this way is limited, usually to something like 256 characters. 

Also, you may want to use bzip2 compression (`tar cjf` in instead of `tar czf`) as that compresses much better than `zip` for our profiles.  Or at least I think it does, you may want to check maybe...

src/radical/pilot/utils/session.py
```diff
@@ -91,7 +91,6 @@ def fetch_profiles (sid, dburl=None, src=None, tgt=None, access=None,
     log.debug("Number of pilots in session: %d", num_pilots)
 
     for pilot in pilots:
-
```
andre-merzky, line 94: It is a minor point maybe, by please don't change code outside of the scope you work on.  Please also have a look at our [coding guidelines](https://github.com/radical-cybertools/radical.pilot/wiki/Coding-Guidelines) for max line length, indentation rules etc.

src/radical/pilot/utils/session.py
kartikmodi: 1. Will change to profiles.lst way

2. I used `z` as gzip just because later in the code it was using  - 
```                    tarball = tarfile.open(ftgt.path, mode='r:gz')
```
which I thought as gzip. Also, the file name ending with `.tgz`.
Should I change to bzip2?

andre-merzky: What was the problem you faced with paths?  You could change the command line from `find $pilot_sandbox -name \*.prof ...` to `cd $pilot_sandbox; find . -name \*.prof` maybe?

kartikmodi: The path problems were my deficit understanding of these commands. I found find can run command via `-exec` option so used it, and by then the path things were fixed.
So i didn't changed to file way. 

Now, updated the command to use `profiles.lst` file and passing it to tar. 
Using `cd` before running everything, makes it clearer. 


Lastly, still using c**z**f with tarball.

src/radical/pilot/utils/session.py
kartikmodi: Read and trying to understand the python things here. I will try to enforce the PIP, pylint with flake8 thing in my dev env.

src/radical/pilot/utils/session.py
andre-merzky: > Lastly, still using czf with tarball.

how did you decide - did you measure how both versions behave? :-)

kartikmodi: I have changed to bzip2 algo. I did not any measurement. 
As mentioned above, I decided based on gzip mode `r:gz` that was expected with in extraction and naming `tgz`.

I also realized that after switching to bzip2 the tarball size was reduced to 21K from 34K (in case of gzip).

src/radical/pilot/utils/session.py
```diff
@@ -111,35 +110,47 @@ def fetch_profiles (sid, dburl=None, src=None, tgt=None, access=None,
...
+                    cmd = "cd %s; find . -name \\*.prof > profiles.lst; " \
+                          "tar cjf %s -T profiles.lst" % (sandbox.url.path, PROFILES_TARBALL)
+                    j = js.run_job(cmd)
+                    j.wait()
+                    log.debug('tarball cmd for profiles : %s', cmd)
+                    log.debug('tar result : %s\n---\n%s\n---\n%s',
+                              j.get_stdout_string(), j.get_stderr_string(),
+                              j.exit_code)
+                    if not j.exit_code:
+                        log.info("profiles tarball created successfully")
```
andre-merzky, line 135: What happens if the above failed?

src/radical/pilot/utils/session.py
```diff
@@ -111,35 +110,47 @@ def fetch_profiles (sid, dburl=None, src=None, tgt=None, access=None,
...
+
+                ftgt = rs.Url('%s/%s' % (tgt_url, PROFILES_TARBALL))
```
andre-merzky, line 137: This is probably inconsistent throughout the code, but please switch from `rs.Url` to `ru.Url`.  The `Url` class was moved to RU a while back.

src/radical/pilot/utils/session.py
```diff
@@ -111,35 +110,47 @@ def fetch_profiles (sid, dburl=None, src=None, tgt=None, access=None,
...
+                if skip_existing and os.path.isfile(ftgt.path) \
+                        and os.stat(ftgt.path).st_size > 0:
 
-                        tarball_available = True
+                    log.info("skip fetching of '%s/%s' to '%s'.",
```
andre-merzky, line 141: If we skip fetching here, then we should not need to create the remote tarball in the first place - s maybe move that test up front?

andre-merzky (commented):
Great, I think we are getting there.  Left some minor comments...

src/radical/pilot/utils/session.py
kartikmodi: what is exact expected behaviour for skip fetching.  Examaple - 
1. If local tarball exists don't get it
2. if tarball exists in sandbox don't recreate it
3. don't fetching any profiles if profiles exist locally

src/radical/pilot/utils/session.py
kartikmodi: Adding warn log exit exit code non-zero. We cannot do much here. THe tarball_available stays False, and profiles will be fetched individually.

andre-merzky: Ack, we can't do much, but a log warning sounds appropriate.

src/radical/pilot/utils/session.py
kartikmodi: Done

andre-merzky (approved):
Thanks!

PR #2681, opened 2022-09-05
task description update

andre-merzky:
This PR follows up on several discussions on consistency of the task description attributes.  Among others, we interpret `gpu_processes` as `gpus_per_rank`, and allow for conflicting values for `cpu_process_type` and `gpu_process_type`.  

This PR cleans this up and consistently uses the notion of `ranks` which is widely used in the community.   The changes are backward compatible.  Introducing them now allows us to clean up our internal code, and makes it possible to retire the old attributes on a future 2.0 API version.

I marked priority of this PR as `high`, not because this is urgent to get in, but because this PR has a high potential of conflict with upcoming work.


src/radical/pilot/task_description.py
```diff
@@ -508,16 +525,25 @@ class TaskDescription(ru.TypedDict):
...
-        GPU_PROCESS_TYPE: str         ,
-        GPU_THREADS     : int         ,
-        GPU_THREAD_TYPE : str         ,
-        LFS_PER_PROCESS : int         ,
-        MEM_PER_PROCESS : int         ,
+        RANKS           : int         ,
+        CORES_PER_RANK  : int         ,
+        GPUS_PER_RANK   : int         ,
+        THREADING_TYPE  : str         ,
+        GPU_TYPE        : str         ,
```
mtitov, line 532: @andre-merzky quick first couple comments
(1) should we still use term "threads" - `threads_per_rank`? since `cores_per_rank` may refer to either physical cores or hwthreads, kinda we do mean "virtual" cores
(2) do we actually need `threading_type`? isn't it assumed that with `>1` threads per rank we do use OpenMP? or we will use it to just enforce setting env variable `OMP_NUM_THREADS`?

andre-merzky: > (1) should we still use term "threads" - threads_per_rank? since cores_per_rank may refer to either physical cores or hwthreads, kinda we do mean "virtual" cores

Fair - but for now it's at least in sync with the pilot description where we request `cores`, not `hwthreads`.

> (2) do we actually need threading_type? isn't it assumed that with >1 threads per rank we do use OpenMP? or we will use it to just enforce setting env variable OMP_NUM_THREADS?

Also good point - maybe we don't need it (similar to how the `process_type` is gone).  We could re-introduce it once we see conflicting settings as we do for `gpu_type`.  But yes, the intent was to trigger that env setting and to keep the env clean by default...

mtitov: > Fair - but for now it's at least in sync with the pilot description where we request cores, not hwthreads.

that one is arguable as well :) but my point is to keep it less confusing, and switch from `cpu_threads` to `threads_per_rank` seems more straightforward.

Speaking about `cores` in PD - I would propose to review that as well (either in this PR or in another) - it did make sense before, but not lately with more machines having hwthreads (which also could be managed with SMT parameter). Also, for example in EntK, we use `cpus` in resource description, and such notation is also used within OpenMPI (in Summit's user guide it is used for physical cores and hwthreads). Maybe we can enforce such parameter name with "logical" (e.g., `logical_cpus`)? But now, users should be more encouraged to use `nodes`

> But yes, the intent was to trigger that env setting and to keep the env clean by default...

actually agree to keep it for this

andre-merzky: > > Fair - but for now it's at least in sync with the pilot description where we request cores, not hwthreads.
> 
> that one is arguable as well :) but my point is to keep it less confusing, and switch from `cpu_threads` to `threads_per_rank` seems more straightforward.

I disagree with that really.  The point of this PR (or one point) is to remove the notion of threads: we do spawn ranks (processes if you so will), but we do not spawn threads.  In fact, what we do is we schedule cores for that rank, and then we don't care what the rank does with the cores: create more processes, create threads, do nothing, etc.

Using `threads` introduce confusion IMHO: it's the same term as we used before but means something different.  Also, we don't use that notion anywhere in the code: not in the API, not in the scheduler (which schedules `cores`), not in the resource utilization (we utilize cores), not in the pilot description.    It looks wrong to me to request 2 pilot cores on summit and then submit tasks with 2 threads on them unless we expose SMT - but that...
 
> Speaking about `cores` in PD...

... would be out of scope for this PR IMHO.  Once that happens it might be worthwhile to reconsider the choice for tasks also: we could then support `threads_per_rank` additionally, for example.

> > But yes, the intent was to trigger that env setting and to keep the env clean by default...
> 
> actually agree to keep it for this

:+1:

mtitov: I think I'm convinced :) thanks!

andre-merzky: Thanks!

examples/misc/raptor.py
```diff
@@ -119,8 +119,8 @@ def task_state_cb(task, state):
                         'action': rp.TRANSFER})
         pilot.prepare_env(env_name='ve_raptor',
                           env_spec={'type'   : 'venv',
-                                    'version': '3.8',
-                                    'path'   : '$RP_RESOURCE_SANDBOX/ve_raptor/',
+                                    'version': '3.6',
+                                    'path'   : '/tmp/ve_raptor/',
```
AymenFJA, line 123: `/tmp/` is only accessible from within the compute node itself. It might problematic in the future if `raptor_script` for some reason needed to use the `ve_raptor` from another node,  otherwise, this looks fine for me.

AymenFJA (approved):
This looks fine to me, thanks.

examples/misc/raptor.py
andre-merzky: Correct - but since we need to specify *some* path, the example won't be easily portable out of the box anyway... :-/

andre-merzky:
@andre-merzky : add test for backward compatibility

mturilli (approved):
This is good to go for me. Documentation needs to be updated on readthedocs and I would offer a notebook-based example in that documentation.

andre-merzky:
> This is good to go for me. Documentation needs to be updated on readthedocs and I would offer a notebook-based example in that documentation.

Yes to both, but the notebook examples will enter the stream with the nb3 branch.

PR #2689, opened 2022-09-08
Fix tests cleanup

mtitov:


andre-merzky (approved):
Thanks @mtitov , much apprciated!

Issue #2218, opened 2020-09-07
Task overlay example has unknown and unpropagated error.


eirrgang:
will elaborate shortly...

Please provide all error messaages and the output of the following commands:

```
$ python -V
$ radical-stack
```

If you can reproduce the problem, it would greatly simplify debugging if you can provide a minimal code example which we can use to repoduce the problem too.

On **most** tickets we will ask you to provide additional information, most often log files. You can obtain detailed logs by rerunning your code after setting the following environment variables:

```
$ export RADICAL_VERBOSE=DEBUG
$ export RADICAL_LOG_TGT=r.log
```

This will create the logfile r.log. Additional logfiles will be located in your correct directory, in a subdirectory named after your session ID.

Log files are also created on the (remote) host you are targetting with your run. In order to fetch log files from target hosts, please run:

```
$ radical-pilot-fetch-logfile <sid>
```

where `sid` is again the session ID in question.

eirrgang:
Update: Per conversation with @andre-merzky this is two very separate issues.
1. BUG: task_overlay example indicates 2 FAILED tasks in the environment represented by the referenced dockerfile (which appears to satisfy the documented environment requirements)
2. Feature request: task_overlay.py should check for the task success and give an error if tasks have failed so that the example can be used easily in automated testing contexts.

andre-merzky:
Alas, the docker seems not as reproducible as we want: I get a different error:
```
radical.saga.exceptions.DoesNotExist: attribute key is invalid: system_architecture
```
This seems to indicate an inconsistency in the RCT stack: it should be either all devel, or all master.  RS and RU are released versions, but RP is from the `docker` branch which is in sync with devel:
```
$ radical-stack

  python               : 3.6.9
  pythonpath           : 
  virtualenv           : /home/rp/rp-venv

  radical.gtod         : 1.5.0
  radical.pilot        : 1.5.2-v1.4.0-408-g3b142cad4@docker
  radical.saga         : 1.5.2
  radical.utils        : 1.5.3
```

How would you suggest to proceed?

andre-merzky:
When I install RS devel, the stack seems functional.  The masters fail though, as you observe.  In `~/radical.pilot.sandbox/<sid>/pilot.0000/unit.000000/STDERR`, you will find:
```
[...]

PR #2691, opened 2022-09-09
Clarify and troubleshoot raptor.py example.

eirrgang:
Try to clarify how the raptor.py script should be used, what configuration is necessary, and what results to expect.

Update some symbol names, add some logging, and try to perform some additional checks.

Ref #2218


examples/misc/raptor.py
```diff
@@ -47,52 +77,58 @@ def task_state_cb(task, state):
...
+    nodes_per_worker  = cfg.nodes_per_worker
     nodes_rp    = cfg.nodes_rp
     workload    = cfg.workload
     tasks_rp    = cfg.tasks_rp
     nodes_agent = cfg.nodes_agent
 
+    cores_per_task = cfg.cores_per_task
+
     # each master uses a node, and each worker on each master uses a node
     # use 8 additional cores for non-raptor tasks
```
eirrgang, line 100: Is this two sentences or one sentence with extra words?
I'm not sure what is meant, what determines these values, and to what degree they are configurable.

examples/misc/raptor.py
```diff
@@ -1,4 +1,29 @@
...
+"""Demonstrate the "raptor" features for remote Task management.
+
+This script and its supporting files may use relative file paths. Run from the
+directory in which you found it.
+
+Refer to the ``raptor.cfg`` file in the same directory for configurable run time details.
+
+By default, this example uses the ``local.localhost`` resource with the ``local``
+access scheme. Before running, confirm that you are able to ssh to localhost
+without being prompted for a password.
```
eirrgang, line 11: Don't need to ssh with the `local` scheme.

```suggestion
access scheme.
```

andre-merzky: correct - I removed that sentence now.

examples/misc/raptor.py
andre-merzky: Thanks - I clarified the comment.

examples/misc/raptor_worker.py
```diff
@@ -23,6 +23,7 @@ def __init__(self, cfg):
 
     # --------------------------------------------------------------------------
     #
+    # TODO: This does not appear to be used.
```
andre-merzky, line 26: It is used [here](https://github.com/radical-cybertools/radical.pilot/blob/devel/examples/misc/raptor.py#L164)

examples/misc/raptor.py
```diff
@@ -1,4 +1,29 @@
...
+
+In this example, we
+* Launch one or more raptor "master" task(s), which self-submits additional tasks.
+  (TODO: How do we check/confirm their results?)
```
andre-merzky, line 15: We don't really need to as this is not a stringent test, but in principle the test would go into the master's `result_cb`.

examples/misc/raptor.py
```diff
@@ -1,4 +1,29 @@
...
+* Stage scripts to be used by a raptor "Worker" (TODO: unused?)
```
andre-merzky, line 16: They are used in the master and worker task descriptions which run the staged master and worker scripts.

examples/misc/raptor.py
```diff
@@ -1,4 +1,29 @@
...
+* Provision a Python virtual environment with :py:func:`~radical.pilot.prepare_env`
+* Submit several tasks that will be routed through the master(s) to the worker(s).
+* Submit a non-raptor task in the same Pilot environment
+
+The raptor.cfg file should be adjusted to avoid oversubscribing your resources.
+In particular,
+TODO: what are the constraints of total cores, master cores, worker
+    and task cores with respect to each other? And what are the corresponding
+    TaskDescription fields?
```
andre-merzky, line 25: Up to `cores_per_host` masters can run on a node (`masters_per_node`).  That's the only real constraint I think.  Otherwise the arithmetic in `raptor.py` should guarantee that the pilot has enough resources and that no oversubscription happens.

> And what are the corresponding TaskDescription fields?

The given values ultimately defined the number of cores used by a master task (`threads_per_process`) and number of ranks used for each worker (`cpu_processes`).

andre-merzky:
Thanks for the PR, @eirrgang !  I iterated over it, please have a look if this works for you.  I addressed some TODOs and left comments for others.

examples/misc/raptor.py
eirrgang: If someone running the example is not supposed to have any sort of run time confirmation that the script did what it was supposed to, then there should at least be some text describing what to look for.

It would also be helpful if the examples could be used as integration tests, or to validate a RP installation, like for https://github.com/radical-cybertools/radical.pilot/tree/devel/docker or to decrease the overhead of testing dependent projects like scalems. For the purposes of the current PR, I recognize that this is not currently a condoned use case for the example scripts. But these clarifications should help downstream projects to better design such tests.

examples/misc/raptor.cfg
```diff
@@ -1,48 +1,49 @@
...
+    "nodes_agent"     :   0,
 
     # pilot runtime in min
-    "runtime"        : 120,
-    "sleep"          :  30,
+    "runtime"         : 120,
+
+    # task configuration
+    "tasks_rp"        :   1,
+    "tasks_raptor"    :   1,
```
eirrgang, line 22: I'm still not clear what these are supposed to mean/do.

examples/misc/raptor.py
```diff
@@ -8,6 +37,8 @@
 
 from radical.pilot import PythonTask
 
+logger = ru.Logger('radical.pilot')
```
eirrgang, line 40: Oh! This is ultimately a wrapper for `logging.Logger`, right? Is this generally how client code is supposed to use the `logging` module when RCT is present?

Is there a way to override the absence of the `RADICAL_PILOT_LOG_LVL` environment variable programmatically?

examples/misc/raptor.py
```diff
@@ -47,52 +77,58 @@ def task_state_cb(task, state):
...
+    masters_per_node = cfg.masters_per_node
+    nodes_per_worker = cfg.nodes_per_worker
+    nodes_rp         = cfg.nodes_rp
+    tasks_raptor     = cfg.tasks_raptor
+    tasks_rp         = cfg.tasks_rp
+    nodes_agent      = cfg.nodes_agent
+
+    cores_per_task   = cfg.cores_per_task
+
+    # Each master uses a node, and each worker also uses a node.
```
eirrgang, line 99: The worker resources are allocated from the master resources, right? Or... somehow the master owns a pool of resources that it allocates to workers, doesn't it? Or does that not happen until the master actually submits workers (in which case, is it up to the user to be careful not to oversubscribe resources for workers)?

Please clarify whether the `cpu_threads` allocated to the `master` become unavailable to workers and their tasks.

Just to check: the scheduler does not trigger the Worker to pop a task unless the scheduler believes the Worker has enough idle threads to perform the task. Is that correct?

examples/misc/raptor.py
```diff
@@ -1,5 +1,34 @@
...
+'''
+Demonstrate the "raptor" features for remote Task management.
+
+This script and its supporting files may use relative file paths. Run from the
+directory in which you found it.
+
+Refer to the ``raptor.cfg`` file in the same directory for configurable run time
+details.
+
+By default, this example uses the ``local.localhost`` resource with the
```
eirrgang, line 12: In such a context, how should the reader/user interpret "nodes", with respect to `cores_per_node`, `nodes_per_worker`, etc., and the comment "Each master uses a node, and each worker also uses a node."?

examples/misc/raptor.py
eirrgang: `cores_per_host` does not appear anywhere in the project. Do you mean `cores_per_node` from the config file?

`threads_per_process` does not occur in the project. Do you mean `td.cpu_threads`?

> Otherwise the arithmetic in raptor.py should guarantee that the pilot has enough resources and that no oversubscription happens.

Do you agree with [the assessment](https://github.com/radical-cybertools/radical.pilot/issues/2218#issuecomment-1239824592) that my most recent tests for #2218 failed due to lack of resources? If so, what additional documentation or checks can you suggest in order to mitigate?

examples/misc/raptor.py
```diff
@@ -108,128 +144,152 @@ def task_state_cb(task, state):
...
 
-        # submit some test tasks
+        # Issue an RPC to provision a Python virtual environment for the later
+        # raptor tasks.  Note that we are telling prepare_env to install
+        # radical.pilot and radical.utils from sdist archives on the local
+        # filesystem. This only works for the default resource, local.localhost.
+        logger.info('Call pilot.prepare_env()')
+        pilot.prepare_env(env_name='ve_raptor',
+                          env_spec={'type' : 'venv',
+                                    'path' : '/tmp/ve_raptor/',
```
eirrgang, line 175: Is the future environment variable available for expansion?
```suggestion
                                    'path' : '$RP_RESOURCE_SANDBOX/ve_raptor/',
```
If not, can we use a `ru.Url`, like `resource:///ve_raptor/`?

eirrgang (commented):
I'll continue to work on text to address the remaining "to-do"s

examples/misc/raptor.py
eirrgang: Is the terminal output suppressed for `'radical.pilot'` even with RADICAL_PILOT_LOG_LVL=DEBUG? Should I attach a regular character stream handler from the `logging` module if I want to see these messages on the console?

examples/misc/raptor.cfg
```diff
@@ -1,48 +1,49 @@
...
-    "tasks_rp"       :    1,
+    # raptor configuration
+    "n_masters"       :   1,
+    "n_workers"       :   1,
+    "masters_per_node":   1,
+    "nodes_per_worker":   1,
 
-    "nodes_agent"    :    0,
+    # extra nodes for rp tasks and for the rp agent
+    "nodes_rp"        :   0,
```
eirrgang, line 14: nodes reserved for non-raptor tasks. should be made consistent with whether we are including non-raptor tasks

examples/misc/raptor.cfg
```diff
@@ -1,48 +1,49 @@
...
+    "nodes_agent"     :   0,
```
eirrgang, line 15: Extra nodes for overhead. Probably unused.

examples/misc/raptor.cfg
eirrgang: These are used as the range of the for loops for defining and submitting non-raptor and raptor tasks, respectively.

eirrgang (commented):
making some notes. (updates to the default values appear necessary)

examples/misc/raptor.py
andre-merzky: Alas no, not supported at the moment :-/

eirrgang: note that `path` should be optional. RP should choose something reasonable.

eirrgang:
```suggestion
```

examples/misc/raptor.cfg
eirrgang:
```suggestion
    "nodes_rp"        :   1,
```

eirrgang:
One problem was that `nodes_rp` was 0, so there weren't enough cores allocated for both the raptor and non-raptor tasks. I'm not sure why this didn't become clear at #2218. But that wasn't the only problem.

I couldn't get the master to reach state AGENT_EXECUTING. I thought it might be some hidden dependency or data flow problem, so I tried rearranging the script.

Good news! I got much farther. And now I'm basing testing on the Dockerfile that is include in the repository.

Current status:
* the `task.mpi_se_func` is failing
* I haven't checked whether the tasks that are remotely submitted (self-submitted by the master and worker) complete as expected. Those tasks should also probably be explicitly represented in the script output by inspecting some artifact returned by the master. Or at least there should be documentation about inspecting the filesystem artifacts, and we should check that things at least look like they're working.

andre-merzky (commented):
responded to some comments / TODOs

andre-merzky:
> * I haven't checked whether the tasks that are remotely submitted (self-submitted by the master and worker) complete as expected. Those tasks should also probably be explicitly represented in the script output by inspecting some artifact returned by the master. Or at least there should be documentation about inspecting the filesystem artifacts, and we should check that things at least look like they're working.

I don't think this should be turned into a unit test, so complete self-testing is overkill IMHO.  I added a comment about the results being logged in the result callback of the master which makes that inspectable at least.

examples/misc/raptor.py
andre-merzky: We generally avoid to log to stdout/stderr.  If you want to force the logger to do so, you can set `RADICAL_LOG_TGT=stdout` (see [here](https://github.com/radical-cybertools/radical.utils/blob/devel/src/radical/utils/logger.py#L192) (*))

(*) the RU documentation is not building correctly for readthedocs right now, looking into it.

andre-merzky (approved):
LGTM

examples/misc/raptor.py
eirrgang: > We generally avoid to log to stdout/stderr. If you want to force the logger to do so

I do, but I see that `Reporter` is used in other examples. Would that be more appropriate, do you think?

> the RU documentation is not building correctly for readthedocs right now, looking into it.

See also https://github.com/radical-cybertools/radical.utils/issues/359

examples/misc/raptor.py
```diff
@@ -10,23 +10,19 @@
...
 
 By default, this example uses the ``local.localhost`` resource with the
-``local`` access scheme.
+``local`` access scheme where RP oversubscribes resources to emulate multiple
+nodes.
 
 In this example, we
   - Launch one or more raptor "master" task(s), which self-submits additional
-    tasks.  (TODO: How do we check/confirm their results?)
+    tasks (results are logged in the master's `result_cb` callback).
```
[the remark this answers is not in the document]
andre-merzky: To the master's log file (`https://github.com/radical-cybertools/radical.pilot/blob/devel/examples/misc/raptor_master.py#L238`)

andre-merzky:
>> We generally avoid to log to stdout/stderr. If you want to force the logger to do so
> I do, but I see that Reporter is used in other examples. Would that be more appropriate, do you think?

Maybe.

As a general note: we know that the raptor example code is convoluted and not really instructive to a user who wants to become familiar with it.  It's living under `misc/` for that reason and is not referenced in documentations or tutorials.  We will be working on simpler and more streamlined examples which will be usable for that purpose, but will likely have to do some API-changes to simplify the usability of raptor.  I would thus like to avoid to spend too much effort in streamlining this specific code - it will likely be replaced in the future.  I don't want to discourage the PR, not at all - but also want you to be aware so that you can judge how much time and effort you intent to place here...

eirrgang:
> As a general note: we know that the raptor example code is convoluted and not really instructive to a user who wants to become familiar with it. It's living under `misc/` for that reason and is not referenced in documentations or tutorials. We will be working on simpler and more streamlined examples which will be usable for that purpose, but will likely have to do some API-changes to simplify the usability of raptor. I would thus like to avoid to spend too much effort in streamlining this specific code - it will likely be replaced in the future. I don't want to discourage the PR, not at all - but also want you to be aware so that you can judge how much time and effort you intent to place here...

Thanks. :-)

This PR serves additional purposes, though. It allows me to check my understanding of raptor details, as well as to help locate and possibilty contribute to resolving documentation short-falls. Most importantly, it is a way to try to establish a common set of conditions where we all agree what should work so that we can confirm that issue #2218 is resolved and that the new Dockerfile works. In the future, it will help to identify raptor details that have changed, but which possibly escaped documentation updates or notifications to early adopters.

I'm happy to help to try to contribute to useful documentation-by-example. Thank you for clarifying that this doesn't need to achieve a "final" form. But this PR is largely self-serving. ;-)

I'd like to give it one more pass to make sure it is clear and correct before merging.

I'd also like to confirm (at least personally) that all of the tasks ran as expected.

I'll try to look a little closer, but can you have a look at the MPI task that failed for me?
1. We should reach a point where failures and successes are equally reproducible in terms of the bundled Dockerfile and the documentation/comments.
2. Assuming you are able to reproduce the failure, I expect you can solve it faster than I can.

eirrgang:
I believe I have confirmed that the example works in a way that is verifiable by and helpful to users.

From the local repository root:

```
cd docker/
docker compose up --build -d
docker compose exec -ti -u rp login bash -c "cd && . /home/rp/rp-venv/bin/activate && cd /radical.pilot/examples/misc && python raptor.py"
docker compose exec -ti -u rp login bash -c "grep result_cb /radical.pilot/examples/misc/*/pilot.0000/master.000000.log"
docker compose down
```

Issue #2590, opened 2022-04-22
Request for clearer behavior/documentation in `stage_in` directive targets


eirrgang:
https://radicalpilot.readthedocs.io/en/stable/datastaging.html#staging-directives does not describe or link to the documentation on recognized URI schemes, but there seem to be additional unexpected semantics.

The implication in the documentation is that `target` refers to the environment of the Task for input staging directives, so stage_in.target does not work as expected with absolute paths.

In RP 1.14, I have tried to use an absolute path for the `target` on a `pilot.stage_in()` call. I use a basic string representation of a path.

When I look at `stager.0000.log` on the client side, I see that the target path has been translated to `file:///mypath` and the file was staged locally.

In the same log file, I can see that automatically generated data staging was correctly targeted to ` sftp://rp@compute/home/rp/radical.pilot.sandbox/rp.session.e28...`

I think that translating an absolute path to a local file path for a stage_in.target is inconsistent with the documentation at https://radicalpilot.readthedocs.io/en/stable/datastaging.html#staging-directives

I also think it would be very helpful to link from that section (or from https://radicalpilot.readthedocs.io/en/stable/apidoc.html#radical.pilot.Pilot.stage_in) to documentation about the URI schemes that are recognized in staging directives.

```
$ radical-stack 

  python               : /home/rp/rp-venv/bin/python3
  pythonpath           : 
  version              : 3.9.5
  virtualenv           : /home/rp/rp-venv

  radical.gtod         : 1.13.0
  radical.pilot        : 1.14.0
  radical.saga         : 1.14.0
  radical.utils        : 1.14.0

$ cat rp.session.e287de52-c240-11ec-ba50-0242ac120004/stager.0000.log 
1650634450.583 : stager.0000          : 17794 : 140495621834496 : INFO     : connect pub to state_pubsub: tcp://172.18.0.4:42893
1650634450.584 : stager.0000          : 17794 : 140495621834496 : DEBUG    : registered publisher for state_pubsub
1650634450.585 : stager.0000          : 17794 : 140495621834496 : INFO     : connect pub to control_pubsub: tcp://172.18.0.4:37679
1650634450.585 : stager.0000          : 17794 : 140495621834496 : DEBUG    : registered publisher for control_pubsub
1650634450.586 : stager.0000          : 17794 : 140495621834496 : INFO     : connect sub to control_pubsub: tcp://172.18.0.4:42625
[...]

PR #2695, opened 2022-09-13
Use a linking target for staging directives references.

### LINKED ISSUE #2441: Broken Sphinx links in datastaging.rst
datastaging.rst includes what appear to be attempts to link to the "Staging Directives" section using bare `` `staging directives` `` syntax. There is no `.. _staging directives:` label. If `sphinx.ext.autosectionlabel` were in use (it is not), I believe the link would still fail because it is not case sensitive. Nor is there a `default_role` in the conf.py.

It may be that a link was not intended, and the single-backticks were supposed to be double-backticks for monospace (code) text, but there are a number of ways to make useful links with minimal intervention.

If https://github.com/radical-cybertools/radical.pilot/pull/2440 is merged, the occurrences of `` `staging directives` `` could be replaced with `` `staging directives`_`` and ``.. _staging directives: :py:mod:`staging directives <radical.pilot.staging_directives>` `` could be added somewhere in the file. Links from other documents can simply use `` `:py:mod:`radical.pilot.staging_directives` ``

eirrgang:
Use the :py:mod:`radical.pilot.staging_directives` role for the previous `staging directives` occurrences in datastaging.rst (which did not resolve).

Fixes #2441

### DIFF
--- a/docs/source/datastaging.rst
+++ b/docs/source/datastaging.rst
@@ -26,8 +26,8 @@ Task I/O
 
 To instruct RADICAL-Pilot to handle files for you, there are two things to
 take care of. First you need to specify the respective input and output files
-for the Task in so called `staging directives`. Additionally you need
-to associate these `staging directives` to the the Task by means of
+for the Task in so called :any:`staging directives <staging_directives>`. Additionally you need
+to associate these :any:`staging directives <staging_directives>` to the the Task by means of
 the ``input_staging`` and ``output_staging`` members.
 
 What it looks like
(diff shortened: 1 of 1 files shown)


eirrgang:
Looking at more of the documentation, I'm not sure that the occurrences of \`data staging\` were intended to be links. (It may be that a lot of docs are accidentally using MarkDown syntax and merely intending pre-formatted text for such syntax.)

I also think that it could be more appropriate to move this section of text to the module docstring of `data_staging.py` and use `..automodule::` to extract it.

Then, the extensive docstring for `TaskDescription` could be reconciled with and consolidated to ensure better consistency. (And the Staging Directive section of `TaskDescription` could be replaced with a `.. seealso::` to `staging_directives`.

It is not clear to me whether `staging_directives.py` applies only to `TaskDescription`, or whether its docs should also be consolidated/reconciled with [radical.pilot.Pilot.stage_in](https://radicalpilot.readthedocs.io/en/stable/apidoc.html#radical.pilot.Pilot.stage_in). (See also #2590)

andre-merzky (approved):
LGTM, thanks!

Issue #2668, opened 2022-08-23
Offline read access to resource definition.


eirrgang:
The complete data represented by the https://radicalpilot.readthedocs.io/en/stable/apidoc.html#radical.pilot.PilotDescription.resource is not easily programmatically accessible to clients.

The resource definition can come from multiple places, has a schema that is either loosely specified or poorly documented (see also #2588), and is not easily accessed (at least without an active Session).

### Check resource accessibility

When the *resource* is being managed abstractly (symbolically) by a script, it would be helpful to be able to discover the connection method to the resource. This would allow scripted debugging of connection issues, as well as extra checks for environment suitability (e.g. checking for installed software or software versions, or even preconfiguring Python virtual environments before submitted large HPC job scripts).

### Discover filesystem endpoint

Staging large data _can_ be done through the `Pilot.stage_in`, but may result in wasted compute allocation if a large pilot job starts while data is still transferring.

Large data should be staged before the Pilot is submitted, and preferably without requiring the resources of an active Session. If the `radical.utils.Url` of the filesystem endpoint could be retrieved, more basic RCT/Saga utilities could be used to stage data before allocating HPC resources.

Even if there were a canonical / blessed way to retrieve the string representation of the filesystem endpoint while offline, client scripts could find ways to programmatically and consistently access the appropriate remote filesystem and path to stage data before or after a Session (or even from separate client-side utility scripts that manage RP inputs and artifacts). See also #2590.

## Possible fixes
* Support `radical.pilot.session.Session.get_resource_config()` as a `@classmethod`, usable without an active Session.
* Provide a simple _public_ utility to aggregate a resource definition from its various sources.
* Document the stable parts of the resource schema and provide getters.
* Provide a query tool to get arbitrary fields or fail with an appropriate exception.

See also #2588 

## Deferred

It would be great to have API for setting or updating resources known to the Session, but I intend to explicitly exclude that topic from this feature request to limit the scope of the issue.

eirrgang:
@mturilli asked me to document this feature request. We were discussing how best to separate the abstraction of execution management from data management. I observed that the "interface" between the two spaces would be clearer and cleaner if it were easier to consistently identify the same filesystem node (or endpoint) from inside and from outside a Session.

andre-merzky:
[...]

PR #2696, opened 2022-09-14
Document the supported way to fetch resource definitions.

eirrgang:
Refs #2668


src/radical/pilot/utils/misc.py
```diff
@@ -157,9 +157,12 @@ def get_resource_fs_url(resource: str,
     For the given resource label, return the contact URL of the resource's file
     system.
 
+    This URL is equivalent to the base ``endpoint:///`` URL available for use
```
eirrgang, line 160: I have not confirmed yet if this is correct. But whatever the correct mapping is, documenting it would resolve one of the use cases documented for #2668

andre-merzky: This is correct.

PR #2715, opened 2022-09-23
Split LMs for Summit among dedicated configurations 

mtitov:
Configurations for Summit
- `ornl.summit` will set MPIRun LM
- `ornl.summit_jsrun` will set JSRun LM

@andre-merzky should we set `"virtenv_mode"                : "local",` for default Summit config (`ornl.summit`)?


andre-merzky (approved):
LGTM, thanks

PR #2716, opened 2022-09-28
Update and consolidate docker documentation.

eirrgang:
* Consolidate primary documentation in docker/README.md
* Rename stack.yml to compose.yml (which is now a strong standard).
* Update for new Compose specification. Ref https://docs.docker.com/compose/compose-file/


eirrgang:
I updated some docs in addition to moving them. We should probably do a line-by-line test of the example syntax before merging.

I can add notes on pushing to / pulling from Docker Hub, if the RADICAL team has (or creates) and organizational account to reference.

eirrgang:
There are some errors. I'll push another commit shortly.

andre-merzky (approved):
LGTM - Thanks!

Issue #2700, opened 2022-09-19
obtain resource sandbox depending on pilot's project ID w/o session


andre-merzky:

eirrgang:
See, for instance, #2668.

Use case:

Identify a readable/writable path to which data can be staged before submitting the pilot, with a well-defined translation between the available abstractions inside and outside a session. E.g. paths expressed in terms of the 'filesystem_endpoint' field (or other fields in the resource definition) and paths expressed in terms of the "sandbox" properties on RP objects or the URI schemes available to staging directives.

Ultimately,
* allow scripts to be written without hard-coding for specific resources
* allow a data provider abstraction to be conceived of separately from the RP compute provider abstraction. (specify the details needed such that usable paths can be robustly defined programmatically across the RP API boundary)

eirrgang:
Presumably, this is possible in terms of `radical.pilot.utils.misc.get_resource_fs_url`.

I think an acceptable solution to this issue could be generalized beyond, specifically, the resource sandbox. The goal would be to robustly programmatically map the *same path*, *writable by the user*, between "offline" code (presumably based on `get_resource_fs_url()`) and a URI based on `endpoint:///`, `resource:///`, or `session:///` (or `pilot:///` or `task:///`) for use in a Task (such as a raptor Master, Worker, or Python function) or staging directive.

andre-merzky:
It will not be possible to predict the session, pilot and task sandboxes - they can be set by the application at runtime via the session uid and the pilot and task descriptions and are thus not predictable.  To some extend that also holds for the resource and  the fs endpoint, as those are configurable in the resource configs - but its probably prudent to assume that those don't change between invocations.  Either way, I suggest to limit the requested functionality to the fs endpoint and the resource sandbox.

eirrgang:
> It will not be possible to predict the session, pilot and task sandboxes - they can be set by the application at runtime via the session uid and the pilot and task descriptions and are thus not predictable.

It is not necessary to predict path elements that can be either queried or explicitly set. 

> To some extend that also holds for the resource and the fs endpoint, as those are configurable in the resource configs - but its probably prudent to assume that those don't change between invocations.

It is unnecessary to make such an assumption because of the tools in rp.utils.misc.

> Either way, I suggest to limit the requested functionality to the fs endpoint and the resource sandbox.

I'm not sure that I am asking for any functionality, unless it is easier to provide new functionality than to document a recipe in terms of existing functionality.

[...]

PR #2718, opened 2022-09-28
clarify endpoint URLs point to /

### LINKED ISSUE #2699: are path elements of `endpoint://` URLs always absolute paths?

andre-merzky:
This fixes #2699

### DIFF
--- a/src/radical/pilot/staging_directives.py
+++ b/src/radical/pilot/staging_directives.py
@@ -170,6 +170,9 @@ def complete_url(path   : str,
 
     NOTE: URL parsing is not really cheap, so this method should be used
     conservatively.
+
+    NOTE: For `endpoint://`, the notion of 'root' dependends of the access
+    protocol and the providing service implementation.
     '''
 
     # FIXME: consider evaluation of env vars
--- a/src/radical/pilot/staging_directives.py
+++ b/src/radical/pilot/staging_directives.py
@@ -145,7 +145,7 @@ def complete_url(path   : str,
     certain locations, namely relative to
 
         * `client://`  : the client's working directory
-        * `endpoint://`: the root file system on the target resource
+        * `endpoint://`: the root of the target resource's file system
         * `resource://`: the sandbox base dir on the target resource
         * `session://` : the session sandbox  on the target resource
         * `pilot://`   : the pilot   sandbox  on the target resource
--- a/src/radical/pilot/staging_directives.py
+++ b/src/radical/pilot/staging_directives.py
@@ -168,11 +168,14 @@ def complete_url(path   : str,
     The method returns an instance of `:py:class:`radical.utils.Url`.  Even if
     the URL is not altered, a new instance (deep copy) will be returned.
 
+    `endpoint://` is based on the `filesystem_endpoint` attribute of the
+    resource config and points to the file system accessible  via that URL.
+    Note that the notion of 'root' dependends of the access protocol and the
+    providing service implementation.
+
     NOTE: URL parsing is not really cheap, so this method should be used
     conservatively.
 
-    NOTE: For `endpoint://`, the notion of 'root' dependends of the access
-    protocol and the providing service implementation.
     '''
 
     # FIXME: consider evaluation of env vars
--- a/src/radical/pilot/utils/misc.py
+++ b/src/radical/pilot/utils/misc.py
@@ -154,8 +154,8 @@ def get_resource_config(resource: str) -> Union[None, ru.Config]:
 def get_resource_fs_url(resource: str,
                         schema  : str = None) -> Union[None, ru.Url]:
     '''
-    For the given resource label, return the contact URL of the resource's file
-    system.
+    For the given resource label, return the contact URL pointing to the root
+    of the resource's file system
 
     This URL is equivalent to the base ``endpoint:///`` URL available for use
     in a `staging_directive` once a `Session` is active.
--- a/src/radical/pilot/utils/misc.py
+++ b/src/radical/pilot/utils/misc.py
@@ -160,7 +160,7 @@ def get_resource_fs_url(resource: str,
     of 'root' dependends of the access protocol and the providing service
     implementation).
 
-    This URL is equivalent to the base ``endpoint:///`` URL available for use
+    For example, `rs.filesystem.directory(get_resource_fs_url(...)).change_dir('/')` is equivalent to the base ``endpoint:///`` URL available for use
     in a `staging_directive` once a `Session` is active.
 
     Args:
--- a/src/radical/pilot/utils/misc.py
+++ b/src/radical/pilot/utils/misc.py
@@ -154,14 +154,15 @@ def get_resource_config(resource: str) -> Union[None, ru.Config]:
 def get_resource_fs_url(resource: str,
                         schema  : str = None) -> Union[None, ru.Url]:
     '''
-    For the given resource label, return the contact URL
-    of the resource's file system.  This corresponds to the
-    `filesystem_endpoint` setting in the resource config. Note that the notion
-    of 'root' dependends of the access protocol and the providing service
-    implementation).
+    For the given resource label, return the contact URL of the resource's file
+    system.  This corresponds to the `filesystem_endpoint` setting in the
+    resource config. Note that the notion of 'root' dependends of the access
+    protocol and the providing service implementation).
 
-    For example, `rs.filesystem.directory(get_resource_fs_url(...)).change_dir('/')` is equivalent to the base ``endpoint:///`` URL available for use
-    in a `staging_directive` once a `Session` is active.
+    For example,
+    `rs.filesystem.directory(get_resource_fs_url(...)).change_dir('/')`
+    is equivalent to the base ``endpoint:///`` URL available for use
+    in a `staging_directive`.
 
     Args:
         resource (:obj:`str`): resource label for which to return the url
--- a/src/radical/pilot/utils/misc.py
+++ b/src/radical/pilot/utils/misc.py
@@ -154,7 +154,7 @@ def get_resource_config(resource: str) -> Union[None, ru.Config]:
 def get_resource_fs_url(resource: str,
                         schema  : str = None) -> Union[None, ru.Url]:
     '''
-    For the given resource label, return the contact URL pointing to the root
+    For the given resource label, return the contact URL
     of the resource's file system.  This corresponds to the
     `filesystem_endpoint` setting in the resource config. Note that the notion
     of 'root' dependends of the access protocol and the providing service
--- a/src/radical/pilot/utils/misc.py
+++ b/src/radical/pilot/utils/misc.py
@@ -156,8 +156,7 @@ def get_resource_fs_url(resource: str,
     '''
     For the given resource label, return the contact URL of the resource's file
     system.  This corresponds to the `filesystem_endpoint` setting in the
-    resource config. Note that the notion of 'root' dependends of the access
-    protocol and the providing service implementation).
+    resource config.
 
     For example,
     `rs.filesystem.directory(get_resource_fs_url(...)).change_dir('/')`
--- a/src/radical/pilot/utils/misc.py
+++ b/src/radical/pilot/utils/misc.py
@@ -155,7 +155,8 @@ def get_resource_fs_url(resource: str,
                         schema  : str = None) -> Union[None, ru.Url]:
     '''
     For the given resource label, return the contact URL pointing to the root
-    of the resource's file system
+    of the resource's file system (note that 'root' is here dependent of the
+    access protocol and the providing service implementation).
 
     This URL is equivalent to the base ``endpoint:///`` URL available for use
     in a `staging_directive` once a `Session` is active.
--- a/src/radical/pilot/utils/misc.py
+++ b/src/radical/pilot/utils/misc.py
@@ -155,8 +155,10 @@ def get_resource_fs_url(resource: str,
                         schema  : str = None) -> Union[None, ru.Url]:
     '''
     For the given resource label, return the contact URL pointing to the root
-    of the resource's file system (note that 'root' is here dependent of the
-    access protocol and the providing service implementation).
+    of the resource's file system.  This corresponds to the
+    `filesystem_endpoint` setting in the resource config. Note that the notion
+    of 'root' dependends of the access protocol and the providing service
+    implementation).
 
     This URL is equivalent to the base ``endpoint:///`` URL available for use
     in a `staging_directive` once a `Session` is active.


src/radical/pilot/staging_directives.py
```diff
@@ -145,7 +145,7 @@ def complete_url(path   : str,
     certain locations, namely relative to
 
         * `client://`  : the client's working directory
-        * `endpoint://`: the root file system on the target resource
+        * `endpoint://`: the root of the target resource's file system
```
eirrgang, line 148: A previous comment qualified this definition with "if possible". Can you clarify what happens in the not-possible scenario?

src/radical/pilot/utils/misc.py
```diff
@@ -154,8 +154,8 @@ def get_resource_config(resource: str) -> Union[None, ru.Config]:
 def get_resource_fs_url(resource: str,
                         schema  : str = None) -> Union[None, ru.Url]:
     '''
-    For the given resource label, return the contact URL of the resource's file
-    system.
+    For the given resource label, return the contact URL pointing to the root
+    of the resource's file system
```
eirrgang, line 158: I may have misunderstood. Is the result of this call the value of the filesystem_endpoint configuration field or is it the equivalent of `endpoint:///`?

I thought it was the former, but the comment implies the latter.

src/radical/pilot/staging_directives.py
andre-merzky: We access the system via some protocol - the service providing the protocol could do whatever, and we have no way of asserting that the accessed file system location is indeed the file system root.  It could be a sandbox the service lives in, for example.

src/radical/pilot/utils/misc.py
andre-merzky: I added more clarification (I hope).

src/radical/pilot/staging_directives.py
eirrgang: Would it be fair to say something like the "root-most directory reachable from the filesystem endpoint defined for the resource"?

src/radical/pilot/utils/misc.py
eirrgang: The documentation still isn't clear whether this is identically equal to the `filesystem_endpoint` field (or the `endpoint:///` URL) or how it maps to other known or knowable paths or symbols.

src/radical/pilot/staging_directives.py
andre-merzky: I don't think that would add any clarity, really, and trying to detail this more depending on the respective protocol is of little use.  I expect that a user or application who gets an URL with a certain protocol knows how to use that protocol.

src/radical/pilot/utils/misc.py
andre-merzky: The next sentence says:

```
This URL is equivalent to the base ``endpoint:///`` URL available for use in a `staging_directive` once a `Session` is active.
```

eirrgang: Is that true, though? I thought you said that `endpoint:///` might be equivalent to a URL that can't be represented as an `sftp://` URL (or... can you use relative paths that go to parent directories of `sftp://user@host/`?). Is the `filesystem_endpoint` translated so that the return value of this function is _always_ equivalent to `endpoint:///`?

If so, that's great! Then this documentation is correct, and I was mistaken.

src/radical/pilot/staging_directives.py
eirrgang: But there is no indication of the protocol here. In order to know what `root` should mean to the application programmer, we need to know what it means to RP. It is not necessarily obvious whether this scheme is derived from the `filesystem_endpoint` field or from `cd /` in the pilot/task environment.

That may be a lot to document right here. Elsewhere, I have suggested consolidating the staging directives documentation to a module docstring and providing more explanation. But if `endpoint:///` is strongly defined to be equivalent to the result of `radical.pilot.utils.misc.get_resource_fs_url()`, a "see also" would probably be sufficient.

src/radical/pilot/utils/misc.py
andre-merzky: Well, the same discussion we had on the other ticket I guess.  Yes, it depends on the protocol, but it is 'true enough'.  If somebody uses an `sftp` URL, she needs to know how it is interpreted.

eirrgang: > If somebody uses an sftp URL, she needs to know how it is interpreted.

What I'm looking for is some sort of note (and cross-linking) for an application programmer about how to understand `endpoint:///` in terms of other seemingly-related details, like the resource config fields and the `rp.utils.misc` functions.

The full details may be beyond the scope of this issue, but I think that it is exactly in the scope of this issue to clarify the caveats of how `endpoint:///` might point to something other than what is expected.

From the current documentation, the application programmer may expect that `endpoint:///` is equivalent to `chdir('/')` within a Task, `cd /` from the `job_manager_endpoint`, or `rs.filesystem.directory(get_resource_fs_url(...)).change_dir('/')`.

eirrgang: ... except I'm still not sure whether `rs.filesystem.directory(get_resource_fs_url(...)).change_dir('/')` is identically equal to `get_resource_fs_url(...)`

andre-merzky: Updated again - further documentation should later go to the resource configs docs.

src/radical/pilot/utils/misc.py
```diff
@@ -155,7 +155,10 @@ def get_resource_fs_url(resource: str,
                         schema  : str = None) -> Union[None, ru.Url]:
     '''
     For the given resource label, return the contact URL pointing to the root
-    of the resource's file system
+    of the resource's file system.  This corresponds to the
```
eirrgang, line 158: "corresponds to" is vague. This should be clear about whether the correspondence is exact or not.
```suggestion
    of the resource's file system.  This is derived from the
```

eirrgang: Ugh. I still seem to be confused. Looking at the source, I see `rcfg[schema]['filesystem_endpoint']`, implying that this is exactly equal to the `filesystem_endpoint` field. But that is inconsistent with the definition of `endpoint:///` as `rs.filesystem.directory(get_resource_fs_url(...)).change_dir('/')`

andre-merzky: No, it's not derived, it is exactly that URL.  I can remove the reference to `endpoint://` again if that turns out to be confusing after all...

eirrgang: I very much want some documentation somewhere of whether and how this value is related to `endpoint:///` because `endpoint:///` is the only way to reference paths outside of the session directory from the client side (right? At least in staging directive targets?)

eirrgang:
```suggestion
    of the resource's file system.  This is the
```

src/radical/pilot/utils/misc.py
```diff
@@ -154,8 +154,11 @@ def get_resource_config(resource: str) -> Union[None, ru.Config]:
...
     '''
-    For the given resource label, return the contact URL of the resource's file
-    system.
+    For the given resource label, return the contact URL pointing to the root
+    of the resource's file system.  This corresponds to the
+    `filesystem_endpoint` setting in the resource config. Note that the notion
+    of 'root' dependends of the access protocol and the providing service
+    implementation).
 
     This URL is equivalent to the base ``endpoint:///`` URL available for use
```
eirrgang, line 163:
```suggestion
    For example, `rs.filesystem.directory(get_resource_fs_url(...)).change_dir('/')` is equivalent to the base ``endpoint:///`` URL available for use
```

src/radical/pilot/utils/misc.py
```diff
@@ -154,8 +154,11 @@ def get_resource_config(resource: str) -> Union[None, ru.Config]:
 def get_resource_fs_url(resource: str,
                         schema  : str = None) -> Union[None, ru.Url]:
     '''
-    For the given resource label, return the contact URL of the resource's file
-    system.
+    For the given resource label, return the contact URL pointing to the root
```
eirrgang, line 157:
```suggestion
    For the given resource label, return the contact URL
```

src/radical/pilot/utils/misc.py
```diff
@@ -154,8 +154,11 @@ def get_resource_config(resource: str) -> Union[None, ru.Config]:
...
+    of the resource's file system.  This corresponds to the
+    `filesystem_endpoint` setting in the resource config. Note that the notion
+    of 'root' dependends of the access protocol and the providing service
```
eirrgang, line 160: This note seems more appropriate for the staging directive docstring, doesn't it?

src/radical/pilot/staging_directives.py
andre-merzky: > But if endpoint:/// is strongly defined...

it is.

eirrgang: > > But if endpoint:/// is strongly defined...
> 
> it is.

This contradicts https://github.com/radical-cybertools/radical.pilot/issues/2700#issuecomment-1262907802

If `radical.pilot.utils.misc.get_resource_fs_url()` returns the `filesystem_endpoint`, and `filesystem_endpoint` does not lead to the root of a filesystem, then `endpoint:///` cannot both be an absolute path and equal to the result of `radical.pilot.utils.misc.get_resource_fs_url()`.

What have I misunderstood?

andre-merzky: Well, yes, apart from any service containerization we cannot do nothing about - but again, that is a bit academic.  

I think we are running a bit in a circle here.

So, `fs_endpoint + cd / == endpoint:///`.  This usually points to the root of the systems' file system - and where it does not, we can not do anything about it.

andre-merzky:
Hi @eirrgang - can you please summarize what you think is left to do on this PR?  There are too many related threads right now and I am worried that I miss items...

eirrgang:
> Hi @eirrgang - can you please summarize what you think is left to do on this PR? There are too many related threads right now and I am worried that I miss items...

* staging_directives line 148: `endpoint://: the root of the target resource's filesystem_endpoint
* Move the sentence "Note that the notion of 'root'..." from misc.py to staging_directives.py
* If you know the path to the `filesystem_endpoint` field as a TypedDict attribute, let's link to it.

I think anything else can probably wait until there is progress on resource configs as TypedDicts, updates to machconf, or general updates to staging_directives docs.

src/radical/pilot/staging_directives.py
andre-merzky: That is because an URL does not have a notion of a relative path.  The need for `cd /` is an artifact of the service which interprets that URL and is outside the scope of what we can define.  We happen to handle it that way because SAGA works that way - but other tools may (and do) work differently.  The exposed URL is, per your use case, expected to be used outside of RP, and I just don't want to make assumptions on how it is used.

For example, using that URL with `curl` or `rsync` does not allow a notion of `cd /`, and there is nothing we can do about it...

andre-merzky:
> > Hi @eirrgang - can you please summarize what you think is left to do on this PR? There are too many related threads right now and I am worried that I miss items...
> 
>     * staging_directives line 148: `endpoint://: the root of the target resource's filesystem_endpoint
>     * Move the sentence "Note that the notion of 'root'..." from misc.py to staging_directives.py

I moved the sentence - does that fix both items?  If not, what does the first refer to?

>     * If you know the path to the `filesystem_endpoint` field as a TypedDict attribute, let's link to it.

That will happen as part of the resource dict typing.

Thanks, Andre.

eirrgang:
> > ```
> > * staging_directives line 148: `endpoint://: the root of the target resource's filesystem_endpoint
> > * Move the sentence "Note that the notion of 'root'..." from misc.py to staging_directives.py
> > ```
> 
> I moved the sentence - does that fix both items? If not, what does the first refer to?

I think it is important to note the connection between `endpoint://` and `filesystem_endpoint` (as opposed to `job_endpoint` or some other internal logic.

Current text:

    `endpoint://`: the root of the target resource's file system

Suggested text:

    `endpoint://`: the root of the target resource's `filesystem_endpoint`

It sounds like there are a lot of caveats to using `endpoint://`. I will try to submit documentation PRs as further clarification seems necessary and appropriate.

andre-merzky:
> It sounds like there are a lot of caveats to using `endpoint://`. I will try to submit documentation PRs as further clarification seems necessary and appropriate.

I don't think there are really - it's just the root of the filesystem.  I would like to avoid adding more documentation to edge cases which likely don't affect users and which would ultimately add confusion and not clarity.

Erik, I am puzzled by the amount of tweaking of the docs we try to do here.  Maybe I am missing something fundamental.  Do you have a reproducer of something actually broken?  Is there an actual case where `endpoint://` does not point to the system's root file system?

eirrgang:
> Erik, I am puzzled by the amount of tweaking of the docs we try to do here. Maybe I am missing something fundamental. Do you have a reproducer of something actually broken? Is there an actual case where `endpoint://` does not point to the system's root file system?

That's what I've been trying to establish. 🤣 As I understand your feedback, `endpoint://` points to the root of the filesystem pointed to by filesystem_endpoint (which isn't necessarily the same filesystem as the Task experiences) when possible, but it isn't always possible. Sometimes this is equal to `filesystem_endpoint`, but not generally.

The two things I wanted to clarify were the configuration value from which it is derived (and thus the filesystem hierarchy that it attempts to be the root of) and how hard RP tries to get to a directory that is "more root" than `filesystem_endpoint`.

andre-merzky:
> > Erik, I am puzzled by the amount of tweaking of the docs we try to do here. Maybe I am missing something fundamental. Do you have a reproducer of something actually broken? Is there an actual case where `endpoint://` does not point to the system's root file system?
> 
> That's what I've been trying to establish. rofl As I understand your feedback, `endpoint://` points to the root of the filesystem pointed to by filesystem_endpoint (which isn't necessarily the same filesystem as the Task experiences) when possible, but it isn't always possible. Sometimes this is equal to `filesystem_endpoint`, but not generally.
> 
> The two things I wanted to clarify were the configuration value from which it is derived (and thus the filesystem hierarchy that it attempts to be the root of) and how hard RP tries to get to a directory that is "more root" than `filesystem_endpoint`.

I think we established that `endpoint:// == filesystem_endpoint + 'cd /'` - and that is really all there is.  I don't think there are 'a lot of caveats', unless I miss anything?

eirrgang:
> I think we established that `endpoint:// == filesystem_endpoint + 'cd /'` - and that is really all there is. I don't think there are 'a lot of caveats', unless I miss anything?

I was reacting to "For example, using that URL with curl or rsync does not allow a notion of cd /, and there is nothing we can do about it...".

For the purpose of the present issue #2718, I think it is sufficient to state that `endpoint:///` is identically equal to `rs.filesystem.directory(get_resource_fs_url(...)).change_dir('/')`.

Based on the various schemes and varying notions of `cd /`, I don't have a sense right now on how often the *path* from a `endpoint://` URL is not usable as an absolute path to a task, but it seems like you are saying that it is, as much as possible. If it becomes tricky to convert an `endpoint://` URL to a filesystem path in cases that have not yet been encountered, I am content to address those as they arise, either in #2723 or other ad hoc issues.

eirrgang (approved):

PR #2719, opened 2022-09-28
Introduced 2 JSRun LMs: `JSRUN` and `JSRUN_ERF`

mtitov:


mtitov:
(*) RP Scheduler `NOOP` is not applicable for any LM right now, since it doesn't produce `slots-ranks` for tasks, but that is required for RP Spawner `POPEN`, thus it will be reviewed in a dedicated ticket/PR

andre-merzky (approved):
Thanks @mtitov , LGTM.  One question: should we allow an LM option to define the full path to the `jsrun` executable?  That's basically the only difference to `temp/jsrun_hpcw` if I am not mistaken?

mtitov:
@andre-merzky you mean the following part, right?
```python
        if os.environ.get('JSM_ROOT'):
            return ['$JSM_ROOT/bin/jsm &', '$JSM_ROOT/bin/jsrun' + cmd_options]
        else:
            return self._command + cmd_options
```
That was a solution proposed by OLCF ([link](https://docs.olcf.ornl.gov/systems/summit_user_guide.html#error-when-using-erf)), but it didn't work in our cases, thus I'm ok to skip it.

andre-merzky:
> That was a solution proposed by OLCF ([link](https://docs.olcf.ornl.gov/systems/summit_user_guide.html#error-when-using-erf)), but it didn't work in our cases, thus I'm ok to skip it.

Makes sense - thanks!

PR #2720, opened 2022-09-29
Docs/nov21

kartikmodi:
PR to see if we should merge Docs/nov21 into devel.


andre-merzky:
All work from this branch is now also captured in #2559 - this PR will be closed in favor of the other.

PR #2724, opened 2022-10-03
Use more recent Python versions in Docker

eirrgang:
* Py 3.8 is the most recent available to the Mongodb image, based on ubuntu/bionic.
* For the Docker compose stack, use Py 3.9.
* Also, normalize the path to which the local radical.pilot repository is copied.

### DIFF
--- a/docker/README.md
+++ b/docker/README.md
@@ -98,7 +98,7 @@ Once the services are up, start a shell in the `loging` service container with
 
 or invoke a Radical Pilot test with, e.g.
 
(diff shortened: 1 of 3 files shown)


eirrgang:
I tested both the rp-complete and the docker compose stack with example 00... and with misc/raptor.py.

Also, if there is an appropriate DockerHub account to use, I can make corresponding updates to suggested image names.

andre-merzky (approved):
LGTM

PR #2725, opened 2022-10-04
make sure we issue a `bootstrap_0_stop` event

andre-merzky:
See also radical-cybertools/radical.saga/pull/856


AymenFJA (approved):
I like the `trap` approach, LGTM, thanks for addressing this @andre-merzky.

PR #2726, opened 2022-10-07
Adding TASK_BULK_MKDIR_THRESHOLD as configurable Fixes #2089

### LINKED ISSUE #2089: Expose UNIT_BULK_MKDIR_THRESHOLD as configurable paramter
* Scenario: workload with 8k+ tasks on single pilot that have to be executed from Frontera's headsnode due to AA policies. 
* Problem: SAGA layer fails due to the number of threads required to bulk create the units' stage directories. This is an optimization introduced to minimize the impact of filesystem performance issues on RP overheads. Note that the failure is not fatal: SAGA will produce a trace error in the logs (assuming DEBUG is enabled) and a certain amount of units will be marked as failed. A separate issue might be open about better reporting this error.
* Solution: turn off bulk optimization in https://github.com/radical-cybertools/radical.pilot/blob/devel/src/radical/pilot/tmgr/staging_input/default.py#L30 by increasing the value so that the bulk mkdir does not kick in. `UNIT_BULK_MKDIR_THRESHOLD` should be exposed in the configuration layer and properly documented.

kartikmodi:


src/radical/pilot/configs/tmgr_default.json
```diff
@@ -14,7 +14,8 @@
 
     # time to sleep between database polls (seconds)
     "db_poll_sleeptime" : 1.0,
-
+    #
```
andre-merzky, line 17: This line is probably not needed?

kartikmodi:
@andre-merzky Please review. I think there was one more comment from you in the call, which I kind of forget. I did 2 changes.

src/radical/pilot/tmgr/staging_input/default.py
```diff
@@ -68,7 +68,8 @@ def initialize(self):
         # we subscribe to the command channel to learn about pilots being added
         # to this task manager.
         self.register_subscriber(rpc.CONTROL_PUBSUB, self._base_command_cb)
-
+        self.TASK_BULK_MKDIR_THRESHOLD = self.cfg.get('task_bulk_mkdir_threshold',
+                     TASK_BULK_MKDIR_THRESHOLD)
```
andre-merzky, line 72: class member variables should be lower case and short (only constants are all uppercase).  Also, private variables should have a leading underscrore. Maybe use `self._mkdir_threshold`?

andre-merzky (changes requested):
Looks good to me - but left a comment on the variable name.

andre-merzky (approved):
Thanks,    LGTM

PR #2744, opened 2022-10-24
agent does not need to pull failed tasks

### LINKED ISSUE #1592: failed units are pushed to pilot
see #1591: units which failed input staging get correctly moved into FAILED state, but then get pushed to the pilot nevertheless.  They should not.

andre-merzky:
This fixes #1592


src/radical/pilot/tmgr/staging_input/default.py
```diff
@@ -272,7 +272,13 @@ def work(self, tasks):
                          publish=True, push=True)
 
         for task,actionables in staging_tasks:
-            self._handle_task(task, actionables)
+            try:
+                self._handle_task(task, actionables)
+            except:
+                # staging failed - do not pass task to agent
+                task['control'] = 'tmgr'
```
mtitov, line 279: do we need to set it here? since we don't push task further and raise an exception right after. Or we can skip raising the exception and have the following
```python
task['control']      = 'tmgr_pending'
task['target_state'] = rps.FAILED
self.advance(task, rps.FAILED, publish=True, push=False)
```
or collect all such failed tasks and `advance` them together
(*) `task['$all']    = True` is already set earlier

andre-merzky: The raised exception will cause the task to fail in the component base class which will then push an update to MongoDB.  The `control` setting is needed to that the agent does not pull the task from there.  OTOH, the `tmgr` does not need to pull the task either, as the state update is pushed on the local `STATE_PUBSUB` anyway and the tmgr will update the task from there - the detour via MongoDB would not add anything.

andre-merzky: Wait, scrap that - I missed what happened to the other tasks.  Let me fix that.  But either way, it should not go through the DB...

andre-merzky: should be fixed - thanks!

mtitov: yeah, looks good! I'll add tests if you don't mind

andre-merzky: > yeah, looks good! I'll add tests if you don't mind

That's great, thanks!

mtitov:
(*) coverage got decreased, b/c somehow module `src/radical/pilot/tmgr/staging_input/default.py` wasn't tracked before, and only with this PR its coverage got checked.

mtitov (approved):
LGTM!

andre-merzky:
Thanks for the tests! :-)

PR #2745, opened 2022-10-24
implement task timeout for popen executor

### LINKED ISSUE #2352: Task timeout
During the HPC-workflow discussion on Friday, March 19, Wenjie mentioned that some of his tasks are executing indefinitely (RCT is not the cause of this behavior). As a result, he would like to add a timeout value per task.

1. ***Description***: Provide a timeout attribute in the task description. The executor will monitor task execution time and when the time has passed terminate the said task.
2. ***Example(s)***:
```python
 td = rp.TaskDescription()
td.executable    = '/bin/sleep'
td.arguments = ['120']
td.cpu_processes = 1
td.timeout = 60 # seconds

tmgr.submit_tasks(td)
...
# 60 seconds pass task get killed
```

Thank you!

andre-merzky:
This closes #2352


src/radical/pilot/agent/executing/base.py
```diff
@@ -117,7 +125,54 @@ def work(self, tasks):
...
+                # running tasks for next check
+                to_list = list()
+                for to, start, task in self._to_tasks:
+                    if now - start > to:
+                        self._prof.prof('task_timeout', uid=task['uid'])
+                        self.cancel_task(task)
+                    else:
+                        to_list.append([to, start, task])
+
+                self._to_tasks = to_list
```
mtitov, line 164: I would propose not to create a new list every time we go through the tasks (no need to make garbage collector's job harder :) )

either
```python
n_to_tasks = len(self._to_tasks)
while n_to_tasks:
    to, start, task = self._to_tasks.pop(0)
    if now - start > to:
        ...
    else:
        self._to_tasks.append([to, start, task])
    n_to_tasks -= 1
```
or just go through each element and remove cancelled tasks only
```python
idx = 0
for _ in range(len(self._to_tasks)):
    to, start, task = self._to_tasks[idx]
    if now - start > to:
        ...
        del self._to_tasks[idx]
    else:
        idx += 1
```

mtitov (commented):
Have a general comment - we should use one approach for cancelled tasks:
- have one list with cancelled tasks - `self._tasks_to_cancel` (could be set within the base class) - and apply method `cancel_task` to elements of this list only (and not outside of the method where this list is checked)
- this list is filled up either by RP (== based on timeout if it is set) or by user (== using command `cancel_tasks`)
   - for this purpose we can use `self._cancel_queue` (= `queue.Queue()`), in a similar way as `self._watch_queue` works within `Popen`

(*) this will help to avoid lock on `self._tasks_to_watch` when task is checked there (that would be not necessary) or when is removed from there, as as make it consistent.

src/radical/pilot/agent/executing/base.py
andre-merzky: I disagree with that - don't optimize before measuring, and even then, make sure it's worth the added complexity (we use the scheme above in many places).

andre-merzky:
> Have a general comment - we should use one approach for cancelled tasks:
> 
>     * have one list with cancelled tasks - `self._tasks_to_cancel` (could be set within the base class) - and apply method `cancel_task` to elements of this list only (and not outside of the method where this list is checked)
> 
>     * this list is filled up either by RP (== based on timeout if it is set) or by user (== using command `cancel_tasks`)
>       
>       * for this purpose we can use `self._cancel_queue` (= `queue.Queue()`), in a similar way as `self._watch_queue` works within `Popen`
> 
> 
> (*) this will help to avoid lock on `self._tasks_to_watch` when task is checked there (that would be not necessary) or when is removed from there, as as make it consistent.

This would be out of scope for this PR for sure I think?

src/radical/pilot/agent/executing/base.py
andre-merzky:
```py
#!/usr/bin/env python3

import time
import random

def test_1(n):
    items = list(range(n))
    n_items = len(items)
    while n_items:
        item = items.pop(0)
        if random.choice([True, False]):
            pass
        else:
            items.append(item)
        n_items -= 1


def test_2(n):
    items = list(range(n))
    idx = 0
    for _ in range(len(items)):
        item = items[idx]
        if random.choice([True, False]):
            del items[idx]
        else:
            idx += 1


def test_3(n):
    items = list(range(n))
    new_items = list()
    for item in items:
        if random.choice([True, False]):
            new_items.append(item)

    items = new_items


if __name__ == '__main__':

    for m in [1, 2, 4, 8, 16, 32, 64, 128, 256, 512]:

        n = 1024 * m

        start = time.time()
        test_1(n)

        print('%6dk : ' % m, end='')
        print(' %7.1f' % (time.time() - start), end='')

        start = time.time()
        test_2(n)
        print(' %7.1f' % (time.time() - start), end='')

        start = time.time()
        test_3(n)
        print(' %7.1f' % (time.time() - start), end='')

        print()
```

```sh
$ ./t.py 
     1k :      0.0     0.0     0.0
     2k :      0.0     0.0     0.0
     4k :      0.0     0.0     0.0
     8k :      0.0     0.0     0.0
    16k :      0.0     0.0     0.0
    32k :      0.2     0.1     0.0
    64k :      0.7     0.3     0.0
   128k :      3.0     1.0     0.1
   256k :     11.0     3.7     0.1
   512k :     45.0    15.7     0.3
```

andre-merzky:
@mtitov : 

> Have a general comment - we should use one approach for cancelled tasks:
> 
>     * have one list with cancelled tasks - `self._tasks_to_cancel` (could be set within the base class) - and apply method `cancel_task` to elements of this list only (and not outside of the method where this list is checked)
> 
>     * this list is filled up either by RP (== based on timeout if it is set) or by user (== using command `cancel_tasks`)
>       
>       * for this purpose we can use `self._cancel_queue` (= `queue.Queue()`), in a similar way as `self._watch_queue` works within `Popen`
> 
> 
> (*) this will help to avoid lock on `self._tasks_to_watch` when task is checked there (that would be not necessary) or when is removed from there, as as make it consistent.

I agree this can be improved, but I would like to defer refactoring due to lack of time :-/

src/radical/pilot/agent/executing/popen.py
```diff
@@ -499,9 +476,48 @@ def _check_running(self):
...
+        self._prof.prof('task_run_cancel_stop', uid=tid)
+
+        del task['proc']  # proc is not json serializable
+        self._prof.prof('unschedule_start', uid=tid)
+        self.publish(rpc.AGENT_UNSCHEDULE_PUBSUB, task)
+        self.advance(task, rps.CANCELED, publish=True, push=False)
+
+        # we don't need to watch canceled tasks
+        # FIXME: list action should be locked
+        self._tasks_to_watch.remove(task)
```
mtitov, line 515: should we have some "lock" attribute (`self._watch_lock     = mt.RLock()`?), which will be used when tasks are added to `_tasks_to_watch` and removed from it?

src/radical/pilot/agent/executing/base.py
```diff
@@ -117,7 +125,54 @@ def work(self, tasks):
...
+
+            now = time.time()
+            with self._to_lock:
+
+                # running tasks for next check
+                to_list = list()
+                for to, start, task in self._to_tasks:
+                    if now - start > to:
+                        self._prof.prof('task_timeout', uid=task['uid'])
+                        self.cancel_task(task)
```
mtitov, line 160: for this part, should we have `_cancel_lock`? I mean to have it within base class, but Popen also will use it whenever method `cancel_task` is called?

mtitov (commented):
couple more comments/proposals to update locks

src/radical/pilot/agent/executing/base.py
andre-merzky: Hey @mtitov - I refactored the code quite a bit: cancel requests are now pushed into the watch queue, negating the need for a thread-shared variable and locks.  Let me know if that works for you.

On the down side that makes the tests a bit tricky AFAICT - do you have a suggestion on how the failing test could be restructured?  Should we just check if the next `get` on the watch queue yields the task ID?

src/radical/pilot/agent/executing/popen.py
```diff
@@ -52,17 +52,18 @@ class Popen(AgentExecutingComponent):
...
     def initialize(self):
 
       # self._log.debug('popen initialize start')
         AgentExecutingComponent.initialize(self)
 
-        self._cancel_lock     = mt.RLock()
-        self._tasks_to_cancel = list()
-        self._tasks_to_watch  = list()
-        self._watch_queue     = queue.Queue()
+        self._watch_queue = queue.Queue()
```
mtitov, line 66: maybe to have two queues instead? (`_watch_queue`, `_cancel_queue`), just wondering about possibility to have task cancelation delayed, could it be?

src/radical/pilot/agent/executing/popen.py
```diff
@@ -417,73 +437,71 @@ def _watch(self):
     # --------------------------------------------------------------------------
     # Iterate over all running tasks, check their status, and decide on the
     # next step.  Also check for a requested cancellation for the tasks.
-    def _check_running(self):
+    def _check_running(self, to_watch, to_cancel):
+
+        action = False
+        for task in to_watch:
```
mtitov, line 443: `for task in list(to_watch):`? (or some other iteration over `to_watch` list)

just found it as a long standing bug - some tasks could be skipped until the following round(s), due to other tasks are deleted during the iteration process - gives some delay for tasks processing

```
>>> tasks = [{1:1}, {2:2}, {3:3}]
>>> for task in tasks:
...     print('Current task: %s (from %s)' % (task, tasks))
...     if 1 in task or 2 in task:
...         tasks.remove(task)
... 
Current task: {1: 1} (from [{1: 1}, {2: 2}, {3: 3}])
Current task: {3: 3} (from [{2: 2}, {3: 3}])
```

src/radical/pilot/agent/executing/popen.py
```diff
@@ -369,42 +379,52 @@ def _handle_task(self, task):
...
+        # watch task for completion
+        self._watch_queue.put([self.TO_WATCH, task])
 
 
     # --------------------------------------------------------------------------
     #
     def _watch(self):
 
+        to_watch  = list()  # list of task dicts
+        to_cancel = list()  # list of task IDs
```
mtitov, line 394: maybe to have it as a `set`? it gives a better performance for lookup and in case of duplication requests for task cancelation

mtitov (approved):
LGTM! Thank you!

PR #2749, opened 2022-10-24
Feature/parsl raptor executor

andre-merzky:
@AymenFJA: the parsl raptor executor seems functional so far.  As we discussed, it would be better placed in parsl or should be moved to a separate repository.  Which way should we go?  I opened this PR so that we can discuss what changes in other files should or should not make it into RP.


examples/parsl/colmena_rp/colmena_rp.py
```diff
@@ -0,0 +1,168 @@
+#!/usr/bin/env python3
```
andre-merzky, line 1: I think this file (and related files under `example/parsl/`) should move to where-ever the executor ends up.

src/radical/pilot/agent/agent_0.py
```diff
@@ -839,6 +839,9 @@ def _prepare_env(self, env_name, env_spec):
 
         # if the ve lives outside of the pilot sandbox, link it
         if path:
+            # expand any env variables in path
+            path, _, _ = ru.sh_callout('echo -n "%s"' % path, shell=True)
+            path = os.path.abspath(path)
```
andre-merzky, line 844: This expansion either needs documentation in `pilot.prepare_env`, or we should remove it again.  If we keep it, then error checks should be added to the code.

src/radical/pilot/pmgr/launching/default.py
```diff
@@ -0,0 +1,1242 @@
+# pylint: disable=protected-access
```
andre-merzky, line 1: This file does not exist in `devel` anymore

andre-merzky (commented):
Some comments.

andre-merzky:
@AymenFJA : given that https://github.com/Parsl/parsl/pull/2507 is open now, should we close this PR?  Or should are there parts which are still merge candidates and should be moved into a separate PR?

AymenFJA:
> @AymenFJA : given that [Parsl/parsl#2507](https://github.com/Parsl/parsl/pull/2507) is open now, should we close this PR? Or should are there parts which are still merge candidates and should be moved into a separate PR?

@andre-merzky  I think we can close this PR safely. Thanks for raising this.

PR #2750, opened 2022-10-24
cancel  unscheduled raptor tasks

andre-merzky:


src/radical/pilot/agent/scheduler/base.py
```diff
@@ -223,11 +223,12 @@ def initialize(self):
         # waitlist to a binned list where tasks are binned by size for faster
         # lookups of replacement tasks.  And outdated binlist is mostly
         # sufficient, only rebuild when we run dry
-        self._waitpool   = dict()  # map uid:task
+        self._lock       = mt.Lock()  # lock for waitpool
```
mtitov, line 226: should this lock be moved next to `raptor_lock` (L636) - keep it within `_schedule_tasks`

andre-merzky: `raptor_lock` needs to be in `_schedule_tasks` as that method runs in a different process.  `self._lock` however applies to the parent process, and thus needs to be defined here - `initialize` is the canonical place for it really I think.

andre-merzky: I would like to sort this out a bit more carefully - please allow me a couple of days to look through this again.

PR #2755, opened 2022-10-28
Added possibility to set option `cpus-per-task` off for SRun LM

mtitov:
(*) will be updated with LMOptions - will be controlled by LM config


andre-merzky (approved):
LGTM!

PR #2758, opened 2022-10-31
Document attributes as `attr`, not `data`.

### LINKED ISSUE #2757: Class attributes should be documented with `.. py:attribute::`, not `.. py:data::`.
Attributes of TaskDescription and PilotDescription are documented with the wrong Sphinx directive. They do not work as targets of `:py:attr:` roles and may not be formatted appropriately, depending on the theme.

[py:data](https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#directive-py-data) elements describe global data in a module.

[py:attr](https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#directive-py-attribute) elements describe an object data attribute.

@eirrgang will submit a PR.

eirrgang:
Fixes #2757


andre-merzky (approved):
Thanks Eric!

Issue #2759, opened 2022-11-01
missing `cores_per_node` setting in resource config results in pilot with no resources configured


andre-merzky:
That might even be ok - but we should at least bail out on the client side instead of submitting and failing the pilot on `agent.0.cfg` interpretation...

mtitov:
@andre-merzky does PR #2761 resolve this?

andre-merzky:
Yes, thanks!

PR #2760, opened 2022-11-01
update amarel cfg

andre-merzky:
This is urgent because of #2759

### DIFF
--- a/src/radical/pilot/configs/resource_rutgers.json
+++ b/src/radical/pilot/configs/resource_rutgers.json
@@ -31,6 +31,7 @@
         "python_dist"                 : "default",
         "virtenv_dist"                : "default",
         "virtenv_mode"                : "create",
-        "rp_version"                  : "local"
+        "rp_version"                  : "local",
+        "cores_per_node"              : 64
     }
 }


AymenFJA:
@andre-merzky  I just confirmed this fix the issue that Mason is facing in his session on Amarel.

AymenFJA (approved):
LGTM

mtitov:
@andre-merzky @AymenFJA we didn't set this parameter since it was a heterogeneous cluster with nodes of different number of cores [1], and we let SAGA to determine the correct `cpn`, was it changed since then?

[1] https://sites.google.com/view/cluster-user-guide#h.kyrykrouyxxz

andre-merzky:
[This line](https://github.com/radical-cybertools/radical.pilot/blob/devel/src/radical/pilot/pmgr/launching/base.py#L580) defaults `cores_per_node` to `0` when unset in the config, which makes [this line](https://github.com/radical-cybertools/radical.pilot/blob/devel/src/radical/pilot/pmgr/launching/base.py#L806) to derive `allocated_cores` to `0` also, which is what ends up in the [agent config](https://github.com/radical-cybertools/radical.pilot/blob/devel/src/radical/pilot/pmgr/launching/base.py#L858).  That in turn will be used as [`requested_cores`](https://github.com/radical-cybertools/radical.pilot/blob/devel/src/radical/pilot/agent/resource_manager/base.py#L213) by the resource manager, which then calculates [`requested_nodes`](https://github.com/radical-cybertools/radical.pilot/blob/devel/src/radical/pilot/agent/resource_manager/base.py#L239) from it, resulting in 0 nodes which causes it to bail out.

So something likely changed indeed - for now this PR is only a fix for amarel, the actual issue is tracked in #2759.

mtitov:
> That in turn will be used as [requested_cores](https://github.com/radical-cybertools/radical.pilot/blob/devel/src/radical/pilot/agent/resource_manager/base.py#L213) by the resource manager, which then calculates [requested_nodes](https://github.com/radical-cybertools/radical.pilot/blob/devel/src/radical/pilot/agent/resource_manager/base.py#L239) from it, resulting in 0 nodes which causes it to bail out.

this part suppose to be updated by [this](https://github.com/radical-cybertools/radical.pilot/blob/devel/src/radical/pilot/agent/resource_manager/slurm.py#L29-L34)

mtitov:
sorry, misread some part - yeah, so we need to provide to agent either `allocated_cores` or `requested_cores` (as we do for job description)
```python
agent_cfg['cores'] = allocated_cores or requested_cores
```

andre-merzky:
> ```python
> agent_cfg['cores'] = allocated_cores or requested_cores
> ```

Yes, that would work.

PR #2762, opened 2022-11-02
Some scheduler improvements

andre-merzky:
This fixes some performance bottlenecks of the agent scheduler.

Notably it moves tag interpretation out of the scheduler into a new `AGENT_RESOLVER` component.


andre-merzky:
ping @andre-merzky : salvage parts and close

PR #2778, opened 2022-11-21
agent services

kartikmodi:


examples/misc/agent_services.cfg
```diff
@@ -0,0 +1,29 @@
+
```
andre-merzky, line 1: I don't really understand what this config and the example below are for: there are many values defined which are not used?

src/radical/pilot/agent/agent_0.py
```diff
@@ -83,7 +86,9 @@ def __init__(self, cfg, session):
         self._cmgr.start_components()
 
         # start any services if they are requested
-        self._start_services()
+        # TODO: Remove this. do not start services from here
```
andre-merzky, line 89: You inserted `start_services` below, so you can indeed remove these commented lines.

src/radical/pilot/agent/agent_0.py
```diff
@@ -340,6 +347,12 @@ def _start_services(self):
         If a `./services` file exist, reserve a compute node and run that file
         there as bash script.
         '''
+        # TODO
+        #  there is no pilot description/pilot instance accessible here. how to get services list
```
andre-merzky, line 351: The pilot gets it's configuration from the pilot launcher.  Currently service commands are passed as arguments to the bootstrapper (see [here](https://github.com/radical-cybertools/radical.pilot/blob/devel/src/radical/pilot/pmgr/launching/base.py#L851)).  That should be changed to add them to the `agent_config` (same file, 5 lines below, the agent config is created).  If that happens, then the service descriptions will be available here as `self._cfg.services`.

src/radical/pilot/agent/agent_0.py
```diff
@@ -340,6 +347,12 @@ def _start_services(self):
...
+        agent_service_tasks = [how to get services TaskDescription]
+        self.advance(agent_service_tasks, publish=False, push=True)
```
andre-merzky, line 353: You need to be careful: what you get from the client is a `TaskDescription` dictionary, what you want to `advance` here is a `Task` dictionary.  So you need to create a `Task` dictionary from the descriptions.  Look into the Raptor master for an example on how this can be done: [here](https://github.com/radical-cybertools/radical.pilot/blob/devel/src/radical/pilot/raptor/master.py#L325)

src/radical/pilot/agent/agent_0.py
```diff
@@ -355,6 +368,7 @@ def _start_services(self):
         service_task_uid = 'rp.services'
         service_task     = {
             'uid'               : service_task_uid,
+            'rank_per_node'     : 2,
```
andre-merzky, line 371: Why was that added?

src/radical/pilot/agent/agent_0.py
```diff
@@ -408,6 +422,10 @@ def _start_services(self):
         self._log.info('create services: %s' % cmdline)
         ru.sh_callout_bg(cmdline, stdout='services.out', stderr='services.err')
 
+        while(all_services_not_started)
+            # WHERe to check state of executing
+            service.state == AGENT_EXECUTING
```
andre-merzky, line 427: State upates are published via ZMQ on the `STATE_PUBSUB`  channel.  Have a look at the Raptor Master again [here](https://github.com/radical-cybertools/radical.pilot/blob/devel/src/radical/pilot/raptor/master.py#L55) and [here](https://github.com/radical-cybertools/radical.pilot/blob/devel/src/radical/pilot/raptor/master.py#L219) on how to subscribe to state updates and how to evaluate them (while here the agent waits for service task state updates, the master waits for worker task updates -- same thing :-) )

andre-merzky (changes requested):
This looks like progress - left a couple of comments.

examples/misc/agent_services.cfg
kartikmodi: I inteded to read this config file as an input for service resource config. Ranks_per_node is the new config which doesn't exist as of now

src/radical/pilot/agent/agent_0.py
kartikmodi: yes

src/radical/pilot/agent/agent_0.py
kartikmodi: done

src/radical/pilot/agent/agent_0.py
kartikmodi: tried. See now. @andre-merzky

src/radical/pilot/agent/agent_0.py
kartikmodi: commented for now

src/radical/pilot/agent/agent_0.py
kartikmodi: Someone is publishing the cmd like - 
```        if buckets['raptor']:
            self.publish(rpc.STATE_PUBSUB, {'cmd': 'raptor_state_update',
                                            'arg': buckets['raptor']})
```

src/radical/pilot/agent/agent_0.py
```diff
@@ -408,9 +430,44 @@ def _start_services(self):
         self._log.info('create services: %s' % cmdline)
         ru.sh_callout_bg(cmdline, stdout='services.out', stderr='services.err')
 
+        self.publish(rpc.STATE_PUBSUB, {'cmd': 'service_state_update',
+                                        'arg': services})
         self._log.debug('services started done')
 
+    def _state_cb_of_services(self, topic, msg):
```
kartikmodi, line 437: I have no idea what I am doing here. Help @andre-merzky

andre-merzky (commented):
push pending commit

src/radical/pilot/agent/agent_0.py
kartikmodi: addressed

src/radical/pilot/agent/agent_0.py
```diff
@@ -340,6 +349,18 @@ def _start_services(self):
...
+        for service_desc in service_descriptions:
+            task = dict()
+
+            task['description'] = TaskDescription(service_desc).as_dict()
+            task['state'] = rps.AGENT_STAGING_INPUT_PENDING
+            task['status'] = 'NEW'
+            task['type'] = 'task'
+            services.append(task)
+
+        self.advance(services, publish=False, push=True)
```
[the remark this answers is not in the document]
kartikmodi: understood

src/radical/pilot/agent/agent_0.py
```diff
@@ -446,27 +446,33 @@ def _state_cb_of_services(self, topic, msg):
...
+                    pass
+                    # TODO We should do something.
+
+                # if uid in self._task_service_data:
+                #     # update task info and signal task service thread
+                #     self._log.debug('unlock 2 %s', uid)
+                #     self._task_service_data[uid][1] = task
+                #     self._task_service_data[uid][0].set()
+        # Waiting for every callback from every service, and then notify the main thread.
+        e.wait()
```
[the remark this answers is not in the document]
kartikmodi: I have used the Event class but no proper multithreading in python. 
Want to wanted to do - 


```
_start_services()
threading
... spawn threads
... event 
... event.set() //blcoked until all threads are done


thread_method()
... start and check service for a particular
... e.wait()
```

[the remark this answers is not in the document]
andre-merzky: Ah, now I see what you are trying.  Allow me to suggest a simpler path.

In line 364, you advance the service tasks and push them to the stager / scheduler / executor components.  Right after that `advance` call is when you want to wait until the tasks are executing, so you can add there something like 

```py
self._service_wait_event.wait()
```

(possibly with some timeout and error check).

Here, in `_state_cb_of_services`, you will get state notifications for all tasks.  There is no need to issue your own events (`service_state_update`) - the executor component will send a state update already when the service tasks are entering `AGENT_EXECUTING` state.  You pasted the code from `https://github.com/radical-cybertools/radical.pilot/blob/devel/src/radical/pilot/raptor/master.py#L245` which I referred to earlier in the comment below - that is the code you should actually use (just the state needs changing).

So you basically just have to count the state update messages - if you start 3 service tasks, you should get 3 state update messages there (to be safe, you can also check the task IDs to make sure you check the right messages).  Once you received all three it would be save to set
```py
self._service_wait_event.set()
```
to let the other thread (which does the `...wait()` above) continue.

```py
if cmd == 'update':
    for thing in ru.as_list(arg):
        uid   = thing['uid']
        state = thing['state']
        if uid in self._service_task_ids and state == rps.AGENT_EXECUTING:
            self._active_services += 1
            if self._active_services == len(self._service_task_ids):
            self._service_wait_event.set()
```
or something like that...

src/radical/pilot/agent/agent_0.py
```diff
@@ -354,106 +356,137 @@ def _start_services(self):
...
 
     def _state_cb_of_services(self, topic, msg):
         cmd = msg['cmd']
-        arg, e = msg['arg']
-        if cmd == 'service_state_update':
+        services = msg['arg']
 
-            tasks = ru.as_list(arg)
+        if cmd == 'update':
+            self._log.info('received services for launching: %s' % services)
```
andre-merzky, line 456: The state pubsub will get state updates for *all* tasks, not only for service tasks.  So you will need to check if the task update belongs to one of your service tasks, by checking the task uid.

src/radical/pilot/agent/agent_0.py
```diff
@@ -354,106 +356,137 @@ def _start_services(self):
...
+            e.set()
 
-            for task in tasks:
-                uid   = task['uid']
-                state = task['state']
+            self._log.debug('services started done')
 
-                if state == rps.AGENT_EXECUTING:
+            running_services = []
+            yet_to_run = []
```
andre-merzky, line 463: I don't think you need to track services which are not in the `AGENT_EXECUTING` state.  Also, you will double-count because the service tasks go through several states before `AGENT_EXECUTING`.

src/radical/pilot/agent/agent_0.py
kartikmodi: Got it.

src/radical/pilot/agent/agent_0.py
kartikmodi: Ohh. Yes. The `yet_to_run` can get multiple entries for different states of same task. Ofcourse we can solve that easily, but I am not sure of what I'll do with this `yet_to_run` in future. So will remove

src/radical/pilot/agent/agent_0.py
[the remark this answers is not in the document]
kartikmodi: made changes. please see

src/radical/pilot/agent/agent_0.py
```diff
@@ -340,77 +353,130 @@ def _start_services(self):
...
+            task['state'] = rps.AGENT_STAGING_INPUT_PENDING
+            task['status'] = 'NEW'
+            task['type'] = 'task'
+            task['uid'] = ru.generate_id('services.%(item_counter)04d',
+                                             ru.ID_CUSTOM, ns="TODO_WHATTOUSE")
+            self._service_task_ids.add(task['uid'])
+            task['task_sandbox_path'] = self._sbox
+            task['task_sandbox'] = 'file://localhost/' + self._sbox
+            task['pilot_sandbox'] = self._cfg.pilot_sandbox
+            # task['session_sandbox'] = os.environ['RP_SESSION_SANDBOX']
```
kartikmodi, line 372: @andre-merzky Need help here. How to get these

src/radical/pilot/agent/agent_0.py
```diff
@@ -340,77 +353,130 @@ def _start_services(self):
...
-        nodes = self._rm.info.service_node_list
-        assert nodes
-
-        bs_name = "%s/bootstrap_2.sh"     % self._pwd
-        ls_name = "%s/services_launch.sh" % self._pwd
-        ex_name = "%s/services_exec.sh"   % self._pwd
-
-        service_task_uid = 'rp.services'
-        service_task     = {
-            'uid'               : service_task_uid,
```
kartikmodi, line 357: All this code shall not be used @andre-merzky  I think. do you see the same

src/radical/pilot/agent/agent_0.py
andre-merzky: `self._cfg.session_sandbox` etc.  You may want to have a look at an agent config, it contains quite a number of config items you are using throughout the code.

src/radical/pilot/agent/agent_0.py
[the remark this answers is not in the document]
andre-merzky: The task_resources and the sandboxes will be needed also.

On a general note: comment threads should be `resolved` by whoever opens them.  If you close them, I have to reopen them all to check if the are really done.  Thanks :-)

src/radical/pilot/agent/agent_0.py
```diff
@@ -340,77 +353,130 @@ def _start_services(self):
...
+        #
+        # spawn the sub-agent
+        # cmdline = './%s' % ls_name
+        #
+        # self._log.info('create services: %s' % cmdline)
+        # ru.sh_callout_bg(cmdline, stdout='services.out', stderr='services.err')
 
-        with ru.ru_open(ex_name, 'w') as fout:
-            fout.write(tmp)
+        # self.publish(rpc.STATE_PUBSUB, {'cmd': 'service_state_update','arg': services})
```
andre-merzky, line 456: The code you commented out above can be removed.

src/radical/pilot/agent/agent_0.py
```diff
@@ -340,77 +353,130 @@ def _start_services(self):
...
 
+    def _state_cb_of_services(self, topic, msg):
+        cmd = msg['cmd']
+        services = msg['arg']
```
andre-merzky, line 460: You do not yet know if the state update is for your service - it could be for any other task.  You check if it is service task only on line 470 below.  So maybe `tasks = msg['arg']` is more appropriate?

It's a minor thing, but makes the code more readable.

src/radical/pilot/agent/agent_0.py
```diff
@@ -340,77 +353,130 @@ def _start_services(self):
...
 
-        self._log.info('create services: %s' % cmdline)
-        ru.sh_callout_bg(cmdline, stdout='services.out', stderr='services.err')
+        running_services = []
 
-        self._log.debug('services started done')
+        if cmd == 'update':
 
+            for service in ru.as_list(services):
+                if "service" in service['uid']:
```
andre-merzky, line 471: What happens if a user starts a task and assignes it the uid `service.foo.1`?  This code will go boom...

3 lines below you have a different check `if uid in self._service_task_ids` - that seems much safer.  You only need to check once really, so I would suggest to use the second one here.

src/radical/pilot/agent/agent_0.py
```diff
@@ -6,11 +6,13 @@
...
 import pprint
 import stat
+import threading
 import time
+from concurrent.futures import ThreadPoolExecutor
 
 import radical.utils       as ru
 
-from ..   import utils     as rpu
+from .. import utils as rpu, session
```
andre-merzky, line 15: This syntax is confusing.  Do you need to import `..session`?  I don't see it being used.  If you need it, please import on a separate line.

src/radical/pilot/agent/agent_0.py
```diff
@@ -6,11 +6,13 @@
 import os
 import pprint
 import stat
+import threading
 import time
+from concurrent.futures import ThreadPoolExecutor
```
andre-merzky, line 11: Why is the threadpool executor imported?

src/radical/pilot/agent/agent_0.py
```diff
@@ -21,6 +23,9 @@
 
 # ------------------------------------------------------------------------------
 #
+from ... import pilot
```
andre-merzky, line 26: Where is `pilot` being used?

src/radical/pilot/agent/agent_0.py
```diff
@@ -95,6 +103,8 @@ def __init__(self, cfg, session):
         rpu.Worker.__init__(self, self._cfg, session)
 
         self.register_subscriber(rpc.CONTROL_PUBSUB, self._check_control)
+        # TODO do I need to pass pilot sandbox
```
andre-merzky, line 106: No - the callback  runs in a separate thread, but has access to the class state, so you don't need to pass anything.

src/radical/pilot/agent/agent_0.py
```diff
@@ -108,7 +118,8 @@ def __init__(self, cfg, session):
                                 term_cb=self._hb_term_cb,
                                 log=self._log)
         self._hb.start()
-
+        # Event to handle for services started
+        self.services_event = threading.Event()
```
andre-merzky, line 122: It makes sense to define this in the constructor - but why is it mixed into the hearbeat code?  Please keep related code together and move it up to the service state cb.

src/radical/pilot/agent/agent_0.py
```diff
@@ -340,77 +353,130 @@ def _start_services(self):
         If a `./services` file exist, reserve a compute node and run that file
         there as bash script.
         '''
-
-        if not os.path.isfile('./services'):
-            return
+        service_descriptions = self._cfg.services
+        services = list()
+        self._service_task_ids = {}
```
andre-merzky, line 358: If possible, please declare class members in the constructor.  Otherwise you imply a certain sequence of method calls which might not always be given (remember that the state callback is being called asynchronously).

src/radical/pilot/agent/agent_0.py
kartikmodi: Woops. will remove. I tried several several things with this earlier, when didn't had the clarity that these service tasks will be executed by rp.

src/radical/pilot/agent/agent_0.py
kartikmodi: removed

src/radical/pilot/agent/agent_0.py
kartikmodi: ok

src/radical/pilot/agent/agent_0.py
kartikmodi: Yes definitely.

src/radical/pilot/agent/agent_0.py
kartikmodi: removed. not used

src/radical/pilot/agent/agent_0.py
kartikmodi: ok

src/radical/pilot/agent/agent_0.py
kartikmodi: Woops. I see now. I dont know why I did that specifically.

src/radical/pilot/agent/agent_0.py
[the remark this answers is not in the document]
kartikmodi: Ohh okay. I closed several of them now. Will reopen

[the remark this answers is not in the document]
kartikmodi: Didn't understood `The task_resources and the sandboxes will be needed also.`

src/radical/pilot/agent/agent_0.py
kartikmodi: removed

src/radical/pilot/agent/agent_0.py
kartikmodi: moved

kartikmodi:
Looking into tests failure

src/radical/pilot/agent/agent_0.py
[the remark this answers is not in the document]
kartikmodi: Need more clarity on this part.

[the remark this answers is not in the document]
andre-merzky: Can you pleas clarify what you would like to get feedback on?

src/radical/pilot/agent/agent_0.py
```diff
@@ -340,77 +351,61 @@ def _start_services(self):
...
+            task['description'] = TaskDescription(service_desc).as_dict()
+            task['state'] = rps.AGENT_STAGING_INPUT_PENDING
+            task['status'] = 'NEW'
+            task['type'] = 'task'
+            task['uid'] = ru.generate_id('services.%(item_counter)04d',
+                                             ru.ID_CUSTOM, ns="TODO_WHATTOUSE")
+            self._service_task_ids.append(task['uid'])
+            task['pilot_sandbox'] = self._cfg.pilot_sandbox
+            task['task_sandbox'] = self._cfg.pilot_sandbox + task['uid'] + '/'
+            task['task_sandbox_path'] = 'file://localhost/' + self._cfg.pilot_sandbox + task['uid'] + '/'
```
kartikmodi, line 370: @andre-merzky this is wrong. Making path -> `/home/workstation/radical.pilot.sandbox/rp.session.debian-x.workstation.019375.0001/pilot.0000/file:/localhost/home/workstation/radical.pilot.sandbox/rp.session.debian-x.workstation.019375.0001/pilot.0000/services.0021`

I copied this. should this be `self._cfg.pilot_sandbox + task['uid'] + '/'`

andre-merzky: I may be missing something, but what is wrong?  The line you reference above will not result in a paths like `/home/.../file://...pilot.sandbox`, but rather in a path like `file://localhost/home/.../service.0000/`.  That might be wrong, I don't know, but the point is I don't understand your comment.

src/radical/pilot/agent/agent_0.py
andre-merzky: right

src/radical/pilot/agent/agent_0.py
andre-merzky: Please don't resolve comments which are not addressed.  Again: the policy is that the person who comments is also the one who resolves the comment.

src/radical/pilot/agent/agent_0.py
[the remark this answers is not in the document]
kartikmodi: This is clear now

src/radical/pilot/agent/agent_0.py
kartikmodi: This is sorted after the last commit I pushed after this comment.

andre-merzky:
@kartikmodi : If this works now then the following steps should follow:
  - remove the `draft` flag from the PR
  - fix the linting errors
  - make sure the tests cover the new code

After linting is green I'll do another review for details.  Thanks!

src/radical/pilot/agent/agent_0.py
kartikmodi: resolved.

kartikmodi:
@andre-merzky 

> remove the draft flag from the PR 
 DONE
> fix the linting errors 
 DONE

> make sure the tests cover the new code 
 ok. I'll check the test case coverage. If not covered, I'll have to write a test case for this, but maybe we can start review.

mtitov:
@andre-merzky should we keep this PR as a draft, and when it will be ready for merge just create another one? (since this PR was more as a testing one with many tryouts)

andre-merzky:
> @andre-merzky should we keep this PR as a draft, and when it will be ready for merge just create another one? (since this PR was more as a testing one with many tryouts)

I don't mind either way, really - whatever converges faster is good IMHO.

examples/agent_services_example.py
```diff
@@ -0,0 +1,119 @@
+#!/usr/bin/env python3
+
+__copyright__ = 'Copyright 2013-2014, http://radical.rutgers.edu'
```
andre-merzky, line 3: Date is wrong

examples/agent_services_example.py
```diff
@@ -0,0 +1,119 @@
...
+        pd_init = {'resource'      : resource,
+                   'runtime'       : 30,  # pilot runtime (min)
+                   'exit_on_error' : True,
+                   'project'       : config.get('project'),
+                   'queue'         : config.get('queue'),
+                   'access_schema' : config.get('schema'),
+                   'cores'         : config.get('cores', 1),
+                   'gpus'          : config.get('gpus',  0),
+                   # TODO create shell script
+                   'services'       :[rp.TaskDescription({'executable':'free -h'}),rp.TaskDescription({'executable':'free -h'}) ]
```
andre-merzky, line 55: max line length: 80 chars

examples/agent_services_example.py
```diff
@@ -0,0 +1,119 @@
+#!/usr/bin/env python3
```
andre-merzky, line 1: The name of the file does not need `example` - it is in the `examples` dir after all.

examples/misc/agent_services.py
```diff
@@ -0,0 +1,100 @@
+#!/usr/bin/env python3
```
andre-merzky, line 1: This file is not used AFAICT?

src/radical/pilot/agent/agent_0.py
```diff
@@ -6,11 +6,12 @@
 import os
 import pprint
 import stat
+import threading
 import time
 
 import radical.utils       as ru
 
-from ..   import utils     as rpu
+from .. import utils as rpu
```
andre-merzky, line 14: why this change?

src/radical/pilot/agent/agent_0.py
```diff
@@ -41,22 +43,22 @@ class Agent_0(rpu.Worker):
...
-        self._cfg     = cfg
+        self._cfg: ru.Config = cfg
         self._pid     = cfg.pid
         self._pmgr    = cfg.pmgr
         self._pwd     = cfg.pilot_sandbox
+
         self._session = session
         self._log     = ru.Logger(self._uid, ns='radical.pilot')
-
+        self._log.info("LoggingConfig %s",self._cfg)
```
andre-merzky, line 53: The cfg is too large to log every time.

src/radical/pilot/agent/agent_0.py
```diff
@@ -41,22 +43,22 @@ class Agent_0(rpu.Worker):
...
         self._starttime   = time.time()
         self._final_cause = None
 
         # this is the earliest point to sync bootstrap and agent profiles
         self._prof = ru.Profiler(ns='radical.pilot', name=self._uid)
         self._prof.prof('hostname', uid=cfg.pid, msg=ru.get_hostname())
 
-        # run an inline registry service to share runtime config with other
-        # agents and components
+        # run an inline registry service to share runtime config with other agents and components
```
andre-merzky, line 61: line too long

src/radical/pilot/agent/agent_0.py
```diff
@@ -82,8 +84,12 @@ def __init__(self, cfg, session):
...
         self._cmgr.start_components()
 
-        # start any services if they are requested
-        self._start_services()
+        # uid of service tasks
+        self._service_task_ids = []
+        # Event to handle for services started
+        self.services_event = threading.Event()
+        # currently running services
+        self.running_services = []
```
andre-merzky, line 92: Comments should be meaningful and provide information which is not obvious.  This is not a good comment:
```py
self.running_services = ...  # running services
```
The purpose is clear in the var name (as it should be).

Please use a leading underscore for variables which are private.

src/radical/pilot/agent/agent_0.py
```diff
@@ -340,77 +350,59 @@ def _start_services(self):
...
+            task['session_sandbox'] = self._cfg.session_sandbox
+            task['resource_sandbox'] = self._cfg.resource_sandbox
+            task['pilot'] = self._cfg.pid
+            # task['resources'] = {'cpu': td['ranks'] *
+            #                             td.get('cores_per_rank', 1),
+            #                      'gpu': td['ranks'] *
+            #                             td.get('gpus_per_rank', 1)}
+            services.append(task)
+        self.number_of_services_to_launch = len(services)
+        self.advance(services, publish=False, push=True)
```
andre-merzky, line 379: please check the coding guidelines for consistent indentation and spacing.

src/radical/pilot/agent/agent_0.py
```diff
@@ -340,77 +350,59 @@ def _start_services(self):
...
+            task['uid'] = ru.generate_id('services.%(item_counter)04d',
+                                             ru.ID_CUSTOM, ns="TODO_WHATTOUSE")
+            self._service_task_ids.append(task['uid'])
+            task['pilot_sandbox'] = self._cfg.pilot_sandbox
+            task['task_sandbox'] = self._cfg.pilot_sandbox + task['uid'] + '/'
+            task['task_sandbox_path'] = self._cfg.pilot_sandbox + task['uid'] + '/'
+            task['session_sandbox'] = self._cfg.session_sandbox
+            task['resource_sandbox'] = self._cfg.resource_sandbox
+            task['pilot'] = self._cfg.pid
+            # task['resources'] = {'cpu': td['ranks'] *
```
andre-merzky, line 373: Resources are needed for post-mortem statistics - why were they commented out?

src/radical/pilot/agent/agent_0.py
```diff
@@ -340,77 +350,59 @@ def _start_services(self):
...
+
+        for service_desc in service_descriptions:
+            task = dict()
+            task['origin'] = 'services'
+            task['description'] = TaskDescription(service_desc).as_dict()
+            task['state'] = rps.AGENT_STAGING_INPUT_PENDING
+            task['status'] = 'NEW'
+            task['type'] = 'task'
+            task['uid'] = ru.generate_id('services.%(item_counter)04d',
+                                             ru.ID_CUSTOM, ns="TODO_WHATTOUSE")
```
andre-merzky, line 365: `ns` should be the session ID (the uid is supposed to be unique in the scope of this session).

src/radical/pilot/agent/agent_0.py
```diff
@@ -340,77 +350,59 @@ def _start_services(self):
         If a `./services` file exist, reserve a compute node and run that file
         there as bash script.
         '''
+        self._log.info("Starting the agent services")
+        service_descriptions = self._cfg.services
+        services = list()
+
+        for service_desc in service_descriptions:
+            task = dict()
+            task['origin'] = 'services'
```
andre-merzky, line 359: `origin` of this service task is the agent.

src/radical/pilot/agent/agent_0.py
```diff
@@ -340,77 +350,59 @@ def _start_services(self):
...
+            task['description'] = TaskDescription(service_desc).as_dict()
+            task['state'] = rps.AGENT_STAGING_INPUT_PENDING
+            task['status'] = 'NEW'
+            task['type'] = 'task'
```
andre-merzky, line 363: please use `service_task` for task type.

src/radical/pilot/pilot_description.py
```diff
@@ -8,6 +8,8 @@
 # ------------------------------------------------------------------------------
 # Attribute description keys
 #
+from radical.pilot import TaskDescription
+
```
andre-merzky, line 12: please use relative imports to avoid circular imports.

src/radical/pilot/pilot_description.py
```diff
@@ -220,7 +222,7 @@ class PilotDescription(ru.TypedDict):
         OUTPUT_STAGING  : [str]      ,
         PREPARE_ENV     : {str: None},
         LAYOUT          : None       ,
-        SERVICES        : [str]      ,
+        SERVICES        : [TaskDescription]      ,
```
andre-merzky, line 225: spaces

src/radical/pilot/pmgr/launching/base.py
```diff
@@ -881,6 +881,7 @@ def _prepare_pilot(self, resource, rcfg, pilot, expand, tar_name):
         agent_cfg['task_post_exec']      = task_post_exec
         agent_cfg['resource_cfg']        = copy.deepcopy(rcfg)
         agent_cfg['debug']               = self._log.getEffectiveLevel()
+        agent_cfg['services']            = services
```
andre-merzky, line 884: please check this file again.  `services` is defined in two different places and is used in two different places.

andre-merzky (changes requested):
added comments.

examples/misc/agent_services.py
```diff
@@ -0,0 +1,100 @@
...
+import logging
+
+
+logger = ru.Logger('agent_services')
+
+
+
+def get_system_info():
+    try:
+        info = {}
```
andre-merzky, line 25: please use 
```py
info = dict()
```
here (and similar in other places).  Python's short nomenclature is harder to parse (for humans) and is also not consistent.  For example:
```py
>>> type([])
<class 'list'>
>>> type([2])
<class 'list'>

>>> type(())
<class 'tuple'>
>>> type((2))
<class 'int'>
```

examples/agent_services_example.py
kartikmodi: corrected

examples/agent_services_example.py
kartikmodi: Fixed

examples/agent_services_example.py
kartikmodi: fixed

examples/misc/agent_services.py
kartikmodi: removing for now

src/radical/pilot/agent/agent_0.py
kartikmodi: reverted

src/radical/pilot/agent/agent_0.py
kartikmodi: yes

src/radical/pilot/agent/agent_0.py
kartikmodi: done

src/radical/pilot/pilot_description.py
kartikmodi: Added spaces if that's what you mean

src/radical/pilot/pmgr/launching/base.py
kartikmodi: I check but did not find any specific problem. Can you help more here

examples/misc/agent_services.py
kartikmodi: Yes agree. But now file removed.

src/radical/pilot/pmgr/launching/base.py
andre-merzky: Check the occurence of `services` in that file - there should be 4 matching lines.  Do you need all of them?  I think you need 2.

examples/agent_services_example.py
andre-merzky: Changed, not fixed - still longer than 80 chars (some other lines in the file, too).

src/radical/pilot/pilot_description.py
kartikmodi: ok

src/radical/pilot/agent/agent_0.py
kartikmodi: made to debug. lmk if removing is required

src/radical/pilot/agent/agent_0.py
kartikmodi: fixed

src/radical/pilot/agent/agent_0.py
kartikmodi: added _

src/radical/pilot/agent/agent_0.py
kartikmodi: got it

src/radical/pilot/agent/agent_0.py
kartikmodi: ok

src/radical/pilot/agent/agent_0.py
kartikmodi: fixed

src/radical/pilot/agent/agent_0.py
kartikmodi: they use td instance to get those params, I don't know where to get those. 

```{'cpu': 1, 'gpu': 0}, ```
this is `resources` key's value  when I don't add myself

File: examples/agent_services_example.py (hunk 26 above)
Comment (andre-merzky) [reply examples/agent_services_example.py]: Sorry, that comment was on the wrong change set...

File: examples/agent_services_example.py (hunk 25 above)
Comment (andre-merzky) [reply examples/agent_services_example.py]: Changed, not fixed - still longer than 80 chars (some other lines in the file, too).

File: src/radical/pilot/agent/agent_0.py (hunk 29 above)
Comment (andre-merzky) [reply src/radical/pilot/agent/agent_0.py]: The log is stored in the filesystem anyway (`agent.0.cfg`), so logging it here is not needed.

File: src/radical/pilot/agent/agent_0.py (hunk 33 above)
Comment (andre-merzky) [reply src/radical/pilot/agent/agent_0.py:374]: Well, you can use the service's task description, right?  Or am I missing something?

File: src/radical/pilot/pilot_description.py (hunk 37 above)
Comment (andre-merzky) [reply src/radical/pilot/pilot_description.py:12]: Sorry, seems I wasn't clear.

Assume have this in `radical/pilot/__init__.py`:
```py
from .pilot_description import PilotDescription
```

and in `radical/pilot/pilot_description.py` we have something like
```py
import radical.pilot
```

then `import radical.pilot` will import `__init__.py` which will import `pilot_description.py` which will import `__init__.py` which will...  you get the picture.

To resolve that, please use this:
```py
from .task_description import TaskDescription
```

You will find that pattern throughout the code as well.

src/radical/pilot/agent/agent_0.py
kartikmodi: So should I add these values in my examples/agent_services.py

src/radical/pilot/pilot_description.py
kartikmodi: got it. Cyclic dep

src/radical/pilot/pmgr/launching/base.py
kartikmodi: still didn't get that properly. 
Do you think I would not need this `agent_cfg['services']            = services` ?

src/radical/pilot/agent/agent_0.py
andre-merzky: Default values should kick in if they are not specified in the original  service description.

src/radical/pilot/pmgr/launching/base.py
andre-merzky: Ok, let's do step by step:

> Check the occurence of services in that file

Can you paste the lines you find?

mturilli:
Need updating/writing tests for new methods/classes.

src/radical/pilot/pmgr/launching/base.py
kartikmodi: `SERVICES          = 'services'`

`SERVICES        : [TaskDescription],`

`        SERVICES        : []         ,`

I don't see a problem here, I am sure I am missing something very basic here. 
As per my understanding `SERVICES` var is defined in only 1 place, and used in 2 places

andre-merzky: I think you are looking in the wrong file - this looks different for me:

```sh
$ git pull
Already up to date.

$ grep services src/radical/pilot/pmgr/launching/base.py 
        services         = pilot['description']['services']
        services               += rcfg.get('services', [])
        for arg in services       :   bs_args.extend(['-j', arg])
        agent_cfg['services']            = services
```

Services are pulled from the pilot description, which is correct, but are also still pulled from the resource config which I would think we don't need anymore?  But, more importantly, they are passed via the resource config, which is ok, but still also passed as bootstrapper arguments, which does not make sense at all.

kartikmodi: ok now it get. I have removed the argument -j thing, but kept the resource config way and description.

src/radical/pilot/agent/agent_0.py
```diff
@@ -340,77 +351,59 @@ def _start_services(self):
...
-        self._log.debug('services started done')
+        self._log.info('received services for launching: %s', tasks)
 
+        if cmd == 'update':
+            for service in ru.as_list(tasks):
+                if service['uid'] in self._service_task_ids:
+                    self._log.info('service task has come up %s', service)
+                    if service['state'] == rps.AGENT_EXECUTING:
+                        self._running_services.append(service['uid'])
+                        self._log.info('Number of running services %s %s', len(self._running_services),
```
kartikmodi, line 401: use template from mikhail. remove other logs

src/radical/pilot/agent/agent_0.py
```diff
@@ -340,77 +351,59 @@ def _start_services(self):
...
+            #                      'gpu': td['ranks'] *
+            #                             td.get('gpus_per_rank', 1)}
+            self._service_task_ids.append(task['uid'])
+            services.append(task)
+        self.number_of_services_to_launch = len(services)
+        self.advance(services, publish=False, push=True)
+        # Waiting 2mins for all services to launch
+        self._log.info("Waiting for the agent services to get started")
+        did_timed_out = self.services_event.wait(timeout=60 * 2)
+        self._log.info("All agent services started")
```
kartikmodi, line 384: move down

src/radical/pilot/agent/agent_0.py
```diff
@@ -340,77 +351,59 @@ def _start_services(self):
...
+        services = list()
+
+        for service_desc in service_descriptions:
+            task = dict()
+            task['origin']            = 'agent'
+            task['description']       = TaskDescription(service_desc).as_dict()
+            task['state']             = rps.AGENT_STAGING_INPUT_PENDING
+            task['status']            = 'NEW'
+            task['type']              = 'service_task'
+            task['uid']               = ru.generate_id('services.%(item_counter)04d',
```
kartikmodi, line 365: services -> service

src/radical/pilot/agent/agent_0.py
```diff
@@ -340,77 +351,59 @@ def _start_services(self):
...
+        self.advance(services, publish=False, push=True)
+        # Waiting 2mins for all services to launch
+        self._log.info("Waiting for the agent services to get started")
+        did_timed_out = self.services_event.wait(timeout=60 * 2)
+        self._log.info("All agent services started")
+        if not did_timed_out:
+            raise RuntimeError("Unable to start services")
+
+    def _state_cb_of_services(self, topic, msg):
+        self._log.info("Callback of services called with topic %s",topic)
```
kartikmodi, line 389: use single quotes

src/radical/pilot/task_description.py
```diff
@@ -631,7 +632,7 @@ def _verify(self):
         if not self.get('mode'):
             self['mode'] = TASK_EXECUTABLE
 
-        if self.mode in [TASK_EXECUTABLE, RAPTOR_MASTER, RAPTOR_WORKER]:
+        if self.mode in [TASK_EXECUTABLE, RAPTOR_MASTER, RAPTOR_WORKER, AGENT_SERVICES]:
```
kartikmodi, line 635: remove space

src/radical/pilot/agent/agent_0.py
```diff
@@ -95,6 +102,8 @@ def __init__(self, cfg, session):
         rpu.Worker.__init__(self, self._cfg, session)
 
         self.register_subscriber(rpc.CONTROL_PUBSUB, self._check_control)
+        self._log.info("Registering the callback for agent services")
```
kartikmodi, line 105: remove this

src/radical/pilot/agent/agent_0.py
```diff
@@ -41,13 +42,14 @@ class Agent_0(rpu.Worker):
...
-        self._cfg     = cfg
+        self._cfg: ru.Config = cfg
         self._pid     = cfg.pid
         self._pmgr    = cfg.pmgr
         self._pwd     = cfg.pilot_sandbox
+
         self._session = session
         self._log     = ru.Logger(self._uid, ns='radical.pilot')
-
+        self._log.debug("LoggingConfig %s",self._cfg)
```
kartikmodi, line 52: Remove this

src/radical/pilot/agent/agent_0.py
```diff
@@ -41,13 +42,14 @@ class Agent_0(rpu.Worker):
     def __init__(self, cfg, session):
 
         self._uid     = 'agent.0'
-        self._cfg     = cfg
+        self._cfg: ru.Config = cfg
```
kartikmodi, line 45: @andre-merzky Should we start doing hinting? Or should I remove it? 
Or Should I do hinting in method signature?

andre-merzky: We are not consistent with type hints - but if you use them, then yes, the method signature would be the preferred place to do so.

src/radical/pilot/agent/agent_0.py
```diff
@@ -340,76 +349,53 @@ def _start_services(self):
...
+            task['type']              = 'service_task'
+            task['uid']               = ru.generate_id('service.%(item_counter)04d',
+                                             ru.ID_CUSTOM, ns=self._cfg.sid)
+            task['pilot_sandbox']     = self._cfg.pilot_sandbox
+            task['task_sandbox']      = self._cfg.pilot_sandbox + task['uid'] + '/'
+            task['task_sandbox_path'] = self._cfg.pilot_sandbox + task['uid'] + '/'
+            task['session_sandbox']   = self._cfg.session_sandbox
+            task['resource_sandbox']  = self._cfg.resource_sandbox
+            task['pilot']             = self._cfg.pid
+            self._service_task_ids.append(task['uid'])
```
andre-merzky, line 371: A recent commit removed the (commented) `task['resources']` entry.  We will need that for post mortem run analysis - please re-add and uncomment.

kartikmodi: Yes, I did remove it. Didn't realize it.

tests/unit_tests/test_agent_0/test_agent_0.py
```diff
@@ -206,13 +206,12 @@ def test_state_cb_of_services(self, mocked_init):
         agent_0.services_event = threading.Event()
 
         topic = 'test_topic'
-        msg = {'cmd':'update','arg':[{'uid':'101','state': 'AGENT_EXECUTING'}]}
+        msg = {'cmd':'update', 'arg': [{'uid':'101','state': 'AGENT_EXECUTING'}]}
```
andre-merzky: Still a space missing before `'state'`

src/radical/pilot/agent/agent_0.py
kartikmodi: I am taking `service_desc['ranks']` without default values. Is it assured that this key is always present ?

andre-merzky: Yes, that is ensured by the `TaskDescription` class.

tests/unit_tests/test_agent_0/test_agent_0.py
kartikmodi: Added

examples/agent_services_example.py
kartikmodi: Yes. I have now configure hard wrap at 80. So this will never occur again.

src/radical/pilot/agent/agent_0.py
andre-merzky: I just realized that you pull `ranks` from the `service_description` dict - sorry for missing your point.  You probably want to pull it from the `TakDescription` instance you create a couple of lines above.  The whole purpose of that step is to fill defaults and do sanity checks (otherwise we could just pass the original dict).  So once that description is created, please use it for all look-ups, not the original `service_desc` dict.

kartikmodi: Yes now i get. From the start I had this confusion within me. Thanks for clarifying.

andre-merzky (approved):
LGTM

kartikmodi:
> LGTM

Ohk Thanks

src/radical/pilot/agent/agent_0.py
```diff
@@ -340,75 +349,74 @@ def _start_services(self):
...
+            task['session_sandbox']   = cfg.session_sandbox
+            task['resource_sandbox']  = cfg.resource_sandbox
+            task['pilot']             = cfg.pid
+            task['resources']         = {'cpu': td.ranks * td.cores_per_rank,
+                                         'gpu': td.ranks * td.gpus_per_rank}
+
+            self._service_task_ids.append(tid)
+            services.append(task)
+
+        self.number_of_services_to_launch = len(services)
```
mtitov, line 382: this is not needed and is not used later

tests/unit_tests/test_agent_0/test_agent_0.py
```diff
@@ -165,14 +168,71 @@ def check_agent_task(agent_task, *args, **kwargs):
...
+
+        topic = 'test_topic'
+        msg   = {'cmd': 'update',
+                 'arg': [{'uid'  : '101',
+                          'state': 'AGENT_EXECUTING'}]}
+
+        agent_0._state_cb_of_services(topic, msg)
+        agent_0._running_services.append('102')
+
+        self.assertTrue(agent_0.services_event.is_set())
```
mtitov, line 225: you shouldn't change `agent_0._running_services` on your own, the method should handle that, e.g., you have an empty `_running_services`, then you provide `msg` with 1st task, and then `msg` with the 2nd task

tests/unit_tests/test_agent_0/test_agent_0.py
```diff
@@ -165,14 +168,71 @@ def check_agent_task(agent_task, *args, **kwargs):
...
+        agent_0.services_event = mock.Mock()
+        test_data = [dict({'cores_per_rank': '3', 'executable':'/bin/ls'})]
+        agent_0._cfg = ru.Config(from_dict={'pid':12, 'pilot_sandbox':'/'})
+        agent_0._cfg.services = test_data
+
+        agent_0.services_event.wait = mock.Mock(return_value=True)
+        agent_0._start_services()
+        self.assertTrue(advanced_services[0]['uid'].startswith('service.'))
+        self.assertTrue(test_data[0].items() <=
+                        advanced_services[0]['description'].items())
```
mtitov, line 198: the point of this is not clear, better you can just compare known attributes, e.g., `cores_per_rank` and `executable`, and maybe check that some default attributes are set (e.g., `gpus_per_rank=0`, `ranks=1`)

src/radical/pilot/agent/agent_0.py
kartikmodi: Fixed

andre-merzky:
I will look into the analytics issue - but otherwise this PR is done.  Thanks @kartikmodi !

andre-merzky:
> I will look into the analytics issue - but otherwise this PR is done. Thanks @kartikmodi !

Sorry, missed the last comments from @mtitov - please take care of those still @kartikmodi  - thanks!

src/radical/pilot/agent/agent_0.py
```diff
@@ -340,75 +349,74 @@ def _start_services(self):
         If a `./services` file exist, reserve a compute node and run that file
         there as bash script.
         '''
+        self._log.info('Starting the agent services')
+        service_descriptions = self._cfg.services
```
mtitov, line 353:
```python
if not service_descriptions:
     return
```

mtitov: and move `self._log.info('Starting the agent services')` after it

mtitov: and remove docstring

tests/unit_tests/test_agent_0/test_agent_0.py
kartikmodi: yes agreed

tests/unit_tests/test_agent_0/test_agent_0.py
kartikmodi: The test_data contains `cores_per_rank` and `executable`. As far as I know, `cores_per_rank` is a default attribute.
I am checking both keys and values.

src/radical/pilot/agent/agent_0.py
kartikmodi: fixed

tests/unit_tests/test_agent_0/test_agent_0.py
kartikmodi: @mtitov ? Can you update here

kartikmodi:
@andre-merzky I think it's ready to be merged. I'll merge Monday if nothing comes up before that.

mtitov:
> @andre-merzky I think it's ready to be merged. I'll merge Monday if nothing comes up before that.

@kartikmodi we can discuss it during the call

mtitov (approved):
LGTM!

kartikmodi:
@andre-merzky 
```codecov/project Expected — Waiting for status to be reported
```
THis is stuck. what to do?

andre-merzky:
> THis is stuck. what to do?

Happens now and then - the PR is merged now.  Congrats! :-)

Issue #2652, opened 2022-07-25
Type error in master.py?


eirrgang:
https://github.com/radical-cybertools/radical.pilot/blob/5b704de493bfd4ad6180479a25436fbe1030869f/src/radical/pilot/raptor/master.py#L398 seems to show some sort of error. Immediately after assigning a list to `stats`, `stats` is subscripted as if it were a dictionary. Is this a bug or a dead code block? Or am I misreading it somehow?

andre-merzky:
Fixed by PR #2780

Issue #2766, opened 2022-11-07
`Master.wait()` is broken when `uid`s are specified


andre-merzky:
```
                n = stats['ACTIVE'] + stats['DONE'] + stats['FAILED']
```
should be 
```
                n = len(stats['ACTIVE']) + len(stats['DONE']) + len(stats['FAILED'])
```
Also, keys need to be checked also (or use `defaultdict(list)`)

eirrgang:
Duplicate of #2652 ?

andre-merzky:
An RP ticket is now open for this on radical-cybertools/radical.pilot/issues/2776

andre-merzky:
Fixed by PR #2780

PR #2780, opened 2022-11-28
fix worker state statistics

andre-merzky:


andre-merzky:
This PR fixes #2652 and #2766

mtitov:
@andre-merzky , have several comments

(1) I would propose to have one approach for both `count` and `uids` as following
```python
        if uids:
            uids = ru.as_list(uids)
            for uid in list(uids):
                if uid not in self._workers:
                    uids.remove(uid)
        else:
            uids = list(self._workers.keys())

        if not count or count > len(uids):
            count = len(uids)

        self._log.debug('wait for %d workers: %s', count, uids)

        while True:

            stats = defaultdict(int)
            with self._lock:
                for uid in uids:
                    stats[self._workers[uid]['status']] += 1

            n = sum([stats[state] for state in rps.FINAL])
            self._log.debug('stats [%d]: %s', n, stats)

            if n >= count:
                self._log.debug('wait ok')
                return
            time.sleep(1)
```

(2.A) there is inconsistency in the worker status name: in some places it is `self._workers[uid]['state']` and in some it is `self._workers[uid]['status']`

(2.B) Check worker's status values - there is one place with unused `ACTIVE` status - as I understand currently there are just two statuses `NEW` and `DONE`. Btw, let's also use constants states for that (i.e., `import radical.pilot.states as rps`)

(3) There are couple typos in `_control_cb` method - `if 'uid' not in self._workers`, should be not string, but `uid` variable
https://github.com/radical-cybertools/radical.pilot/blob/9606637d41235355ae61a3a81c15420368546edb/src/radical/pilot/raptor/master.py#L199-L202
https://github.com/radical-cybertools/radical.pilot/blob/9606637d41235355ae61a3a81c15420368546edb/src/radical/pilot/raptor/master.py#L213-L215

andre-merzky:
@mtitov : thanks, applied changes

mtitov (approved):
LGTM! Thank you!

src/radical/pilot/raptor/master.py
```diff
@@ -328,7 +334,7 @@ def submit_workers(self, descr: Dict[str, Union[str, int]],
                 task['origin']            = 'raptor'
                 task['description']       = TaskDescription(td).as_dict()
                 task['state']             = rps.AGENT_STAGING_INPUT_PENDING
-                task['status']            = 'NEW'
+                task['status']            = self.NEW
```
mtitov, line 337: sorry, missed this one, should it take constant value from `rp.states`? (even tho it has the same value as this one)

mtitov: nevermind, looks right!

PR #2788, opened 2022-12-01
update RM PBSPro

mtitov:


src/radical/pilot/agent/resource_manager/pbspro.py
```diff
@@ -20,35 +19,35 @@ class PBSPro(ResourceManager):
...
-        # TODO: $NCPUS?!?! = 1 on archer
+        try:
+            vnodes, rm_info.cores_per_node = self._parse_pbspro_vnodes()
+            nodes = [(node, rm_info.cores_per_node) for node in vnodes]
 
-        cpn = os.environ.get('NUM_PPN') or os.environ.get('SAGA_PPN')
-        if not cpn:
-            raise RuntimeError('$NUM_PPN and $SAGA_PPN not set!')
+        except (IndexError, ValueError) as exc:
+            self._log.debug_2('exec_vnodes not detected')
```
andre-merzky, line 27: What does `exec_vnodes` mean?  Maybe `cannot parse qstat output` or something like that?

src/radical/pilot/agent/resource_manager/pbspro.py
```diff
@@ -20,35 +19,35 @@ class PBSPro(ResourceManager):
...
 
-        pbspro_vnodes = self._parse_pbspro_vnodes()
+            if not rm_info.cores_per_node or 'PBS_NODEFILE' not in os.environ:
+                raise RuntimeError('resource configuration unknown') from exc
```
andre-merzky, line 30: Same here: not sure the error message fits the error.  I would not know what to look for reading that message (or would look in the wrong place, i.e., the resource configs).

andre-merzky (commented):
Overall looks good to me - but I left some comments on error handling.

src/radical/pilot/agent/resource_manager/pbspro.py
mtitov: with this I actually want to point to `exec_vnodes` (PBSPro specific attribute), since that is what we look for and `qstat` is just a tool, thus we reach this message not only if `qstat` failed, but also if `exec_vnodes` is failed to be parsed

andre-merzky: I actually don't know what `exec_vnodes` is.  from looking at the code I assume that's some qstat output field - but that may not be obvious to the user.

mtitov: btw, that is under `DEBUG_2`, kinda just for thorough debug, it is not needed for regular (`info`) and debug messages - extra info to let us know why it was switched from `_parse_pbspro_vnodes` to using `PBS_NODEFILE`

src/radical/pilot/agent/resource_manager/pbspro.py
mtitov: extended

src/radical/pilot/agent/resource_manager/pbspro.py
andre-merzky: Fair point...

andre-merzky (approved):
LGTM!

PR #2791, opened 2022-12-08
generic schema lookup

andre-merzky:
@mtitov: I this lookup the one you referred to?


andre-merzky:
> also need to update the line before it ([link](https://github.com/radical-cybertools/radical.pilot/blob/89e6c174104a0178d9d97224a9309e5f3a70cfda/src/radical/pilot/session.py#L99)) -> `for schema in rcfg.get('schemas', []):`

Ah, right, that was it - thanks! :-)

PR #2793, opened 2022-12-08
generic schema lookup

andre-merzky:
follow suggestions from @mtitov.  This is a re-instantiation of #2791 - the forced push to the branch closed the original PR.


mtitov (approved):
LGTM! Thanks!)

PR #2796, opened 2022-12-14
Fix-up Sphinx syntax.

eirrgang:
The directive corresponding to the `:py:attr:` role is `.. py:attribute::`, not `.. py:attr::`.

The fix was correctly described in #2757, but apparently I made an error in #2758. Now fixed.


eirrgang:
Oops! Sorry!

andre-merzky (approved):
Thanks Eric!

PR #2801, opened 2022-12-15
Feature/interactive amarel

andre-merzky:
Uh, I thought we had this merged already, but apparently... :-/


mtitov (approved):
LGTM!

one thing to confirm regarding this line `"default_remote_workdir" : "/scratch/$USER",` - do we have set `$USER` within the interactive mode? otherwise we can update it as `/scratch/${USER:-$SLURM_JOB_USER}`

PR #2805, opened 2023-01-04
fixed inconsistency in launching order if some LMs failed to be created

mtitov:
If some methods were not created due to exceptions then `self._launch_order` should be adjusted accordingly, otherwise method `find_launcher` could fail (if not created method will be checked first)

https://github.com/radical-cybertools/radical.pilot/blob/9c74ffbeb0d8033bc080facc888caa8b9a2d07c5/src/radical/pilot/agent/resource_manager/base.py#L418-L423

### DIFF
--- a/src/radical/pilot/agent/resource_manager/base.py
+++ b/src/radical/pilot/agent/resource_manager/base.py
@@ -320,30 +320,26 @@ class ResourceManager(object):
         launch_methods  = self._cfg.resource_cfg.launch_methods
 
         self._launchers    = {}
-        self._launch_order = None
+        self._launch_order = launch_methods.get('order') or list(launch_methods)
 
-        for name, lm_cfg in launch_methods.items():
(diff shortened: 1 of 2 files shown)


src/radical/pilot/agent/resource_manager/base.py
```diff
@@ -320,31 +320,27 @@ def _prepare_launch_methods(self, rm_info):
         launch_methods  = self._cfg.resource_cfg.launch_methods
 
         self._launchers    = {}
-        self._launch_order = None
+        self._launch_order = launch_methods.get('order') or list(launch_methods)
```
andre-merzky, line 323: I didn't even know that `list(some_dict)` gives  a list of keys - always used `list(some_dict.keys())` for that (which removes the iterator).  Thanks for that tidbit :-)  We should clean that up throughout the code when we stumble over it...

andre-merzky (approved):
Nice!

PR #2806, opened 2023-01-04
Modify resource_anl.json to support polaris machine on ANL.

GKNB:
Two new cases, polaris and polaris_interactive are added. These two are machines on ANL with pbspro rm. They are both tested with Tianle and Mikhail on polaris login node (for polaris) and polaris interactive node (for polaris_interactive).


andre-merzky (approved):
Thanks, looks good to me.  One question though: the used python presumably comes from `module load conda`.  Should `"python_dist": "anaconda"` be set also, or is that not needed because of the pre-set VE?

PR #2808, opened 2023-01-05
Fix/register mode

andre-merzky:
This PR fixes / enables the `register_mode` for MPI raptors.


examples/misc/raptor_worker.py
```diff
@@ -13,12 +13,38 @@ class MyWorker(rp.raptor.MPIWorker):
...
     '''
 
+    class MyMPIWorkerRank(rp.raptor.MPIWorkerRank):
+
+        def __init__(self, rank_task_q_get, rank_result_q_put, event,
+                           log, prof, base):
+            super().__init__(rank_task_q_get, rank_result_q_put, event,
+                             log, prof, base)
+
+            self.register_mode('foo', self._dispatch_foo)
```
mtitov, line 23: close to off topic - currently we don't limit modes that user can provide, but maybe to use one of supported task modes here just as part of the example? (just thinking out of load, it's ok to skip it)

mtitov (approved):
LGTM!

examples/misc/raptor_worker.py
andre-merzky: It's a fair point and I tried to do that - but to make this remotely useful would make the example too complex IMHO, so let's leave it if you don't mind...

eirrgang: I'm not sure I understand the suggestion. The supported modes are already registered in the source code. This part of this example is demonstrating extensibility. What is being left out? Overriding built-in modes?

andre-merzky: Yes.  Also, overloading a built-in mode this way would explain nicely how the build-in modes are implemented, thus hopefully providing a better understanding of the system.  But, as said, I think that this would be beyond the useful scope for an example (the raptor example is already way too bloated anyway).

eirrgang: > I think that this would be beyond the useful scope for an example (the raptor example is already way too bloated anyway)

It would also strengthen the raptor specification in a way you may not be ready for. Personally, I find it sufficient that a supported extension mechanism is documented by example, and it is easy enough to look at the source for other implementation details.

If you want to strongly specify a scheme for overriding built-in modes, then it could be documented either in docstrings or by-example (such as here). But first it should be tested.

PR #2816, opened 2023-01-13
Feature/raptor resilience

andre-merzky:
This PR reliably terminates a raptor MPI worker if any worker rank misses to send heartbeats for some time.  It also fixes a problem with the `popen` executor task cancellation which now will kill the executed process group (instead of only the execution script).  Finally, it changed the access to the master's `self._workers` data structures to atomic accesses, thus removing the need for locking that structure - that mitigates some of the performance penalty of introducing the heartbeat recording.


mturilli:
Should we close this as it is contained in #2819? Note that this is now behind compared to #2819 so careful! :)

andre-merzky:
> Should we close this as it is contained in #2819? Note that this is now behind compared to #2819 so careful! :)

Right - closed.

Issue #870, opened 2015-10-22
job not killed when agent launch fails because of missing sdists


leelasd:

andre-merzky:
Missing sdists should pop up as an error in the bootstrapper, and AFAICS the bootstrapper should fail then -- so I think your observation says that this is not what happens, but that the bootstrapper is continuing / hanging?

Would you be able to provide the bootstrap_1.out/err/log files from the pilot sandbox?  Thanks!

leelasd:
Snafu from Mark.

PR #2820, opened 2023-01-25
Set `blocked_cores/gpus` part of `system_architecture` in resource config

mtitov:
Parameter `system_architecture` is provided to SAGA and its sub-parameter `blocked_cores` is used to define a core specialization (currently with focus on SLURM adaptor - RS PR [#870](https://github.com/radical-cybertools/radical.saga/pull/870))

### DIFF
--- a/src/radical/pilot/agent/resource_manager/base.py
+++ b/src/radical/pilot/agent/resource_manager/base.py
@@ -217,14 +217,15 @@ class ResourceManager(object):
         rm_info.lfs_per_node     = self._cfg.lfs_size_per_node
         rm_info.lfs_path         = ru.expand_env(self._cfg.lfs_path_per_node)
 
-        rcfg = self._cfg.resource_cfg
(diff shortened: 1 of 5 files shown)


andre-merzky (approved):
Thanks @mtitov - that looks good to me. I am not sure about the `module load / unload` related changes but assume you know what you are doing there :-)

mtitov:
Andre, thank you! Along with your approval and before merging, I will test it on target machines.

>  I am not sure about the module load / unload related changes but assume you know what you are doing there :-)

yeah, that was part of the cleaning :) used these changes for the latest runs and moved it here while editing corresponding configurations

andre-merzky:
> > I am not sure about the module load / unload related changes but assume you know what you are doing there :-)
> 
> yeah, that was part of the cleaning :) used these changes for the latest runs and moved it here while editing corresponding configurations

Sounds good, thanks for the context!

mtitov:
tested on Crusher@OLCF (and confirmed changes on Summit@OLCF)

PR #2822, opened 2023-01-27
Fixed core-option in JSRun LM

mtitov:
Option `-c` is actually about physical cores - https://docs.olcf.ornl.gov/systems/summit_user_guide.html#controlling-number-of-threads-for-tasks, thus we have to calculate it from `cores_per_rank` while considering SMT level through `RADICAL_SMT`

### DIFF
--- a/src/radical/pilot/agent/launch_method/jsrun.py
+++ b/src/radical/pilot/agent/launch_method/jsrun.py
@@ -1,7 +1,9 @@
 
-__copyright__ = 'Copyright 2016-2022, The RADICAL-Cybertools Team'
+__copyright__ = 'Copyright 2016-2023, The RADICAL-Cybertools Team'
 __license__   = 'MIT'
 
+import math
+import os
 
 import radical.utils as ru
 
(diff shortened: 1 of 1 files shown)


src/radical/pilot/agent/launch_method/jsrun.py
```diff
@@ -160,12 +161,16 @@ def get_launch_cmds(self, task, exec_path):
                 self._create_resource_set_file(slots, uid, sbox)
 
         else:
+            # for OpenMP threads a corresponding parameter should be provided
+            # in task description - `td.threading_type = rp.OpenMP`,
+            # and RP will set `export OMP_NUM_THREADS=<cores_per_rank>`
+            cores_per_rs = math.ceil(
+                td['cores_per_rank'] / self._rm_info['threads_per_core'])
             # -b: bind to RS
```
AymenFJA, line 169: Can we explain what is `RS: Resource Sets` in here? So if someone reads the code can understand the acronym.

AymenFJA (approved):
Other than the comment above, it looks good to me.

PR #2824, opened 2023-01-31
Fix thread-safety of PilotManager staging operations.

### LINKED ISSUE #2823: Pilot data staging commands are not thread safe.
If `Pilot.stage_out()` / `Pilot.stage_in()` commands are issued from multiple threads at the same time, the locking in the PilotManager member functions is insufficient to protect the integrity of its internal state. The functions are not reentrant.

There are opportunities for the thread to yield while executing `_pilot_staging_output` or `_pilot_staging_input` and for another thread to replace the `_active_sds` member dictionary before the thread continues.

This allows the member functions to accidentally complete before the file staging has completed, which is clearly not the intent.

The issue should be easy to resolve by managing a single dictionary instance for the lifetime of the PilotManager (instead of replacing it), using just the `uids` for which each function invocation is responsible.

The issue was observed in RP 1.20.1 and is still present in the devel branch.

eirrgang:
The `_active_sds` dictionary member is created at initialization. To be reentrant, member functions that use the dictionary should modify or iterate on it only with `_sds_lock`, and should only add/remove entries relevant to the current function invocation.

Fixes #2823

### DIFF
--- a/src/radical/pilot/pilot_manager.py
+++ b/src/radical/pilot/pilot_manager.py
@@ -434,7 +434,6 @@ class PilotManager(rpu.Component):
         # prepare to wait for completion
         with self._sds_lock:
 
-            self._active_sds = dict()
             for sd in sds:
                 sd['state'] = rps.NEW
                 self._active_sds[sd['uid']] = sd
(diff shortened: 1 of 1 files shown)


eirrgang:
A (far-from-minimal) exhibition of the problem can be explored with https://github.com/SCALE-MS/scale-ms/commit/d8fb103af5fe87654f87ca4e9e47fda58a3c7a0e. Without the patch, we see output like https://github.com/SCALE-MS/scale-ms/actions/runs/4054243610/jobs/6975833535#step:8:97. With the patched RP, the `stage_out` commands correctly block until the staged directories exist locally.

src/radical/pilot/pilot_manager.py
```diff
@@ -944,4 +950,3 @@ def check_uid(self, uid):
 
 
 # ------------------------------------------------------------------------------
-
```
andre-merzky, line 947: We actually do empty lines at end of files.

andre-merzky (approved):
LGTM, thanks!

PR #2826, opened 2023-02-02
fixed `ranks_per_node` calculation (APRun LM)

mtitov:


andre-merzky (approved):
LGTM

mtitov:
@GKNB Tianle, can you please also confirm that this works for you?

GKNB (approved):
I tested it on theta machine and it works correctly

PR #2830, opened 2023-02-10
allow access to full node memory by default

andre-merzky:


src/radical/pilot/agent/launch_method/srun.py
```diff
@@ -159,6 +159,12 @@ def get_launch_cmds(self, task, exec_path):
             mapping += ' --threads-per-core %d' % \
                        self._rm_info['threads_per_core']
 
+        if td.get('mem_per_rank'):
+            mapping += ' --mem %s' % td.get('mem_per_rank')
+        else:
+            # allow access to full node memory by default
+            mapping += ' --mem 0'
```
mtitov, line 166: this would work only if task is allocated to a node exclusively, otherwise tasks, which share a node, will not be executed concurrently.

mtitov: or I'm confused here.. if `mem` set within the batch script then it is memory per node for a job, but within `srun` isn't it memory per node for a job step?

mtitov:
TBD: test on Crusher

src/radical/pilot/agent/launch_method/srun.py
andre-merzky: This is true for `srun` jobs, but the behavior is different for job steps.  From the [`srun` man pages](https://slurm.schedmd.com/srun.html#OPT_mem):

**_Specifying a memory limit of zero for a job step will restrict the job step to the amount of memory allocated to the job, but not remove any of the job's memory allocation from being available to other job steps._**

mtitov (approved):

PR #2832, opened 2023-02-17
Introduced GPU sharing for JSRun LM

mtitov:
- Updated slots setup within Scheduler when GPU sharing is on;
- Updated JSRun (including ERF) LM to consider GPU sharing.


mtitov:
TBD: Tests

mtitov:
Simple tests on Summit are successful (including JSRUN_ERF LM)

mtitov:
Will continue to work on expanding tests coverage, but PR itself is ready for review @andre-merzky

src/radical/pilot/agent/scheduler/continuous.py
```diff
@@ -270,8 +274,31 @@ def schedule_task(self, task):
...
             cores_per_slot = 1
 
-        self._log.debug_3('req : %s %s %s %s %s', req_slots, cores_per_slot,
-                        gpus_per_slot, lfs_per_slot, mem_per_slot)
+        # check if there is a GPU sharing
+        if gpus_per_slot and not gpus_per_slot.is_integer():
+            gpus           = td['ranks'] * gpus_per_slot
+            req_slots      = m.ceil(gpus / gpus_per_node)
+            gpus_per_slot  = int(gpus // req_slots)
+            ranks_per_slot = td['ranks'] // req_slots
```
andre-merzky, line 282: Ok, understood that a `slot` is not something which can hold a number of ranks so that all GPUs on those ranks are included.

But I don't think the math works out:
```py
import math as m

td = {'ranks'         : 9,
      'cores_per_rank': 1,
      'gpus_per_rank' : 1 / 3}

cores_per_node = 128
gpus_per_node  =   2


# from `schedule_task()`
cores_per_slot = td['cores_per_rank']
gpus_per_slot  = td['gpus_per_rank']

gpus           = td['ranks'] * gpus_per_slot
req_slots      = m.ceil(gpus / gpus_per_node)
gpus_per_slot  = int(gpus // req_slots)
ranks_per_slot = td['ranks'] // req_slots

# find the greatest common divisor
di = m.gcd(gpus_per_slot, ranks_per_slot)
req_slots       *= di
gpus_per_slot  //= di
ranks_per_slot //= di

print('req_slots     : ', req_slots)
print('ranks_per_slot: ', ranks_per_slot)
print('gpus_per_slot : ', gpus_per_slot)
```

which gives 
```sh
$ ./t.py
req_slots     :  2
ranks_per_slot:  4
gpus_per_slot :  1
```

I would have expected 
```sh
$ ./t.py
req_slots     :  3
ranks_per_slot:  3
gpus_per_slot :  1
```

Am I missing something?

src/radical/pilot/raptor/master.py
```diff
@@ -343,7 +346,8 @@ def submit_workers(self, descr: Dict[str, Union[str, int]],
             task['resource_sandbox']  = os.environ['RP_RESOURCE_SANDBOX']
             task['pilot']             = os.environ['RP_PILOT_ID']
             task['resources']         = {'cpu': td.ranks * td.cores_per_rank,
-                                         'gpu': td.ranks * td.gpus_per_rank}
+                                         'gpu': int(td.ranks *
+                                                    td.gpus_per_rank)}
```
andre-merzky, line 350: Do we need int here?

src/radical/pilot/raptor/master.py
```diff
@@ -574,8 +578,8 @@ def _submit_executable_tasks(self, tasks):
             task['pilot']             = os.environ['RP_PILOT_ID']
             task['resources']         = {'cpu': td['ranks'] *
                                                 td.get('cores_per_rank', 1),
-                                         'gpu': td['ranks'] *
-                                                td.get('gpus_per_rank', 0)}
+                                         'gpu': int(td['ranks'] *
+                                                    td.get('gpus_per_rank', 0))}
```
andre-merzky, line 582: same

src/radical/pilot/agent/scheduler/continuous.py
```diff
@@ -215,13 +215,14 @@ def _find_resources(self, node, find_slots, cores_per_slot, gpus_per_slot,
...
-            gpu_map  = [[gpu] for gpu in gpus]
+            cores_per_rank = cores_per_slot // ranks_per_slot
+            core_map = [cores[i:i + cores_per_rank]
+                        for i in range(0, len(cores), cores_per_rank)]
 
             slots.append({'node_name': node_name,
                           'node_id'  : node_id,
                           'core_map' : core_map,
-                          'gpu_map'  : gpu_map,
+                          'gpu_map'  : gpus,
```
andre-merzky, line 225: I would prefer if we keep the 'map' semantics for the `gpu_map`, in that is maps a rank to a specific GPU.  That would simplify the changes in the executor code (I think it would not really need changes?).  With that I mean that we map multiple ranks to the same gpu.  So, instead of 
```py
cpu_map = [[0, 1], [2, 3], [4, 5], [6, 7]]
gpu_map = [0, 1]
```
we would have
```py
cpu_map = [[0, 1], [2, 3], [4, 5], [6, 7]]
gpu_map = [[0], [0], [1], [1]]
```
(the double nesting is needed for multiple gpus per rank).  That way we keep the route open for heterogeneous ranks, otherwise we would need to change the executor again later.  Is there a good reason to change that data structure and the executor?

src/radical/pilot/agent/executing/popen.py
```diff
@@ -701,10 +701,15 @@ def _extend_pre_exec(self, td, ranks=None):
 
         if td['gpus_per_rank'] and td['gpu_type'] == rpc.CUDA and ranks:
             # equivalent to the 'physical' value for original `cvd_id_mode`
-            td['pre_exec'].append(
-                {str(rank_id): 'export CUDA_VISIBLE_DEVICES=%s' %
-                               ','.join([str(gm[0]) for gm in rank['gpu_map']])
-                 for rank_id, rank in enumerate(ranks)})
+            rank_id  = 0
+            rank_env = {}
+            for ranks_per_slot in ranks:
```
andre-merzky, line 706: I am confused with naming.  I would think that `ranks_per_slot` is a number, like `2`.  And the `core_map` should contain the mapping of cores to ranks.  I don't think the executor needs to know about the details of the `ranks_per_slot` the schedule had to deal with, does it?

src/radical/pilot/agent/launch_method/jsrun.py
```diff
@@ -121,20 +121,26 @@ def _create_resource_set_file(self, slots, uid, sandbox):
...
-                rs_str += 'rank: %d: {' % rank_id
-                rs_str += ' host: %s;'  % str(rank['node_id'])
-                rs_str += ' cpu: {%s}'  % cores
-                if gpu_maps:
-                    gpus = [str(gpu_map[0]) for gpu_map in gpu_maps]
-                    rs_str += '; gpu: {%s}' % ','.join(gpus)
-                rs_str  += ' }\n'
-                rank_id += 1
+        base_id = 0
+        for ranks_per_slot in slots['ranks']:
```
andre-merzky, line 125: same comment as above

src/radical/pilot/agent/launch_method/mpirun.py
```diff
@@ -152,6 +152,10 @@ def get_launch_cmds(self, task, exec_path):
         td         = task['description']
         sandbox    = task['task_sandbox_path']
         task_cores = td.get('cores_per_rank', 1)
+        task_gpus  = td.get('gpus_per_rank', 0.)
+        # sharing GPUs among multiple ranks not supported
```
andre-merzky, line 156: Why can't we support this actually?  Ultimately it means that we would use the same GPU ID for multiple ranks - or is there more to it?  I don't think `mpirun` needs to be aware of how GPUs are allocated and used at all.

andre-merzky (commented):
Sorry if the comments are out of order - I jumped back and forth a lot.

src/radical/pilot/agent/scheduler/base.py
```diff
@@ -448,15 +448,13 @@ def _change_slot_states(self, slots, new_state):
                 raise RuntimeError('inconsistent node information')
 
             # iterate over cores/gpus in the slot, and update state
-            cores = rank['core_map']
-            for cslot in cores:
-                for core in cslot:
+            slot_ranks = rank['core_map']
```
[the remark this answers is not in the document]
mtitov: I kinda felt that this name fits - by default each slot represents one rank, but in case of shared gpu(s) slot will represent multiple ranks, i.e. ranks per slot

src/radical/pilot/agent/scheduler/base.py
```diff
@@ -753,10 +751,8 @@ def _schedule_waitpool(self):
         for task in scheduled:
             td = task['description']
             task['$set']      = ['resources']
-            task['resources'] = {'cpu': td['ranks'] *
-                                        td['cores_per_rank'],
-                                 'gpu': td['ranks'] *
-                                        td['gpus_per_rank']}
+            task['resources'] = {'cpu': td['ranks'] * td['cores_per_rank'],
+                                 'gpu': int(td['ranks'] * td['gpus_per_rank'])}
```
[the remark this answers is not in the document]
mtitov: honestly, I decided to keep the type as it was (and since the number of gpus is not a fraction, thus to ensure it)

src/radical/pilot/agent/scheduler/continuous.py
```diff
@@ -126,8 +126,8 @@ def unschedule_task(self, task):
 
     # --------------------------------------------------------------------------
     #
-    def _find_resources(self, node, find_slots, cores_per_slot, gpus_per_slot,
-                              lfs_per_slot, mem_per_slot, partial):
+    def _find_resources(self, node, find_slots, ranks_per_slot, cores_per_slot,
+                        gpus_per_slot, lfs_per_slot, mem_per_slot, partial):
```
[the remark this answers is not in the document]
mtitov: yes, this was a tricky part - at this point slot is what handles CPUs and GPUs as a whole, not their fractions - for all LMs `slot == rank`, for JSRun LM `slot == resource set`, thus slot might contain multiple ranks, IF that ranks share GPUs

src/radical/pilot/agent/scheduler/continuous.py
```diff
@@ -215,13 +215,14 @@ def _find_resources(self, node, find_slots, cores_per_slot, gpus_per_slot,
                     gpus.append(gpu_idx)
                 gpu_idx += 1
 
-            core_map = [cores]
-            gpu_map  = [[gpu] for gpu in gpus]
+            cores_per_rank = cores_per_slot // ranks_per_slot
```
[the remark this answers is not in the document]
mtitov: it is actually not the case, because the way `cores_per_slot` is calculated earlier: `cores_per_slot  *= ranks_per_slot`

src/radical/pilot/agent/scheduler/continuous.py
```diff
@@ -215,13 +215,14 @@ def _find_resources(self, node, find_slots, cores_per_slot, gpus_per_slot,
...
+            core_map = [cores[i:i + cores_per_rank]
+                        for i in range(0, len(cores), cores_per_rank)]
```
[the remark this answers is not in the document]
mtitov: both would work, but in this case would need to ensure everywhere that it is a list of tuples

src/radical/pilot/agent/scheduler/continuous.py
mtitov: > That way we keep the route open for heterogeneous ranks, otherwise we would need to change the executor again later.

how it would change?

without double nesting, it actually says that all these gpus are per slot, thus if there are multiple ranks per slot, then all these ranks share all gpus, in your case you assign one gpu per rank within the slot, but in case it is

```python
cpu_map = [[0, 1], [2, 3], [4, 5], [6, 7]]
gpu_map = [[0, 1], [0, 1], [0, 1], [0, 1]]
```

which looks messy, that's why I though to follow `gpu_map = [0, 1]` and it is clear it is per slot (which is either 1 rank or multi-ranks)

src/radical/pilot/agent/executing/popen.py
mtitov: should I rename to `slot_ranks` (as in scheduler)?

src/radical/pilot/agent/launch_method/mpirun.py
mtitov: that statement is to our LM, and not to `mpirun` itself, and that just a follow-up after our discussion, as we wanted to allow GPU sharing for JSRun LM only (at least for now)

src/radical/pilot/agent/executing/popen.py
mtitov: > I don't think the executor needs to know about the details of the ranks_per_slot the schedule had to deal with, does it?

it kinda does, since it sets this env variable per rank

src/radical/pilot/agent/launch_method/jsrun.py
mtitov: JSRun definitely should know and understand the mapping, since it maps to "resource set" concept. Ok to rename `ranks_per_slot` if it brings some confusion

src/radical/pilot/agent/scheduler/continuous.py
mtitov: yeah, need to polish this part, thank you for catching it

src/radical/pilot/agent/scheduler/base.py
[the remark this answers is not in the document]
mtitov: with a quick look, this data (`resources`) is collected for and used by analytics only, we might ensure (type `int`) on analytics side only and relax types here (keeping either `int` or `float`).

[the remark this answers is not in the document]
andre-merzky: Yes, that's what I thought also.

src/radical/pilot/agent/scheduler/continuous.py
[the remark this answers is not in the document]
andre-merzky: Well, slot is somewhat defined for scheduling.  We use the DRMAA definition really, but it's used [elsewhere](https://blogs.cisco.com/performance/hpc-schedulers-what-is-a-slot) too.  In that context a slot really refers to either a full job or to a single rank, so I am bit hesitant to change that notion.  But that's just a concern about terminology, not about the overall approach.

src/radical/pilot/agent/scheduler/continuous.py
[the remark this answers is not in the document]
andre-merzky: But better readability would be nice ;-)  So: does it actually matter if it is a list of tuples?  They get iterated the same way really.  If so, we can either change RU or do a conversion here (`[list(m) for m in core_map]`).

src/radical/pilot/agent/scheduler/continuous.py
andre-merzky: > which looks messy

Why does that look messy to you?  It is nicely symmetric, isn't it?  I don't see a reason to have different data structures for gpus and cpus (or any other discreet resources) really.

Assume you have a task description `gpus_per_rank = 1.5` (which is valid now), and we have 6 ranks.  That in the new notion would result in 3 slots, each allocating 3 gpus.  Each slot would then be something like:

```py
cpu_map = [[0], [1]]
gpu_map = [0, 1, 2]
```

where each rank uses 50% of the specified GPU.   But an equally valid schedule could be that the first rank gets gpu 0, second rank gets gpu 2, and gpu 1 is shared.  That would not be expressible above (ranks are uniform there), but would work with:

```py
cpu_map = [[0], [1]]
gpu_map = [[0, 1], [1, 2]]
```

Yes, I know, it's a  contrived example, but that is what I mean with non-uniform ranks.  I don't see why the current change would be required by this PR.

src/radical/pilot/agent/launch_method/mpirun.py
andre-merzky: Fair.  I wouldn't mind reconsidering this though since it seems to fall out for free and we add an artificial limitation with this check.

src/radical/pilot/agent/scheduler/continuous.py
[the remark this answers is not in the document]
mtitov: btw, not sure it is the case, but `ru.cluster_list` may lost elements if that doesn't fit the cluster size
```python
>>> import radical.utils as ru
>>> cores_per_rank = 2
>>> cores = [1, 2, 5, 8, 6, 9]
>>> list(ru.cluster_list(cores, cores_per_rank))
[(1, 2), (5, 8), (6, 9)]
>>> cores = [1, 2, 5, 8, 6]
>>> list(ru.cluster_list(cores, cores_per_rank))
[(1, 2), (5, 8)]
```

src/radical/pilot/agent/scheduler/base.py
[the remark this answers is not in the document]
mtitov: resolved

src/radical/pilot/agent/scheduler/base.py
```diff
@@ -873,8 +869,8 @@ def _schedule_incoming(self):
                     task['$set']      = ['resources']
                     task['resources'] = {'cpu': td['ranks'] *
                                                 td['cores_per_rank'],
-                                         'gpu': td['ranks'] *
-                                                td['gpus_per_rank']}
+                                         'gpu': int(td['ranks'] *
+                                                    td['gpus_per_rank'])}
```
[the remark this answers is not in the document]
mtitov: resolved

src/radical/pilot/raptor/master.py
mtitov: removed

src/radical/pilot/raptor/master.py
mtitov: removed (rolled back)

src/radical/pilot/agent/scheduler/continuous.py
mtitov: agreed and updated accordingly

src/radical/pilot/agent/launch_method/mpirun.py
mtitov: removed check for both MPIRun and SRun LMs, and since SRun uses actual number of GPUs per task/rank, then it takes such information from `slots` (and not from a corresponding task description)

src/radical/pilot/agent/scheduler/continuous.py
mtitov: resolved

mtitov:
@andre-merzky Andre, it is ready for a next round of review

PR #2838, opened 2023-02-27
add rank documentation

andre-merzky:


src/radical/pilot/task_description.py
```diff
@@ -399,60 +399,33 @@ class TaskDescription(ru.TypedDict):
...
-        * rp.COPY     : local file copy, i.e., not crossing host boundaries
-        * rp.MOVE     : local file move
-        * rp.LINK     : local file symlink
+    The notion of `ranks` is central to RP's `TaskDescription` class.  We here
+    use the same notion as MPI, in that the number of `ranks` refers to the
+    number of individual processes to be spawned by the task execution backend.
+    These processes will be near-exact copies of each other: they run in the
+    same workdir and the same `environment`, are defined by the same
+    `executable` and `arguments`, get the same amount of resources allocated,
+    etc.  Notable exeptions are:
```
mtitov, line 411: typo `exceptions`

src/radical/pilot/task_description.py
```diff
@@ -503,6 +476,60 @@ class TaskDescription(ru.TypedDict):
     and will cache the resulting environment settings - those cached settings
     are then applied to all other tasks with the same directives, without
     executing the directives again.
+    Staging Directives
```
mtitov, line 479: missed two empty lines above

mtitov (approved):
couple small cleanups, but overall LGTM!

mturilli:
I think we already discussed about documentation for users vs documentation for developers. This works well for the latter but not for the former. I don't think we can tell to our users to go and read the source code (or the API documentation generated from it) as the sole way to understand the basic concepts of our stack.

shantenujha:
Strongly echo the importance of this distinction and the necessity to be clear on intended audience.

andre-merzky:
> I think we already discussed about documentation for users vs documentation for developers. This works well for the latter but not for the former. I don't think we can tell to our users to go and read the source code (or the API documentation generated from it) as the sole way to understand the basic concepts of our stack.

Well, it is part of the API documentation and as such I do consider it user facing really.  That is not to say that we can include some of the text also elsewhere.  Do you have a specific place in mind?

mturilli:
I understand you consider API documentation user-facing but I am suggesting this is a problem. It is a narrow understanding of users, that assume to reduce users to developers. 10 years ago we were already saying: "I am a developer therefore I look only to API specs". I believe that is still valid today and we need to explicitly agree on that. Once in agreement, the devel team will understand that answering to the question you posed is part of the RCT developers' duties. In this case, you proposed, implemented and release the ranks interface. It is part of that task to understand how to document that interface for all our types of users, not only for other developers. As we have done already multiple times since we started to follow the new devel processes, at today's meeting, we will bring up the current documentation (in its very sorry state) and discuss together what to add to the API documentation.

mturilli:
Lift some of the text if useful for a general description and then merge as the documentation specific to the API.

mturilli (approved):
Do you need to lift part of what you wrote here and add it to the user-facing documentation, or is this to go as it is? If the latter, it is ready to go for me.

andre-merzky:
> Do you need to lift part of what you wrote here and add it to the user-facing documentation

I am not sure I understand that -- this *is* the user facing documentation, isn't it?  What am I missing?

andre-merzky:
> overview.rst is not currently part of the new documentation tree.

Can you recommend a place where the documentation should move to?  The task description notebook possibly?

mturilli:
Yes,  task description notebook seems reasonable to me.

PR #2844, opened 2023-03-10
fixed tasks binding (e.g., bind task to a certain number of cores)

mtitov:
Initially all cores with RS were bound to ranks within RS, thus happened oversubscription, with this fix binding is per task. I've also made it possible to set different bindings: `none`, `packed` and `rs`


andre-merzky (approved):
Thanks @mtitov , looks good to me.  Left a minor question about a comment.

src/radical/pilot/agent/launch_method/jsrun.py
```diff
@@ -202,10 +203,11 @@ def get_launch_cmds(self, task, exec_path):
...
             # -b: binding of tasks within a resource set (none, rs, or packed:N)
-            if td['threading_type'] == rpc.OpenMP:
-                # for OpenMP threads RP will set:
-                #    export OMP_NUM_THREADS=<cores_per_rank>
-                cmd_options += ' -b packed:%d' % cores_per_rs
+            if ranks_per_rs > 1:
+                if td['threading_type'] == rpc.OpenMP:
+                    # for OpenMP threads RP will set:
+                    #    export OMP_NUM_THREADS=<threads_per_rank>
+                    cmd_options += ' -b packed:%d' % cores_per_rank
```
andre-merzky, line 210: Why is the comment about `threads_per_rank` when the code uses `cores_per_rank` - which the previous version of the comment also used?

mtitov: with this I made it explicit that `cores_per_rank` here is about physical cores; and these are internal variables, not part of task description

andre-merzky: But the term `threads_per_rank` comes out of nowhere?  But either way, good enough for me...

Issue #2845, opened 2023-03-10
Stock RP fails to boostrap on localhost


mturilli:
Clean Ubuntu server LTS installation; default stack installed via pip:

```
$ radical-stack

  python               : /home/mturilli/ve/docs/bin/python3
  pythonpath           :
  version              : 3.10.6
  virtualenv           : /home/mturilli/ve/docs

  radical.analytics    : 1.20.1
  radical.entk         : 1.30.0
  radical.gtod         : 1.20.1
  radical.pilot        : 1.21.0
  radical.saga         : 1.21.0
  radical.utils        : 1.21.0
```

Script starts execution on localhost and hangs forever:

```
$ python het.py
new session: [rp.session.mosto.mturilli.019426.0002]                           \
database   : [mongodb://rct-tutorial:****@95.217.193.116:27017/rct-tutorial]  ok
create pilot manager                                                          ok
submit 1 pilot(s)
        pilot.0000   local.localhost         128 cores      16 gpus           ok
create task manager                                                           ok
submit: ########################################################################
wait  :
```

Only visible error is in `bootstrap_0.out`:

```
# -------------------------------------------------------------------
#
# Create virtualenv
# cmd: /usr/bin/python3 virtualenv-16.7.5/virtualenv.py -p python3 /home/mturilli/radical.pilot.sandbox/ve.local.localhost.1.21.0
#
/home/mturilli/radical.pilot.sandbox/rp.session.mosto.mturilli.019426.0002/pilot.0000/virtualenv-16.7.5/virtualenv.py:24: DeprecationWarning: The distutils package is deprecated and slated for removal in Python 3.12. Use setuptools or check PEP 632 for potential alternatives
  import distutils.spawn
/home/mturilli/radical.pilot.sandbox/rp.session.mosto.mturilli.019426.0002/pilot.0000/virtualenv-16.7.5/virtualenv.py:25: DeprecationWarning: The distutils.sysconfig module is deprecated, use sysconfig instead
  import distutils.sysconfig
Already using interpreter /usr/bin/python3
Using base prefix '/usr'
New python executable in /home/mturilli/radical.pilot.sandbox/ve.local.localhost.1.21.0/bin/python3
Also creating executable in /home/mturilli/radical.pilot.sandbox/ve.local.localhost.1.21.0/bin/python
Traceback (most recent call last):
[...]

PR #2847, opened 2023-03-11
move from custom virtualenv version to `venv` module

andre-merzky:
We have relied on a specific `virtualenv` version for pilot bootstrapping for a long time.  That worked well as it gave us a stable virtualenv stack to work with.  Now that version has fallen too far behind stock python (see #2845) and it's time to revisit that approach.  This PR switches back to the default `venv` module which has become quite stable over the last releases.  This PR though should wait for the matrix tests to kick in (see #2846) before being considered for merge.


mtitov (commented):
This update is very welcome, thank you, Andre!
Changes looks good, just need to cleanup some leftovers - every [reference of `virtenv_dist`](https://github.com/radical-cybertools/radical.pilot/search?q=virtenv_dist), we wouldn't need this anymore - in `bootstrap_0.sh`, `pmgr/launching/base.py` and resource configs

andre-merzky:
> every [reference of `virtenv_dist`](https://github.com/radical-cybertools/radical.pilot/search?q=virtenv_dist)

Ah, of course!  Thanks, that's cleaned up now.

src/radical/pilot/agent/bootstrap_0.sh
```diff
@@ -1445,7 +1442,6 @@ while getopts "a:b:cd:e:f:g:h:i:j:m:p:r:s:t:v:w:x:y:z:" OPTION; do
         d)  SDISTS="$OPTARG"                  ;;
         e)  pre_bootstrap_0 "$OPTARG"         ;;
         f)  FORWARD_TUNNEL_ENDPOINT="$OPTARG" ;;
-        g)  VIRTENV_DIST="$OPTARG"            ;;
```
mtitov, line 1504: really the last one thing - L1440 option character `g` and its description in L1420

mtitov (approved):
LGTM! just one last cleanup is an option description in bootstrapper

andre-merzky:
> LGTM! just one last cleanup is an option description in bootstrapper

Man, am I glad that you review this! :-)  Fixed!

src/radical/pilot/agent/bootstrap_0.sh
mtitov: thanks! :)

andre-merzky:
Alas, I encountered a number of Python deployments which don't have the `venv` installed by default.  ARGH.  So we have to revert this before release, or at least have the old `virtenv` fetching mechanism as fallback...

mtitov:
@andre-merzky, I don't understand that, as I see it - `venv` is part of python since 3.5 ([link](https://docs.python.org/3.6/library/venv.html)), can you please expand on errors you've encountered?

mtitov:
> The CPython distribution allows you to disable individual built-in packages. For some reason, Linux distros frequently do this, and, e.g. on Ubuntu, you have to explicitly install the `python3-venv` package, or the system python won't have the `venv` module. I don't remember seeing such problems with lmod / environment modules, but it is a common problem with the system python installation.

@eirrgang ah, I see, thank you Eric!

PR #2860, opened 2023-03-20
make sure that `radical.pilot.utils` as documented also

### LINKED ISSUE #2706: rp.utils docs don't show up

andre-merzky:
This fixes #2706

### DIFF
--- a/docs/source/apidoc.rst
+++ b/docs/source/apidoc.rst
@@ -36,7 +36,7 @@ PilotManagers
    :special-members: __init__
 
 PilotDescription
------------------------
+----------------
 .. autoclass:: radical.pilot.PilotDescription
    :members:
 
@@ -46,7 +46,7 @@ Pilots
    :members:
 
 Tasks and TaskManagers
-=============================
+======================
 
 TaskManager
 -----------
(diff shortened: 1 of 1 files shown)


mturilli (commented):
Do we have a well documented doc string for every member of each module?

eirrgang:
> Do we have a well documented doc string for every member of each module?

docstrings are missing for most package modules (`__init__.py` files). To support intersphinx links to package modules like `radical.pilot` and `radical.pilot.states`, I sprinkled rst directives like `.. py:module:: radical.pilot` into the docs on previous occasions. There is currently no docstring or target for `radical.pilot.utils`, though.

mturilli:
Thanks @eirrgang . You clearly know more about Sphinx than I do. Is this PR useful as it is? How would you suggest to proceed?

eirrgang:
> Thanks @eirrgang . You clearly know more about Sphinx than I do. Is this PR useful as it is? How would you suggest to proceed?

I think it would be good to make sure that `radical.pilot.utils` is defined in the doctree so that `:py:mod:` directives will work.

## Option 1

Add a module level docstring to `radical/pilot/utils/__init__.py` and a `.. automodule:: radical.pilot.utils` directive in the rst file.

## Option 2

Insert `.. py:module:: radical.pilot.utils` below the "Utilities and helpers" heading, before the automodule directives for modules that _do_ have module-level docstrings.

eirrgang:
> I would suggest to keep this PR about the inclusion of the submodule and to open a new ticket about the missing doc strings. Is that ok for y'all?

That seems fine. I suppose the current issue is "publish the docs". The follow-up issue(s) could be "let the following sphinx directive link to something":

    :py:mod:`radical.pilot.utils`

and, if necessary, "let `.. auto***:: ...` ...  `:members:` find documented members of entities that do not (currently) have docstrings"

andre-merzky:
> I would suggest to keep this PR about the inclusion of the submodule and to open a new ticket about the missing doc strings.

That ticket is open now (#2877).

mturilli (approved):
LGTM, thanks!

Issue #2739, opened 2022-10-21
Missing docs for raptor task modes


eirrgang:
The `examples/misc/raptor.py` example uses `rp.RAPTOR_MASTER` for the `TaskDescription.mode`, but this is not documented in https://radicalpilot.readthedocs.io/en/stable/apidoc.html#tasks-and-taskmanagers

I'm not sure whether this is an oversight. Is this interface not yet "public"? (Related to #2684?)

(`rp.RAPTOR_WORKER` is used internally, but users don't interact directly with this mode unless they override `Master.submit_workers()`.)

andre-merzky:
The way we specify task descriptions and task modes is still evolving indeed.  That is not to say that we shouldn't document the current state of affairs, the ticket is very valid - thanks!

andre-merzky:
This has been [addressed](https://github.com/radical-cybertools/radical.pilot/pull/2861) meanwhile

PR #2861, opened 2023-03-20
Fix/issue 2739

AymenFJA:
Related to #2739.

This PR adds `raptor tasks` entry to the API doc.

### DIFF
--- a/src/radical/pilot/task_description.py
+++ b/src/radical/pilot/task_description.py
@@ -165,8 +165,8 @@ class TaskDescription(ru.TypedDict):
         specific functionality (e.g., pipes, I/O redirection) which cannot easily 
         be mapped to other task attributes.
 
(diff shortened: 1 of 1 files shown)


src/radical/pilot/task_description.py
```diff
@@ -157,6 +165,11 @@ class TaskDescription(ru.TypedDict):
         specific functionality (pipes, I/O redirection) which cannot easily be
         mapped to other task attributes.
 
+        TASK_RAPTOR_MASTER and TASK_RAPTOR_WORKER are two special types of tasks
+        that define RAPTOR's engine master(s) and worker(s) components and their
```
andre-merzky, line 169: I don't think we use the term `engine` anywhere else in the context of Raptor, so would suggest to remove it.  Otherwise looks good to me - thanks!

andre-merzky (approved):

PR #2873, opened 2023-03-24
add documentation for missing properties

### LINKED ISSUE #2734: Document public attributes of Pilot and Task instances
Several `property` attributes of `Pilot` and `Task` do not have docstrings and so do not show up in `apidoc.rst` with autodoc calls like the following.
```
.. autoclass:: radical.pilot.Pilot
   :members:
```

Suggest adding docstrings to
* `Pilot.endpoint_fs`
* `Pilot.resource_sandbox`
* `Pilot.session_sandbox`
* `Pilot.client_sandbox`
* `Task.sandbox`
* `Task.endpoint_fs`
* `Task.resource_sandbox`
* `Task.session_sandbox`
* `Task.pilot_sandbox`
* `Task.client_sandbox`

andre-merzky:
This fixes #2734

### DIFF
--- a/src/radical/pilot/pilot.py
+++ b/src/radical/pilot/pilot.py
@@ -439,15 +439,38 @@ class Pilot(object):
 
     @property
     def endpoint_fs(self):
+        '''
+        Returns the URL which is internally used to access the target resource's
+        root file system.
+
+        **Returns:**
+            * A URL (radical.utils.Url).
+        '''
         return self._endpoint_fs
 
     @property
     def resource_sandbox(self):
+        '''
+        Returns the full URL of the path that RP considers the resource sandbox,
+        i.e., the sandbox on the target resource's file system which is shared
+        by all sessions which access that resource.
+
+        **Returns:**
+            * A URL (radical.utils.Url).
+        '''
         return self._resource_sandbox
 
 
     @property
     def session_sandbox(self):
+        '''
+        Returns the full URL of the path that RP considers the session sandbox
+        on the target resource's file system which is shared
+        by all pilots which access that resource in the current session.
+
+        **Returns:**
+            * A URL (radical.utils.Url).
+        '''
         return self._session_sandbox
 
     @property
--- a/src/radical/pilot/task.py
+++ b/src/radical/pilot/task.py
@@ -220,7 +220,6 @@ class Task(object):
         **Returns:**
             * A :class:`Session`.
         '''
-
         return self._session
 
 
@@ -234,7 +233,6 @@ class Task(object):
         **Returns:**
             * A :class:`TaskManager`.
         '''
-
         return self._tmgr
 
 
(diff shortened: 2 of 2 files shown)


mturilli (commented):
Do we document all the sandboxes we use for the pilot and the task? For example, does a developer know what are the differences/goal among/of resources, session, pilot and client sandboxes? If yes, then I am wondering whether we should add that information in the methods but, seeing the duplication between pilot/task, maybe adding it somewhere else?

Related but likely not affecting this PR. We documented client/agent sandboxes in the user-facing documentation. Is that enough or does a user need to know about the other sandboxes?

andre-merzky:
> Do we document all the sandboxes we use for the pilot and the task? For example, does a developer know what are the differences/goal among/of resources, session, pilot and client sandboxes? If yes, then I am wondering whether we should add that information in the methods but, seeing the duplication between pilot/task, maybe adding it somewhere else?

One place related to this documentation is the [docstring for `complete_url`](https://github.com/radical-cybertools/radical.pilot/blob/devel/src/radical/pilot/staging_directives.py#L151).  Since this is a free function in the RP module and not part of the public API I am hesitant to link that in too many places - that text might be worthwhile to extract and move elsewhere I guess - but where?

> Related but likely not affecting this PR. We documented client/agent sandboxes in the user-facing documentation. Is that enough or does a user need to know about the other sandboxes?

I guess I would reference the same text as above (docstring for `complete_url`)

mturilli:
> One place related to this documentation is the [docstring for `complete_url`](https://github.com/radical-cybertools/radical.pilot/blob/devel/src/radical/pilot/staging_directives.py#L151). Since this is a free function in the RP module and not part of the public API I am hesitant to link that in too many places - that text might be worthwhile to extract and move elsewhere I guess - but where?

Happy to consider any suggestion you may have and then see whether I have any useful opinion to contribute. I do not have a sufficient understand of the whole code base to offer useful suggestions :/
 
> > Related but likely not affecting this PR. We documented client/agent sandboxes in the user-facing documentation. Is that enough or does a user need to know about the other sandboxes?
> 
> I guess I would reference the same text as above (docstring for `complete_url`)

I am a bit lost. You are suggesting a possible solution to an answer that still needs to be answered. Do you think users should know about all the sanboxes? If yes, how that knowledge would help them? The answer to that question would be the base to document them in the user-facing documentation. For example, picking up from another PR, users need to know about the task `ranks` property because when they need to run MPI tasks on say N nodes, they need to define that property accordingly. Analogously,  in what situation users would need to know that there is a session, task, etc. sandbox?

andre-merzky:
I opened a separate issue (#2887) for this to keep this PR focused for the ScaleMS sprint.  Does that work for you @mturilli ?

mturilli:
@andre-merzky yes, it does, thanks.

mturilli (approved):
LGTM.

PR #2876, opened 2023-03-24
Configuration tutorial

mtitov:


andre-merzky (approved):
LGTM.  I left a minor comment in the NBReview thing about disabling the reporter animation.

mturilli:
I committed an iteration of the resource. Pls consider that the documentation is written for the users and that we should explain them what to do. I tried to follow that approach with my iteration. Pls review and iterate as you see fit.

mtitov:
> I committed an iteration of the resource. Pls consider that the documentation is written for the users and that we should explain them what to do. I tried to follow that approach with my iteration. Pls review and iterate as you see fit.

IMO it lost some consistency. I wrote it in a way that starting to talk about how to describe the resource and then how to use that, currently it jumps from resource description to run description then back to resource description and again run description. Maybe then we can start talking about `run description` first and then go for `resource description`?
Also I would propose to refer resource config as a `resource description` (at some point it is similar as pilot description and task description).
Also there are some statements that could be misleading (e.g., `Redefining a key/value pair when coding the radical.pilot.PilotDescription object` - as an example `queue`, we do not redefine it, since config contains the default value in case user doesn't provide any), and couple errors and typos.

Should I take another iteration over yours considering my comments above (if they are valid)?

p.s. I would also propose to keep outputs from the executed cells, since that will be helpful for users

mturilli:
- Our current use of 'resource' is overloaded as it means both platform and cpu/gpu/memory/etc. I have started to put together a vocabulary that uses `facility` (e.g. ORNL), `platform` (e.g. Summit), `resource` (e.g., cpu/gpu/memory/etc). `resource_*.json` files are at facility level, entries in those conf files are at platform level, some key/value pairs in those entries are at resource level. It is a departure from the past in which resource was useful as a synonym of platform because the actual resources were only CPUs and it was easy to call them, well, CPUs. I spot the inconsistent use of `resource` in all the notebooks and, progressively, in our publications. Thus, I started fixing it, using `[HPC] platform` in place of `resource` and leaving `resource` when speaking about CPU/GPU/etc. It is work in progress but I would like for all of us to keep refining this new vocabulary until we are happy with it, instead of just using `resource` inconsistently as we did so far. Yes, this will create an inconsistency with the key `resource` in RP's jargon but I am willing to live with that for now (I trust in the mental flexibility of the users for matter of vocabulary as RP's uses a very unique dictionary anyway). I would also start to discuss among us how to change resource to platform (or whatever else) in RP, instead of considering the current terminology of RP 'fix in stone' and having to adapt all the rest---and especially the publications/presentations---to that. I would say we took the same approach recently with the change/introduction of `ranks` and `*_rank`.
- I agree, strictly speaking, the statement about queue could be misleading but I believe the solution is to find a better entry, not to avoid explaining how to override default values in the given configuration files. The goal here is to give a user useful information, i.e., information that can help them writing scientific applications. Taking that goal into account, we want to show them how to overwrite relevant entries in the configuration file. Feel free to choose some other property than `queue` and extend the documentation with a proper example of overriding default key/pair values. We can still keep the `queue` example, fixing its description. `queue` is one of the most commonly user-defined key in a pilot description object.
- Outputs: the whole point of using notebooks is that the cells will be executed by readthedocs. That will give us a sort of CI on the documentation that, so far, aged very rapidly and kept misleading the users with outdated/invalid/plain wrong code snippets/examples. That requires we push the notebooks to github without running them. See this documentation: https://nbsphinx.readthedocs.io/en/0.9.1/ . If readthedocs will not complete all the compilations within 30 minutes, then we will have to commit the notebooks (or a subset of them) with the cells compiled while we look for a workaround. 
- Narrative: please feel free to organize it as you think the *user* would find it more intuitive.
- Typos: please fix all the ones you see. I will have to check all the notebooks anyway with a spell/grammar corrector before the final publication of the new documentation (this is why I did not do it before submitting my iteration).

docs/source/tutorials/configuration.ipynb
```diff
@@ -1,20 +1,452 @@
 {
```
okilic1, line 1: <p>Some wording suggestions:</p><ul><li>All this information is provided to the pilot of a RP application via the <code>radical.pilot.PilotDescription</code> object -&gt; All this information is provided to the pilot of a RP application via the <code>radical.pilot.PilotDescription</code>  (pd) object</li><li>Among that object's attributes, <code>resource</code> is mandatory and requires an HPC platform ID as its value. -&gt; &nbsp;The resource is mandatory among that object's attributes and requires an HPC platform ID as its value.</li><li>Users can find these pre-defined IDs -&gt; Users can find these predefined IDs</li><li>offers a general-purpose set of configuration parameters -&gt; offers a set of general-purpose configuration parameters</li><li>Users can customize existing platform configuration files by overwriting existing key/value pairs -&gt; Users can customize platform configuration files by overwriting key/value pairs.</li><li>creating a configuration file -&gt; Creating a configuration file</li><li> that ID and specific set of configurations -&gt;  that ID and a specific set of configurations</li></ul>
<br/>

 _Reply via <a href='https://app.reviewnb.com/radical-cybertools/radical.pilot/pull/2876/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='0'/>

okilic1, line 1: <ul><li>the ID for whcih they want to overraide some   -&gt;  the ID for which they want to override some </li></ul><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/radical-cybertools/radical.pilot/pull/2876/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='2'/>

okilic1, line 1: <p>There are references to RADICAL SAGA, shouln't we replace it with psi/j</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/radical-cybertools/radical.pilot/pull/2876/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='6'/>

okilic1, line 1: <p>Maybe we can combine all 4 code boxes into a one.</p><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/radical-cybertools/radical.pilot/pull/2876/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

Issue #2858, opened 2023-03-18
prof_utils.py:689 error


mturilli:
Stack:
```
python               : /home/mturilli/ve/docs/bin/python3
  pythonpath           : 
  version              : 3.10.6
  virtualenv           : /home/mturilli/ve/docs

  radical.analytics    : 1.20.1
  radical.entk         : 1.30.0
  radical.gtod         : 1.20.1
  radical.pilot        : 1.21.0
  radical.saga         : 1.21.0
  radical.utils        : 1.21.0
```

Sessions: https://github.com/radical-cybertools/radical.analytics/tree/docs/fix/docs/source/sessions

Code:
```
metrics = [
        ['Bootstrap', ['boot', 'setup_1']                         , '#c6dbef'],
        ['Warmup'   , ['warm' ]                                   , '#f0f0f0'],
        ['Schedule' , ['exec_queue','exec_prep', 'unschedule']    , '#c994c7'],
        ['Exec RP'  , ['exec_rp', 'exec_sh', 'term_sh', 'term_rp'], '#fdbb84'],
        ['Exec Cmd' , ['exec_cmd']                                , '#e31a1c'],
        ['Cooldown' , ['drain']                                   , '#addd8e']
    ]
rtypes=['cpu', 'gpu']
provided, consumed, stats_abs, stats_rel, info = exp.utilization(metrics=metrics, rtype=rtypes[1])
```

Error trace:
```
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
Cell In[11], line 4
      1 # Type of resource we want to plot: cpu or gpu
      2 rtypes=['cpu', 'gpu']
----> 4 provided, consumed, stats_abs, stats_rel, info = exp.utilization(metrics=metrics, rtype=rtypes[1])

File ~/ve/docs/lib/python3.10/site-packages/radical/analytics/experiment.py:113, in Experiment.utilization(self, metrics, rtype, udurations)
    110 for session in self._sessions:
    112     sid = session.uid
--> 113     p, c, sa, sr, i = session.utilization(metrics, rtype, udurations)
    115     provided [sid] = p
    116     consumed [sid] = c

File ~/ve/docs/lib/python3.10/site-packages/radical/analytics/session.py:992, in Session.utilization(self, metrics, rtype, udurations)
    987     raise ValueError('session utilization is only available on '
    988                      'radical.pilot sessions')
    990 import radical.pilot as rp
[...]

PR #2879, opened 2023-03-25
Docs/nb section3 task description

### LINKED ISSUE #2863: Documentation: task description notebook

andre-merzky:
This fixes #2863.  Parts can also be reused for #2868 I assume.


docs/source/tutorials/describing_tasks.ipynb
```diff
@@ -1,20 +1,563 @@
 {
```
mturilli, line 1: <p>Note that when executing this on my VM, it fails (or just hangs, I did not wait for the timeout) I assume because I do not have 32 cores and 1 gpu available. The idea is to have a notebook that can execute on the read the docs virtual machine, on the docker container we provide, or just on any devel OS when compiling the documentation and/or executing pytest with nbmake.</p><p> </p><p>I know we said we do not want expose "fake_resources" to the users but I see no alternative and I do not think at this point is too much a big deal. We can just add a Waning box explaining the risk/limitation of using it. If you agree, the tutorial would write a suitable <code>$HOME/.radical/pilot/configs/resource_local.json</code> before executing this cell. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/radical-cybertools/radical.pilot/pull/2879/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='3'/>

mturilli, line 1: <p>Should we add a note here explaining how those properties can be used programmatically when running a script instead of an interactive session in a notebook? While editing it I was left wondering that myself. Basically, explain to the users how this can be useful in a real-life situation?</p><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/radical-cybertools/radical.pilot/pull/2879/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

mturilli, line 1: <p>I think about these tutorials more like a guide than an hands on tutorial. This is why I proposed to elminate the exercise and actually show users how to do the change. Would you agree?</p><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/radical-cybertools/radical.pilot/pull/2879/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='14'/>

mturilli, line 1: <p>I am not sure we properly highlight the differences between <code>ranks</code> and <code>cores_per_ranks</code>.I tried to edit it but did not find (yet) a better way. We may want to collectively think about it.   </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/radical-cybertools/radical.pilot/pull/2879/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='16'/>

mturilli, line 1: <p>I did not find the FOO': 'foo', 'BAR': 'bar' very illustrative. I was left with a <code>foo - bar - /bin/bash</code> line that told me little about setting variables, expecially becuase the rest of the variables were in the <code>CAPITAL NAME: value</code> format. Should we expand on the example and so more meaningful variables?  </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/radical-cybertools/radical.pilot/pull/2879/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='22'/>

mturilli, line 1: <p>I am assuming we are still missing the last part? Should we introduce post_exec as a way to perform an arbitrary action at runtime based on the result of the task?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/radical-cybertools/radical.pilot/pull/2879/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='27'/>

andre-merzky: <p>The default <code>local.localhost</code> config has <code>fake_resources</code> set to True already, so my assumption was that this works out of the box on <em>any</em> localhost.  As such I don't think that a local config setting the value again would change things unfortunately.  I would like to  find out what happened here on your VM - any hint in the logs?</p>

andre-merzky: <p>I am not sure I understand what you are aiming at.  <code>how those properties can be used programmatically when running a script</code> - how would that differ from running in a notebook in this context? </p>

andre-merzky: <p>Yeah, that's fine.</p>

andre-merzky: <p>I think the difference between <code>X</code> and <code>something_per_X</code> is intuitively clear, even if one does not exactly know what <code>X</code> is precisely (which I assume is what you are worried about?).  Don't you think so? </p>

andre-merzky: <blockquote>expecially becuase the rest of the variables were in the <code>CAPITAL NAME: value</code> format </blockquote><p>I don't understand that part of the comment, as <code>FOO:foo</code> and <code>BAR:bar</code> obviously follow that very<code>CAPITAL NAME: value</code> format?  But sure, one can use different variable names.  <code>FOO</code> and <code>BAR</code> are very common names for simple examples though, just like <code>alice and bo</code><a href="https://en.wikipedia.org/wiki/Foobar" target="_blank">here</a> see [here](https://en.wikipedia.org/wiki/Foobar).</p>

andre-merzky: <p>Is that question prompted because you see this useful to the users or because you see a need for the tutorial to be comprehensive and complete?</p><p> </p><p>We have very little use cases for <code>post_exec</code> in RP (other than in RE!), and I have not seen it used in the wild really, so I think the added value of including it in the tutorial would be small at best.</p>

mturilli: A user has to write an application. How is he going to use those properties besides the toy example we offer? Why should they care? In what real-life context should they use those properties?

mturilli: Only if `ranks` (X in your example) is properly described. The current text does not do that IMHO. About the intuitively clear, our users seem to disagree as they apparently stumbled over this.

mturilli: Here is the output:
```
foo - bar - /bin/bash
RP_APP_TUNNEL_ADDR=144.76.72.175:27017
RP_BOOTSTRAP_0_REDIR=True
RP_GTOD=/home/merzky/radical.pilot.sandbox/rp.session.rivendell.merzky.019441.0011//pilot.0000//gtod
RP_PILOT_ID=pilot.0000
RP_PILOT_SANDBOX=/home/merzky/radical.pilot.sandbox/rp.session.rivendell.merzky.019441.0011//pilot.0000/
RP_PROF=/home/merzky/radical.pilot.sandbox/rp.session.rivendell.merzky.019441.0011//pilot.0000//prof
RP_PROF_TGT=/home/merzky/radical.pilot.sandbox/rp.session.rivendell.merzky.019441.0011//pilot.0000//task.000065/task.000065.prof
RP_RANK=0
RP_RANKS=1
RP_RESOURCE=local.localhost
RP_RESOURCE_SANDBOX=/home/merzky/radical.pilot.sandbox
RP_SESSION_ID=rp.session.rivendell.merzky.019441.0011
RP_SESSION_SANDBOX=/home/merzky/radical.pilot.sandbox/rp.session.rivendell.merzky.019441.0011/
RP_TASK_ID=task.000065
RP_TASK_NAME=task.000065
RP_TASK_SANDBOX=/home/merzky/radical.pilot.sandbox/rp.session.rivendell.merzky.019441.0011//pilot.0000//task.000065
```

The first line makes little sense compared to the others, and even less sense to the user IMHO. Does that help?

mturilli: Only for the user and I was indeed referring to the use cases we lately discussed in the context of RP. Happy to follow your lead and if you think it would not be useful not adding it. About the first question: `I am assuming we are still missing the last part? ` Is that the case?

andre-merzky: You are saying the examples should be a more realistic reflection of a use case?  I guess in that case we should pick a use case and use that throughout the tutorials - is that what you are aiming at?  

The comment is specifically at the section about obtaining results and error information from completed tasks.  I don't think there is much motivation needed why a user would care about that information: why else would they be running the task?  So I guess your comments aims at the examples in general across the tutorials maybe, is that correct?

andre-merzky: Right, that's what I meant, sorry if I wasn't clear.  We can lift the rank description from the docs and copy it to this place also.

andre-merzky: Yes, that helps.  We can change the command to `echo "TEST OUTPUT: $FOO - $BAR - $SHELL"` so that it's clearer where that first line comes from.  Would that help IYO?

andre-merzky: Yeah, I would leave the post-exec out here.

> About the first question: I am assuming we are still missing the last part? Is that the case?

I am sorry that I seem to have trouble to get on the same page - really am not trying to be difficult...  What is `the last part` we would be missing?  Do you mean some text to conclude the tutorial?  Or does the NB not cover what you think it should cover?  Or something else entirely?

mturilli: No hints, will have to debug. Did you try this on any other OS than yours? Maybe a container as per Mikhail's README?

Here another test I run:

```
(docs) mturilli@mosto:~/github/radical.pilot$ pytest --nbmake docs/source/tutorials/describing_tasks.ipynb
============================================================================== test session starts ===============================================================================
platform linux -- Python 3.10.6, pytest-7.2.2, pluggy-1.0.0
rootdir: /home/mturilli/github/radical.pilot
plugins: anyio-3.6.2, nbmake-1.4.1
collected 1 item

docs/source/tutorials/describing_tasks.ipynb F                                                                                                                             [100%]

==================================================================================== FAILURES ====================================================================================
________________________________________________ /home/mturilli/github/radical.pilot/docs/source/tutorials/describing_tasks.ipynb ________________________________________________
A cell timed out while it was being executed, after 300 seconds.
The message was: Cell execution timed out.
Here is a preview of the cell contents:
-------------------
['# configure reporter output ', "report = ru.Reporter(name='radical.pilot')", "report.title('Tutorial: Describing Tasks (RP version %s)' % rp.version)", '', '# create session and managers']
...
['', '# add the pilot to the task manager and wait for the pilot to become active', 'tmgr.add_pilots(pilot)', 'pilot.wait(rp.PMGR_ACTIVE)', "report.info('pilot state: %s' % pilot.state)"]
-------------------

------------------------------------------------------------------------------- Captured log call --------------------------------------------------------------------------------
ERROR    traitlets:client.py:845 Timeout waiting for execute reply (300s).
============================================================================ short test summary info =============================================================================
FAILED docs/source/tutorials/describing_tasks.ipynb:: - A cell timed out while it was being executed, after 300 seconds.
========================================================================= 1 failed in 308.45s (0:05:08) ==========================================================================
```

Stack:

```
$ radical-stack

  python               : /home/mturilli/ve/docs/bin/python3
  pythonpath           :
  version              : 3.10.6
  virtualenv           : /home/mturilli/ve/docs

  radical.analytics    : 1.20.1
  radical.entk         : 1.30.0
  radical.gtod         : 1.20.1
  radical.pilot        : 1.21.0
  radical.saga         : 1.21.0
  radical.utils        : 1.21.0
```

Maybe the same problem I have reported on #2858?

mturilli: > You are saying the examples should be a more realistic reflection of a use case?

No.

> I guess in that case we should pick a use case and use that throughout the tutorials - is that what you are aiming at?

No.

> The comment is specifically at the section about obtaining results and error information from completed tasks. I don't think there is much motivation needed why a user would care about that information: why else would they be running the task?

I do not understand your question. I try one last time then I will give up. In what situation a user would want to use the two task properties you illustrate? How would they be useful when running a scientific application? How would be introspection useful when a domain scientist runs (or is writing?) an application with RADICAL-Pilot?

> So I guess your comments aims at the examples in general across the tutorials maybe, is that correct?

No.

mturilli: It would be a good start but maybe we can find some 'interesting'/'meaningful' variables to add instead of the two generic FOO/BAR? That would require to imagine/leverage our experience about what a potential user would find Interesting and meaningful.

mturilli: C&P from your notebook:
```
Environment Setup with named_env 

When the same environment is used for many tasks, then the collective sum of theit pre_exec activities can create a significant runtime overhead, both on the shared filesystem and also on the system load.  named_env addresses that problem: applications can prepare a task environment and then use the named_env attribute to exctivate it for the task.  This process is very lightweight on system load and runtime overhead, but has some limitations we will explain after the example below:

pilot.prepare_env(env_name='test_env', 
                  env_spec={'type' :  'venv', 
                            'setup': ['psutil']})

td = rp.TaskDescription({'executable' : '/bin/sh',
                         'arguments'  : ['-c', 'which python3; pip list | grep psutil'],
                         'named_env'  : 'test_env'
                        })
task = tmgr.submit_tasks(td)
tmgr.wait_tasks([task.uid])
print(task.stdout)

report.header('finalize')
session.close()
```
End of notebook, but the user is still waiting for "but has some limitations we will explain after the example below:".

andre-merzky: Ah, thank you -  that's a hanging remark indeed, will fix.

mturilli:
@andre-merzky ping

docs/source/tutorials/describing_tasks.ipynb
andre-merzky: I gave some examples on how the attributes can be used.

andre-merzky: Done.

andre-merzky: Done - added a similar description here.

andre-merzky: I added some real-world examples.

andre-merzky:
> @andre-merzky ping

Thanks, this should be ready for another review.

docs/source/tutorials/describing_tasks.ipynb
mturilli: I will test this as part of the PR of the full documentation.

mturilli (approved):
LGTM. Thank you.

PR #2880, opened 2023-03-26
Docs/nb section3 cleaned

andre-merzky:
This PR will likely get removed at some point - it only exists to track the changes between the two nb3 branches.


mturilli:
@andre-merzky, what do we want to do with this? I think `docs/nb_section3` has reached the point in which it can be tagged, cleaned up, and moved from daft to official PR. Would you mind checking whether the changes you wanted to preserve are still actual and, in case they are, push them to `docs/nb_section3`?

andre-merzky:
> @andre-merzky, what do we want to do with this? I think `docs/nb_section3` has reached the point in which it can be tagged, cleaned up, and moved from daft to official PR. Would you mind checking whether the changes you wanted to preserve are still actual and, in case they are, push them to `docs/nb_section3`?

@mturilli : I pulled over the changes for the data staging notebook now.  That was really screwed up, it created new task managers all over the place.  The cleanup branch can now go away in my opinion.

PR #2881, opened 2023-03-26
added multi-pilots tutorial

mtitov:


docs/source/tutorials/multiple_pilots.ipynb
```diff
@@ -1,20 +1,420 @@
 {
```
mturilli, line 1: <p>Should we list all the available TaskManager 'scheduling' algorithms we support and then give an example of them below, extending the notebook? In production and for our publications we never used round robin, as it would be inefficient. I am assuming users are not likely to use it either. The one we used to most was back filling. Maybe we could also explain the pros/cons of each algorithm (or reference them) and add examples that could be useful to the users?  </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/radical-cybertools/radical.pilot/pull/2881/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='1'/>

okilic1, line 1: <p>I would reduce the number of the cores since this example runs on local host which could be someone's laptop</p><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/radical-cybertools/radical.pilot/pull/2881/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='9'/>

okilic1, line 1: <pre class="ql-syntax" spellcheck="false">Line #7.        td.arguments  = ['pilot_id=$RP_PILOT_ID']
</pre><p>Maybe adding a comment about what is the RP_PILOT_ID  and where can we find details.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/radical-cybertools/radical.pilot/pull/2881/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='14'/>

mtitov: added and updated, also rechecked the default scheduling algorithm - it is round robin.

mtitov: it is not physical resources, thus any number can be set. I've also added a comment regarding this

mtitov: @mturilli @andre-merzky should we provide a list of `RP_*` env variables available within the execution environment? should it be part of task description or configuration tutorial?

mtitov:
@mturilli @okilic1 ready for another round of review

mturilli (approved):
It may need further iterations (e.g., address the user directly) but I think we should move on and use what we have as a first version.

PR #2885, opened 2023-03-31
support sandboxes for raptor tasks

### LINKED ISSUE #2802: `task_sandbox` broken for raptor tasks
I believe this is a known problem, but I don't see that it is explicitly tracked and is likely to cause confusion.

1. The `task_sandbox` property of a raptor task generated on a client has a constructed path that is never created or used in the execution environment.
2. File staging directives with unqualified paths (relative paths instead of URIs) do not resolve to the directory in which raptor tasks actually run.
3. The `task:///` URI scheme does not work in staging directives raptor tasks.

andre-merzky:
This fixes #2802 

Test will be added.

### DIFF
--- a/docs/source/overview.rst
+++ b/docs/source/overview.rst
@@ -45,17 +45,6 @@ Summarizing, RP is the right tool if, for example, you want to:
    concurrently.
 
 
-What is a Task?
-===============
-
-Tasks are wrappers around self-contained executables, executed as one or more
-processes on the operating system of one or more compute nodes of a HPC
-cluster. As such, tasks are independent, i.e., don't execute with a shared
-memory space and tasks are not methods or functions executed by RP on the
-resources of a HPC platform. Task executables can be any 'program' like, for
-example, Gromacs, Namd or stress but also sleep, date, etc.
-
-
 What is a Workload?
 ===================
 
@@ -67,22 +56,51 @@ For RP it makes no difference if new tasks arrive while other tasks are
 executing.
 
 
-What is a Task (Task)?
-============================
+What is a Task?
+===============
 
-In RP, tasks are called ``Tasks`` (Task, or 'task'), indicating that are
-independent and self-contained tasks of computation. Each Task represents a
-self-contained, executable part of the application's workload.  A Task is
-described by the following attributes:
+Tasks are self-contained portions of an application which RP can execute
+independently from each other.  Those tasks are usually application executables
+(such as `Gromacs`, `NAMD` etc) whose execution is further parameterized by
+command line arguments, input files, environment settings, etc.  But those tasks
+can also be individual function calls or small snippets of Python code which RP
+will execute on the target resource - in that case the execution is
+parameterized by function arguments, Python excution context, etc.
 
-  * `executable`    : the name of the executable to be run on the target machines
-  * `arguments`     : a list of argument strings to be passed to the executable
-  * `environment`   : a dictionary of environment variable/value pairs to be set before task execution
-  * `input_staging` : a set of staging directives for input data
-  * `output_staging`: a set of staging directives for output data
+It is important to note that RP considers tasks independent, i.e., they don't
+execute with a shared memory space.
 
 For more details, see the
 :class:`API documentation <radical.pilot.TaskDescription>`
+
+What is a Task Rank?
+--------------------
+
+The notion of `ranks` is central to RP's `TaskDescription` class.  We here
+use the same notion as MPI, in that the number of `ranks` refers to the
+number of individual processes to be spawned by the task execution backend.
+These processes will be near-exact copies of each other: they run in the
+same workdir and the same `environment`, are defined by the same
+`executable` and `arguments`, get the same amount of resources allocated,
+etc.  Notable exceptions are:
+
+  - rank processes may run on different nodes;
+  - rank processes can communicate via MPI;
+  - each rank process obtains a unique rank ID.
+
+It is up to the underlying MPI implementation to determine the exact value
+of the process' rank ID.  The MPI implementation may also set a number of
+additional environment variables for each process.
+
+It is important to understand that only applications which make use of MPI
+should have more than one rank -- otherwise identical copies of the *same*
+application instance are launched which will compute the same results, thus
+wasting resources for all ranks but one.  Worse: I/O-routines of these
+non-MPI ranks can interfere with each other and invalidate those results.
+
+Also: applications with a single rank cannot make effective use of MPI
+- depending on the specific resource configuration, RP may launch those tasks
+without providing an MPI communicator.
 
 
 What is a Pilot?
--- a/src/radical/pilot/task_description.py
+++ b/src/radical/pilot/task_description.py
@@ -94,7 +94,7 @@ METADATA         = 'metadata'
 # ------------------------------------------------------------------------------
 #
 class TaskDescription(ru.TypedDict):
-    """
+    '''
     A TaskDescription object describes the requirements and properties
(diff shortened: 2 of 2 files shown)


src/radical/pilot/raptor/worker_default.py
```diff
@@ -279,18 +279,29 @@ def _worker_proc(res_lock):
                 os.environ['CUDA_VISIBLE_DEVICES'] = \
                              ','.join(str(i) for i in task['slots']['gpus'])
 
+            sbox = task['description'].get('sandbox')
+            if not sbox:
+                sbox = self._sbox + '/' + task['uid']
+                task['description']['sandbox'] = sbox
```
mtitov, line 285: we should follow `sandbox` description in `TaskDescription`
```
That directory *MUST* be relative to the pilot sandbox
```
`sbox` should be taken from `task['task_sandbox_path']` (https://github.com/radical-cybertools/radical.pilot/blob/devel/src/radical/pilot/session.py#L919, https://github.com/radical-cybertools/radical.pilot/blob/devel/src/radical/pilot/tmgr/scheduler/base.py#L349)
```python
if task['description'].get('sandbox'):
    sbox = task['task_sandbox_path']
else:
    sbox = self._sbox + '/' + task['uid']
```
`self._sbox` is referred to Worker sandbox, right?

src/radical/pilot/raptor/worker_mpi.py
```diff
@@ -434,11 +434,19 @@ def run(self):
                 self._prof.prof('advance',    uid=uid, state=AGENT_EXECUTING)
                 self._prof.prof('task_start', uid=uid)
 
+                sbox = task['description'].get('sandbox')
+                if not sbox:
+                    sbox = self._sbox + '/' + task['uid']
+                    task['description']['sandbox'] = sbox
```
mtitov, line 440: similar

tests/unit_tests/test_raptor/test_worker.py
```diff
@@ -111,5 +116,45 @@ def test_call(self, mocked_init, mocked_Logger):
         self.assertEqual(exc, (None, None))
 
 
+    # --------------------------------------------------------------------------
+    #
+    @mock.patch.object(DefaultWorker, '__init__', return_value=None)
+    @mock.patch('radical.utils.Logger')
+    def test_sandbox(self, mocked_init, mocked_Logger):
```
mtitov, line 123: `test_sandbox(self, mocked_Logger, mocked_init)`

tests/unit_tests/test_raptor/test_worker.py
```diff
@@ -111,5 +116,45 @@ def test_call(self, mocked_init, mocked_Logger):
...
+        component.check_pwd = os.getcwd
+        component._prof = mock.Mock()
+        component._log  = mocked_Logger
+        task = {'uid'        : 'task.0000',
+                'description': {
+                        'function': 'check_pwd',
+                        'sandbox' : '/tmp',
+                        'args'    : [],
+                        'kwargs'  : {}}}
+        out, err, ret, val, exc = component._dispatch_func(task)
```
mtitov, line 142: we should test `component._dispatch()` method here,
and for that we can patch `mp.Process` with `side_effect` as some method  that will receive `_worker_proc` and executes it.

mtitov (commented):
added several comments

src/radical/pilot/raptor/worker_default.py
andre-merzky: > self._sbox is referred to Worker sandbox, right?

Indeed

> That directory *MUST* be relative to the pilot sandbox

Hmm, I don't think we actually need this limitation anymore - the handling of absolute paths should be fixed by now.  Well, apart from what I added in this PR :-/  So let me fix this...

Ack on the `session._get_task_sandbox` call, but we need to have that available on the agent side also for raptor submitted tasks.

andre-merzky: This is fixed now.  A side effect though is that the default sboxes are relative to the pilot, not the worker.  Which is consistent thoufg with other task which is good.

src/radical/pilot/raptor/worker_mpi.py
andre-merzky: indeed.

tests/unit_tests/test_raptor/test_worker.py
andre-merzky: Thanks, fixed.

tests/unit_tests/test_raptor/test_worker.py
andre-merzky: @mtitov , would it be imprudent to ask if you can take a stab at this?

mtitov: not at all, ok, will take car of it

andre-merzky: FWIW, another option would be to make `_dispatch` a wrapper around, say, dispatch_2 like this:
```py
def _dispatch(self, ...):
    return self._dispatch_2(...)
```
which would make `_dispatch_2` testable w/o the need for side effects.  Would that simplify things IYO?

mtitov: tests are updated (and `sys.exit` is considered within tests, but a corresponding question above is still valid)

src/radical/pilot/raptor/worker_mpi.py
```diff
@@ -463,6 +467,7 @@ def run(self):
                 finally:
                     # send task back to rank 0
                     # FIXME: task_exec_stop
+                    os.chdir(self._sbox)
```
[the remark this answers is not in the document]
andre-merzky: This was fixed meanwhile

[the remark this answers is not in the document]
andre-merzky: Argh bugger - I fixed the linting in the raptor api branch, not here :-(  
https://github.com/radical-cybertools/radical.pilot/pull/2892/commits/a847ea5033c664343a88cd9c782d3b393837b1ff#diff-b357dc34307f7472a9d3a39e34f4908a9fce886863158a6e5736d1c3aeac4783R393

Let me see how we handle that...

tests/unit_tests/test_raptor/test_worker.py
andre-merzky: > is it expected that _dispatch always finishes with sys.exit?

That is indeed the case for the `DefaultWorker`: it spawns a new process for each task and we use `sys.exit()` to reliably terminate that process again.  The `MPIWorker` though does not spawn new processes and thus does not  use `sys.exit`.

andre-merzky: Thanks for the test update!!

src/radical/pilot/raptor/worker_mpi.py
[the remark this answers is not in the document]
andre-merzky: I pulled that one fix over manually now.  Sorry for fuddling this.

tests/unit_tests/test_raptor/test_worker.py
mtitov: > > is it expected that _dispatch always finishes with sys.exit?
> 
> That is indeed the case for the `DefaultWorker`: it spawns a new process for each task and we use `sys.exit()` to reliably terminate that process again. The `MPIWorker` though does not spawn new processes and thus does not use `sys.exit`.

ah, I see, that makes sense, thank you!

mtitov (approved):
LGTM! Thanks!

Issue #2767, opened 2022-11-07
Update raptor master task example


eirrgang:
In a conversation with @andre-merzky , we concluded that calling `stop` after `join` doesn't make sense.

https://github.com/radical-cybertools/radical.pilot/blob/27ba881e5258cbcd13fd7d2b2fa96ce41e434e20/examples/misc/raptor_master.py#L314

However, calling `stop` before `join` will signal the master to stop as soon as possible (correct?), and it will not wait for tasks to be submitted from the client side.

It seems like the master task `main` function should probably not call `Master.stop()`, but `Master` could call `self.stop()` with some additional logic to determine when it is appropriate.

mturilli:
@eirrgang can we close this?

eirrgang:
> @eirrgang can we close this?

I believe this will be resolved by merging #2892. I will review #2892 to confirm.

PR #2892, opened 2023-04-07
Feature/raptor api 2

### LINKED ISSUE #2731: Formalize the schema for raptor.Master.submit_workers() *descr* parameter.
The expected fields have been evolving for the dictionary passed as the *descr* argument to `rp.raptor.Master.submit_workers()`. Before the expected and allowed fields in the dictionary evolve further, it would be helpful to define a data structure or type hinting for this aggregate.

I think the RCT TypedDict would suffice, and the subclass docstring would be a good home for the notes currently in the `submit_workers()` docstring.

Also note that *environment* and *pre_exec* are currently optional fields, but are not mentioned in the `submit_workers()` docstring.

andre-merzky:
This is a somewhat major overhaul of Raptor and addresses several issues:

  - ensure that `request_cb` and `result_cb` are consistently called for all raptor tasks
  - ensure that `TASK_EXECUTABLE` tasks are correctly routed through raptor
  - add raptor masters and workers as first class API citizens
  - fix the master lifetime management
  - ensure worker termination on master's demise
  - allow raptor to run in the agent's venv if so requested
  - add RPC capabilities to raptor masters (only `stop` command supported right now)
  - consistent use of raptor specific attributes in the task descriptions
  - ensure completeness of task attributes for raptor tasks

One notable semantic change is that the `result_cb` is called *before* `AGENT_STAGING_OUTPUT` - raptor is an executor, and just like the `request_cb` is called before `AGENT_EXECUTING_PENDING`, the `result_cb` is similar bound to the `AGENT_EXECUTING` scope.

TODO: 
  - register RPC handlers
  - worker state callback in the master
  - ensure `master.wait_workers(n)` works as expected 
    - semantics: `AGENT_EXECUTING` state + comm channels are up and registered
  - ensure `executable tasks show up in the `result_cb`
  - workers die on Master termination (state notification)
  
This fixes #2731


andre-merzky:
NOTE: this PR contains #2885 and thus should be merged after that PR.

src/radical/pilot/raptor/master.py
```diff
@@ -271,67 +332,76 @@ def _state_cb(self, topic, msg):
...
-        td = TaskDescription(descr)
-        td.mode       = RAPTOR_WORKER
-        td.executable = 'radical-pilot-raptor-worker'
+            # sharing GPUs among multiple ranks not yet supported
+            if td.gpus_per_rank and not td.gpus_per_rank.is_integer():
+                raise RuntimeError('GPU sharing for workers is not supported')
 
-        # ensure that defaults and backward compatibility kick in
-        td.verify()
+            td.mode      = RAPTOR_WORKER
```
eirrgang, line 382: The mode is assigned by the `submit_workers` call rather than by the user?

src/radical/pilot/raptor/master.py
```diff
@@ -271,67 +332,76 @@ def _state_cb(self, topic, msg):
...
 
-        Note that only one rank (presumably rank 0) should register with the
-        master - the worker ranks are expected to syncronize their ranks as
+        The task `descriptions` specifically support the following keys:
+
+          - raptor_class: str, type name of worker class to execute
+          - raptor_file : str, optional if an RP worker class is used
+
+        Note that only one worker rank (presumably rank 0) should register with
+        the master - the workers are expected to syncronize their ranks as
```
eirrgang, line 366:
```suggestion
        the master - the workers are expected to synchronize their ranks as
```

examples/misc/raptor_master.py
```diff
@@ -276,20 +280,36 @@ def result_cb(self, tasks):
...
+    time.sleep(60)
+
+    out('stop')
+    # TODO: can be run from thread?
     master.stop()
+    out('join')
+
+    # TODO: worker state callback
+    master.join()
+    out('done')
```
eirrgang, line 310: Assuming the updated examples have been tested to complete successfully (such as confirming this log message), then this change resolves #2767

src/radical/pilot/raptor/master.py
```diff
@@ -1,12 +1,10 @@
 
 import os
-import sys
-import copy
 import time
 
 
 from collections import defaultdict
-from typing      import Dict, Union
+from typing      import List
```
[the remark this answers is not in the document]
andre-merzky: Hmm, that does not seem to raise a deprecation warning just yet:
```sh
$ python3.10 -c 'from typing import List; print(List)'
typing.List
```
But thanks for the heads up, we'll keep it in mind!

src/radical/pilot/raptor/master.py
andre-merzky: At this point in the code we consider it safe to assume that the description targets a worker, so we set it to make sure the downstream code knows what to do.  Consider it a service to the end user :-)

andre-merzky (commented):
Thanks for the feedback

eirrgang:
> At this point in the code we consider it safe to assume that the description targets a worker, so we set it to make sure the downstream code knows what to do. Consider it a service to the end user :-)

Okay... but that means it is not an error to set *mode=rp.TASK_EXECUTABLE*, and for a Task to have its values silently altered? I don't foresee any real problems, but it seems like a missed opportunity to catch logic errors.

andre-merzky:
> > At this point in the code we consider it safe to assume that the description targets a worker, so we set it to make sure the downstream code knows what to do. Consider it a service to the end user :-)
> 
> Okay... but that means it is not an error to set _mode=rp.TASK_EXECUTABLE_, and for a Task to have its values silently altered? I don't foresee any real problems, but it seems like a missed opportunity to catch logic errors.

Yeah, we can issue a warning in those cases.

src/radical/pilot/task_manager.py
```diff
@@ -830,52 +834,157 @@ def submit_units(self, descriptions):
...
+            td.arguments = [raptor_file, raptor_class, raptor_id]
+
+            td.environment['PYTHONUNBUFFERED'] = '1'
+
+        return self.submit_tasks(descriptions)
+
+
+    # --------------------------------------------------------------------------
+    #
+    def submit_tasks(self, descriptions=None, tasks=None):
```
mtitov, line 924: have a general proposal here - to have just tasks (`submit_tasks(self, tasks=None)`), and later while going through them (since we still go through all of them) check if certain item is `TD` or `Task`. With the quick look on where `submit_tasks` is called (and in docstrings), need to ensure (or clear up) whether task descriptions are provided or tasks themselves.

src/radical/pilot/agent/executing/base.py
```diff
@@ -175,5 +188,55 @@ def handle_timeout(self, task):
...
+                if task['description'].get('raptor_id'):
+                    if task not in buckets['raptor']:
+                        buckets['raptor'].append(task)
+
+        if buckets['client']:
+            self.advance(buckets['client'], state=state,
+                                            publish=publish, push=push, ts=ts)
+
+        if buckets['raptor']:
+            self.advance(buckets['client'], state=state,
```
mtitov, line 228: `self.advance(buckets['raptor'],...`?

src/radical/pilot/agent/executing/base.py
```diff
@@ -175,5 +188,55 @@ def handle_timeout(self, task):
...
+                                            publish=publish, push=False, ts=ts)
+            self.publish(rpc.STATE_PUBSUB, {'cmd': 'raptor_state_update',
+                                            'arg': buckets['raptor']})
+
+        if buckets['agent']:
+            self.advance(buckets['client'], state=state,
```
mtitov, line 234: `self.advance(buckets['agent'],...`?

src/radical/pilot/agent/executing/base.py
```diff
@@ -175,5 +188,55 @@ def handle_timeout(self, task):
...
+                                            publish=publish, push=False, ts=ts)
+            self.publish(rpc.STATE_PUBSUB, {'cmd': 'agent_state_update',
```
mtitov, line 236: should we use this command in `_service_state_cb` instead of `update`?

p.s. do we need such extra command? `update` is already in, whenever `cb` check for tasks it can find `origin` on its own. Kinda duplication (?)

src/radical/pilot/pilot_manager.py
```diff
@@ -608,10 +607,6 @@ def submit_pilots(self, descriptions):
             with self._pilots_lock:
                 self._pilots[pilot.uid] = pilot
 
-            if self._session._rec:
-                ru.write_json(pd.as_dict(), "%s/%s.batch.%03d.json"
-                        % (self._session._rec, pilot.uid, self._rec_id))
```
mtitov, line 613: there is a leftover of `self._rec_id` in `__init__` (in both PilotManager and TaskManager)

mtitov (commented):
no crucial points, but some are important to review

src/radical/pilot/raptor/master.py
andre-merzky: I changed this now to a check.

src/radical/pilot/raptor/master.py
andre-merzky: Thanks!

src/radical/pilot/agent/executing/popen.py
```diff
@@ -244,6 +226,8 @@ def _handle_task(self, task):
 
         ru.rec_makedir(sbox)
 
+        ru.write_json('%s/%s.json' % (sbox, tid), td)
```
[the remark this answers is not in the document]
andre-merzky: Yeah - at some intermediate point I needed the worker to have it's own description.  That's not necessary anymore, so I removed that again.  Thanks for catching.
For debugging: the descriptions are in the session json files - not as convenient, but good enough for now I think.

src/radical/pilot/agent/agent_0.py
```diff
@@ -214,6 +216,12 @@ def initialize(self):
         self.register_output(rps.AGENT_STAGING_INPUT_PENDING,
                              rpc.AGENT_STAGING_INPUT_QUEUE)
 
+        # before we run any tasks, prepare a named_env `rp` for tasks which use
+        # the pilot's own environment, such as raptors
+        env_spec = {'type': os.environ['RP_VENV_TYPE'],
+                    'path': os.environ['RP_VENV_PATH']}
+        self._prepare_env('rp', env_spec)
```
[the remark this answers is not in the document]
andre-merzky: Yes, happens to be the same, but the purpose is somewhat different and thus i would keep the redundancy in favor of clarity.  On naming - `agent` I would avoid as the don't use the term in the API.  `pilot` would work.  For the end user though the fact that it is the pilot env is not interesting - the point is that RP provides *some* default named env which can run raptor.

src/radical/pilot/task_manager.py
andre-merzky: I agree with that but differently: I reverted the change.  The move to the changed task submission should be a separate PR.

src/radical/pilot/agent/executing/base.py
andre-merzky: Oops - thank you!

src/radical/pilot/agent/executing/base.py
andre-merzky: :+1:

src/radical/pilot/pilot_manager.py
andre-merzky: Thanks - removed.

src/radical/pilot/agent/executing/base.py
andre-merzky: Good catch - that was an intermediate state of the code where that was used, is indeed not needed anymore.  I removed it.

andre-merzky:
@mtitov : this is ready for another round of reviews.  Thanks!

mtitov (approved):
LGTM!

PR #2908, opened 2023-04-15
Fix/virtenv fallback

andre-merzky:
This fixes a recently introduced problem: we removed local virtualenv download/deployment in favor of `python -m venv` as `venv` is now a core python module.  It turns out though that `venv` is not always usable as some Linux python distributions opt for an incomplete deployment of dependent modules.  This PR re-introduces the old `virtualenv` deployment as a fallback, but uses an updated `virtualenv` version to work around issues which prompted the change in the first place.

### DIFF
--- a/src/radical/pilot/agent/bootstrap_0.sh
+++ b/src/radical/pilot/agent/bootstrap_0.sh
@@ -94,6 +94,7 @@ PILOT_ID=
 RP_VERSION=
 PYTHON=
 PYTHON_DIST=
+VIRTENV_DIST=
 SESSION_ID=
 SESSION_SANDBOX=
 PILOT_SANDBOX=`pwd`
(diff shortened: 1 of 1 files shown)


src/radical/pilot/agent/bootstrap_0.sh
```diff
@@ -993,19 +1001,81 @@ virtenv_create()
...
+                echo "ERROR: couldn't create virtualenv"
+                return 1
+            fi
+        fi
+
+        # clean out virtenv sources
+        if test -d "$VIRTENV_DIR"
         then
-            echo "ERROR: couldn't create virtualenv"
-            return 1
```
mtitov, line 1008: should we just install `virtualenv` with `pip`, if `venv` is unavailable?
```
pip install virtualenv
VIRTENV_CMD="virtualenv"
```

also I don't think we need to return back `VIRTENV_DIST`, we can just check (1) `venv`, (2) `virtualenv` through `pip`, (3) a direct link for `virtualenv-16.7.5` (but I don't think we need this third option)

p.s. `"$virtenv_dist" = "20.21.0"` - this versioning is unclear, since it is not related to `virtualenv`, but to `curl` flags; and that bringing back to my point about `virtualenv_dist`

andre-merzky: check `venv`, if not check `virtualenv`,  if not download `virtualenv`.

mtitov (approved):
LGTM!

PR #2909, opened 2023-04-15
move from 3.6 to 3.7

andre-merzky:


.github/workflows/ci.yml
```diff
@@ -55,7 +55,7 @@ jobs:
         python-version: [ '3.7', '3.8', '3.9', '3.10', '3.11' ]
         include:
           - os: ubuntu-20.04
-            python-version: '3.6'
+            python-version: '3.7'
```
mtitov, line 58: these 3 lines could be removed then (3.7 is in matrix already)

.github/workflows/ci.yml
```diff
@@ -97,7 +97,7 @@ jobs:
...
```
mtitov, line 100: same as above, all three lines to be removed

setup.py
```diff
@@ -167,9 +167,9 @@ def get_version(_mod_root):
 
 
 # ------------------------------------------------------------------------------
-# check python version, should be >= 3.6
+# check python version, should be >= 3.7
 if sys.hexversion < 0x03060000:
```
mtitov, line 171: `if sys.hexversion < 0x03070000:`

mtitov (commented):
have a second thought on this change - should it be merged not with the upcoming release? (or even the one after the next?). While SAGA is still the main interface to HPCs, thus not to rush with this enforcement

setup.py
andre-merzky: Ah, I knew we had this somewhere - overlooked...  Thanks!

.github/workflows/ci.yml
andre-merzky: will do

andre-merzky:
> have a second thought on this change - should it be merged not with the upcoming release? (or even the one after the next?). While SAGA is still the main interface to HPCs, thus not to rush with this enforcement

Yeah, I don't mind pushing that down the road a bit.

eirrgang:
Note that you will also need to update the `.readthedocs.yml` or docs builds will fail.

.github/workflows/ci.yml
andre-merzky: @mtitov : Sorry, I lost the context of this comment - what lines should be removed?

mtitov:
@andre-merzky we can close this PR considering that it is part of other one, which we push right now, and I'll add comments there in regards of this topic.

andre-merzky:
> @andre-merzky we can close this PR considering that it is part of other one, which we push right now, and I'll add comments there in regards of this topic.

Agree - thanks!

Issue #2905, opened 2023-04-13
Docstring errors


mturilli:
Compiling `apidoc.rst` returns these errors:
```
radical/pilot/task_description.py:docstring of radical.pilot.task_description.TaskDescription:346: ERROR: Unexpected indentation.
radical/pilot/utils/component.py:docstring of radical.pilot.utils.component.Component:46: ERROR: Unexpected indentation.
radical/pilot/utils/component.py:docstring of radical.pilot.utils.component.Component.advance:11: ERROR: Unexpected indentation.
radical/pilot/utils/prof_utils.py:docstring of radical.pilot.utils.prof_utils.get_consumed_resources:4: WARNING: Definiti on list ends without a blank line; unexpected unindent.
radical/pilot/utils/prof_utils.py:docstring of radical.pilot.utils.prof_utils.get_consumed_resources:16: WARNING: Block quote ends without a blank line; unexpected unindent.
radical/pilot/utils/prof_utils.py:docstring of radical.pilot.utils.prof_utils.get_consumed_resources:17: WARNING: Definition list ends without a blank line; unexpected unindent.
radical/pilot/utils/prof_utils.py:docstring of radical.pilot.utils.prof_utils.get_consumed_resources:18: WARNING: Definition list ends without a blank line; unexpected unindent.
radical/pilot/utils/prof_utils.py:docstring of radical.pilot.utils.prof_utils.get_consumed_resources:20: WARNING: Definition list ends without a blank line; unexpected unindent.
radical/pilot/utils/prof_utils.py:docstring of radical.pilot.utils.prof_utils.get_session_description:3: ERROR: Unexpected indentation.
radical/pilot/utils/session.py:docstring of radical.pilot.utils.session.fetch_filetype:6: ERROR: Unexpected indentation.
radical/pilot/utils/session.py:docstring of radical.pilot.utils.session:1: WARNING: Inline emphasis start-string withoutend-string.
radical/pilot/utils/session.py:docstring of radical.pilot.utils.session.fetch_filetype:1: WARNING: Inline emphasis start-string without end-string.
radical/pilot/utils/session.py:docstring of radical.pilot.utils.session.fetch_filetype:1: WARNING: Inline emphasis start-string without end-string.
radical/pilot/utils/misc.py:docstring of radical.pilot.utils.misc.get_resource_configs:11: ERROR: Unexpected indentation.
```

eirrgang:
[...]

PR #2910, opened 2023-04-19
RsT formatting.

eirrgang:
Mitigate sphinx-build errors.

Ref #2905


eirrgang:
I have not confirmed that the built docs look right. The nb stuff is still running locally, I think. Also, it looks like my editor auto-deleted the extra line-endings that rp usually has at the ends of files. @andre-merzky had previously asked me not to do that. Sorry.

andre-merzky (approved):
Thanks a lot, @eirrgang , much appreciated!

PR #2912, opened 2023-04-24
added resource configurations for NCSA Delta

mtitov:


src/radical/pilot/configs/resource_ncsa.json
```diff
@@ -0,0 +1,176 @@
...
+                                         "order": ["SRUN"],
+                                         "SRUN" : {}
+                                        },
+        "pre_bootstrap_0"             : [
+                                         "module reset",
+                                         "module load python"
+                                        ],
+        "default_remote_workdir"      : "/scratch/%(pd.project)s/$USER",
+        "python_dist"                 : "default",
+        "virtenv_mode"                : "create",
```
andre-merzky, line 32: would it make sense to default to the `local` `virtenv_mode`?

andre-merzky (approved):
Left a question, otherwise looks good to me.

src/radical/pilot/configs/resource_ncsa.json
mtitov: yeah, will do

mtitov: done

PR #2916, opened 2023-04-25
Docs/nb section3 platform frontier

mtitov:


andre-merzky (approved):
Looks great to me!

docs/source/supported/frontier.rst
```diff
@@ -1,4 +1,155 @@
...
+  * ``ornl.frontier`` - ``SRUN``
+
+* Configuration per node (9,408 nodes in total)
+
+  * 64 CPU cores, each core has 2 threads (``SMT=2``)
+  * 8 GPUs (AMD MI250X)
+  * 512 GiB of memory
+
+.. note::
+   RADICAL-Pilot allows to use the ``--constraint`` option
```
mturilli, line 25: I tried to edit this paragraph but I do not understand it. Could you edit it further or could we discuss about it so that I can understand what we need to write?

mturilli (changes requested):
I left a comment about a needed editing. After that, I think it is ready to be merged.

docs/source/supported/frontier.rst
mtitov: reworked

docs/source/supported/frontier.rst
```diff
@@ -96,21 +91,29 @@ not provided by the system)
    python3 -m venv ve.rp
    source ve.rp/bin/activate
 
-Install RADICAL-Pilot after activating a corresponding virtual environment.
+Install RADICAL-Pilot after activating a corresponding virtual environment:
+
+.. code-block:: bash
+
+   pip install radical.pilot
```
mtitov, line 102: @mturilli I actually thought to avoid it here, since we have it in Getting Started tutorial, and it is not a platform specific. That's why I thought just to refer install RCT should be enough

mturilli (approved):
Thanks!

PR #2924, opened 2023-05-04
Update source of "napoleon" Sphinx extension.

eirrgang:
Ref https://github.com/sphinx-contrib/napoleon:
 "As of Sphinx 1.3, the napoleon extension will come packaged with
 Sphinx under sphinx.ext.napoleon. The sphinxcontrib.napoleon
 extension will continue to work with Sphinx <= 1.2"

The sphinx-contrib/napoleon repository is now in an archived state and the package it provides is increasingly incompatile with Python >= 3.9.


andre-merzky (approved):
Thanks Eric!

PR #2934, opened 2023-05-12
Docs/nb section3 debug

andre-merzky:


mturilli (changes requested):
- Should this commit into `Docs/nb section3` instead of `devel`? Also, I am not sure why but I see all the files that I cleaned up already in `Docs/nb section3` I would not want to merge this and having to clean them up again. 
- Please commit the notebook without executing the cells.

andre-merzky:
Ah, of course, that should have gone to the docs branch.  Fixed.

andre-merzky:
> Please commit the notebook without executing the cells.

This was what we discussed on the last call: I did that because I wasn't sure if I should run a session or store a session, so I committed the cells to make it easier to review.  I'll change that to running a session (likely not today though).

mturilli:
> > Please commit the notebook without executing the cells.
> 
> This was what we discussed on the last call: I did that because I wasn't sure if I should run a session or store a session, so I committed the cells to make it easier to review. I'll change that to running a session (likely not today though).

I agree we discussed it on Friday but I believe we had agreed for you to:
1. run a mini-RP script in a single cell without reporter
2. (try to) use a variable to store the session ID
3. Use something like `!` to run example debug commands on the session sandbox(es)

andre-merzky:
> I agree we discussed it on Friday but I believe we had agreed for you to:

Yes, indeed - just did not do that yet...

mturilli (approved):
LGTM

mturilli:
`popen.py` breaks the rank test.

PR #2939, opened 2023-05-18
Fix stalling integration test on 3.6

mturilli:
Integration test with a notebook on Python 3.6 stalls because, likely because of the time pip takes to resolve dependencies. This enables a separate job for CI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -80,8 +80,6 @@ jobs:
         python -m pip install git+https://github.com/radical-cybertools/radical.saga.git@devel
         python -m pip install --upgrade .
         python -m pip install -r requirements-tests.txt
-        python -m pip install -r requirements-docs.txt
-        python -m pip install jupyter
     - name: Unit tests
       env:
         RADICAL_PILOT_DBURL:  mongodb://localhost:${{ job.services.mongodb.ports[27017] }}/test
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -93,16 +91,16 @@ jobs:
...
+    # - name: Integration test
+    #   env:
+    #     MONGODB_HOST: localhost
+    #     MONGODB_PORT: ${{ job.services.mongodb.ports[27017] }}
+    #     RADICAL_PILOT_DBURL:  mongodb://localhost:${{ job.services.mongodb.ports[27017] }}/test
+    #     TARGET_PATH: 'docs/source/getting_started.ipynb'
+    #   run: |
+    #     . testenv/bin/activate
+    #     jupyter nbconvert --clear-output --inplace $TARGET_PATH
+    #     jupyter nbconvert --to notebook --execute --inplace $TARGET_PATH
```
mtitov: this could be cleaned out?

mtitov (approved):
LGTM! just small comment below, and should be in sync with env-vars PR

.github/workflows/ci.yml
mturilli: I would like to leave it there because once we stop having to test 3.6, using that code requires less resources than running a separate whole job. That is because an independent job needs to do all the setup we already do for this job. Would that work for you?

mtitov: I see, makes sense, ok!

PR #2940, opened 2023-05-18
Docs/supported amarel

mtitov:


docs/source/supported/amarel.rst
```diff
@@ -0,0 +1,144 @@
...
+--------------------------
+
+Create a **virtual environment with** ``venv``:
+
+.. code-block:: bash
+
+   export PYTHONNOUSERSITE=True
+   module load python
+   python3 -m venv ve.rp
+   source ve.rp/bin/activate
```
mtitov, line 83: I'm not sure about this part, @AymenFJA can you please check/test it?

AymenFJA:
@mtitov @mturilli @andre-merzky, would we care to add a note in the tutorial regarding that the user needs to enable the VPN in order to access Amarel?

docs/source/supported/amarel.rst
mtitov: Thank you, Aymen! updated accordingly (and added other option using your explored results)

mturilli:
> @mtitov @mturilli @andre-merzky, would we care to add a note in the tutorial regarding that the user needs to enable the VPN in order to access Amarel?

Why not, maybe as a `.. note::` at the beginning when referencing the user guide?

mturilli (approved):
LGTM!

Issue #1577, opened 2018-03-15
Documentation: Add Supported launch methods of RP


georgeha:
Create a table with supported launch methods ( ex. orte, aprun) of radical pilot for each supported machine.

iparask:
I would like to push this issue to version 2.

andre-merzky:
This won't change in v2...

mturilli:
If we create a list here, I will take care of the documentation page.

andre-merzky:
```
aprun
ccmrun
flux
fork
ibrun
jsrun
mpiexec
mpirun
prte (prun)
rsh
spark
srun
ssh
yarn
```
I am uncertain about the state of spark and yarn...

mturilli:
Now we should map these to each resource we support to get a table.

andre-merzky:
```python
#!/usr/bin/env python3

import radical.utils as ru
cfg=ru.Config("radical.pilot.resource", name="*")
for site in cfg:
    for r in cfg[site]:
        try:
            print("%-15s %-25s  %-15s %-15s" % (site, r,
                                                cfg[site][r]['task_launch_method'],
                                                cfg[site][r]['mpi_launch_method']))
        except:
            print("%-15s %-15s" % (site, r))
```

:-)

```
aliases         aliases        
anl             theta                      MPIEXEC         MPIEXEC        
debug           local                      FORK            MPIRUN         
debug           summit                     FORK            FORK           
fub             allegro_rsh                RSH             MPIEXEC        
llnl            lassen                     JSRUN           JSRUN          
local           debug                      FORK            MPIRUN         
local           localhost                  FORK            MPIRUN         
local           localhost_anaconda         FORK            MPIEXEC        
local           localhost_aprun            FORK            APRUN          
local           localhost_flux             FLUX            FLUX           
local           localhost_funcs
local           localhost_orte             ORTE            ORTE           
local           localhost_ortelib          ORTE_LIB        ORTE_LIB       
local           localhost_prte             PRTE            PRTE           
local           localhost_spark            SPARK           MPIEXEC        
local           localhost_spark_anaconda   SPARK           MPIEXEC        
local           localhost_yarn             YARN            MPIEXEC        
local           prte                       JSRUN           JSRUN          
local           scale                      MPIRUN          MPIRUN         
local           traverse                   FORK            MPIRUN         
ncar            cheyenne                   MPIRUN_MPT      MPIRUN_MPT     
[...]

PR #2941, opened 2023-05-18
Addresses ticket #1577

mturilli:
This adds adds lm info to the tables of the supported resources. It also adds the script that generate an alternative table. Not too useful (and buggy because it did not return all the resource files) but we may want to improve upon it when we will get around generating the whole tables automatically.

### DIFF
--- a/docs/source/internals.rst
+++ b/docs/source/internals.rst
@@ -126,7 +126,7 @@ via agent and resource configuration files (see :ref:`chapter_supported`).
 
 
 Round-Robin Scheduler
----------------------
+----------------------
 
 The Round-Robin scheduler (`SCHEDULER_ROUND_ROBIN`) will fairly distribute
 arriving tasks over the set of known pilots, independent of task state, expected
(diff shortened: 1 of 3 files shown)


docs/source/internals.rst
```diff
@@ -126,7 +126,7 @@ via agent and resource configuration files (see :ref:`chapter_supported`).
 
 
 Round-Robin Scheduler
----------------------
+----------------------
```
andre-merzky, line 129: I think rst does not like underlines longer than the text...

andre-merzky (approved):
LGTM!

You wrote:
> buggy because it did not return all the resource files

This is unexpected - can you tell me what files were missing?  The code is essentially the same as used by the `rp.Session`, so missing files would be a bit concerning...

docs/source/internals.rst
mturilli: As strange as it is, it complains for them to be too short if I do not add one more dash. Only in this page, no idea why and did not want to spend even more time on that than I had already to do.

mturilli:
> This is unexpected - can you tell me what files were missing? The code is essentially the same as used by the `rp.Session`, so missing files would be a bit concerning...

When I used it it did not return frontier and rivanna. Now it does. Maybe I had not pull.

PR #2954, opened 2023-06-01
added grace period (20s) for pilot to terminate

mtitov:


src/radical/pilot/pilot_manager.py
```diff
@@ -835,6 +835,7 @@ def cancel_pilots(self, uids=None, _timeout=None):
         # FIXME: the cancellation request should not go directly to the DB, but
         #        through the DB abstraction layer...
         self._session._dbs.pilot_command('cancel_pilot', [], uids)
+        time.sleep(20)
```
andre-merzky, line 838: That will extend the runtime for all applications by 20 seconds.  I'd rather opt to remove the `cancel_pilot` completely and rely on the hard SAGA kill route - what do you think?  Also,  what prompted this change?

andre-merzky (commented):
Unsure...

andre-merzky (approved):
Thanks for the discussion!

PR #2955, opened 2023-06-01
some bash versions require EOL cont for bracketed calls

andre-merzky:
No idea why, but recently my new system started complaining about a shell syntax error.  I guess this went unnoticed before because of some leniency in the bash parser.  Either way, this PR fixes the syntax error.


mtitov (approved):
LGTM! (looks like a leftover, other parts in bootstrapper are covered)

PR #2957, opened 2023-06-05
fixed env dumping for RP Agent

mtitov:


bin/radical-pilot-create-static-ve
```diff
@@ -232,16 +234,20 @@ then
     fi
 fi
 
+if test -z "$EDUMP"
+then
+    mkdir -p "$(pwd)/env"
+    EDUMP="$(pwd)/env/$(basename $PREFIX).env"
```
mtitov, line 240: @andre-merzky I'm not sure about this part - putting it into `env/` - should we keep as it is or set `EDUMP="$(pwd)/$(basename $PREFIX).env"`?

andre-merzky: I like that all envrinonment-related files end up in `env/`  - so the current works for me!

andre-merzky (approved):
This looks reasonable to me - thanks for taking care of it!!

PR #2961, opened 2023-06-07
ensure that conda shell functions are exported

mtitov:


mtitov:
Tested on both Polaris and Summit (fixed the initial issue and doesn't break normal execution respectively)

@andre-merzky on Summit I see another error from `env.log`, which doesn't break the execution (attached corresponding `env.log` and `bs0_pre_0.sh`)
```
...
env/bs0_pre_0.sh: line 266: __add_sys_prefix_to_path: command not found
...
```
[summit-env.log.txt.txt](https://github.com/radical-cybertools/radical.pilot/files/11680881/summit-env.log.txt.txt)
[summit-env-bs0_pre_0.sh.txt](https://github.com/radical-cybertools/radical.pilot/files/11680882/summit-env-bs0_pre_0.sh.txt)

but we can fix it the same way - adding `test "$(set | grep '__add_sys_prefix_to_path ()')" && export -f __add_sys_prefix_to_path` - tested with this fix, and error message was eliminated - should I add this?

andre-merzky:
> but we can fix it the same way - adding `test "$(set | grep '__add_sys_prefix_to_path ()')" && export -f __add_sys_prefix_to_path` - tested with this fix, and error message was eliminated - should I add this?

If it is a non-fatal error I would rather opt not to fix it.  It is a hack after all and fragile by nature, I think it's better to keep it as minimal as possible...

andre-merzky (approved):
Thanks for adding and testing!

mtitov:
> > but we can fix it the same way - adding `test "$(set | grep '__add_sys_prefix_to_path ()')" && export -f __add_sys_prefix_to_path` - tested with this fix, and error message was eliminated - should I add this?
> 
> If it is a non-fatal error I would rather opt not to fix it. It is a hack after all and fragile by nature, I think it's better to keep it as minimal as possible...

just for more context, this is what that function is on Summit
```
__add_sys_prefix_to_path () 
{ 
    if [ -n "${_CE_CONDA}" ] && [ -n "${WINDIR+x}" ]; then
        SYSP=$(\dirname "${CONDA_EXE}");
    else
        SYSP=$(\dirname "${CONDA_EXE}");
        SYSP=$(\dirname "${SYSP}");
    fi;
    if [ -n "${WINDIR+x}" ]; then
        PATH="${SYSP}/bin:${PATH}";
        PATH="${SYSP}/Scripts:${PATH}";
        PATH="${SYSP}/Library/bin:${PATH}";
        PATH="${SYSP}/Library/usr/bin:${PATH}";
        PATH="${SYSP}/Library/mingw-w64/bin:${PATH}";
        PATH="${SYSP}:${PATH}";
    else
        PATH="${SYSP}/bin:${PATH}";
    fi;
    \export PATH
}
```

Summit does add `CONDA_EXE` to the PATH, that's why it is not an issue there, and seems like Polaris export this function on its own (that's why also not an issue there). I worry that it might be a future issue if some system will not set conda "properly" and this `__add_sys_prefix_to_path ` is conda-related function.

PR #2970, opened 2023-06-23
Use a universal and unique identifier for registered callbacks.

eirrgang:
Allow distinct instances of defined callables to be used as Pilot callbacks, whether or not the `cb` objects have `__name__` attributes (such as for `functools.partial` instances). Normalize `Pilot.{un,}register_callback()`` with the updates to `PilotManager.{un,}register_callback()`` from c0085457.

### DIFF
--- a/src/radical/pilot/pilot.py
+++ b/src/radical/pilot/pilot.py
@@ -468,9 +468,9 @@ class Pilot(object):
             raise ValueError ("invalid pmgr metric '%s'" % metric)
 
         with self._cb_lock:
-            cb_name = cb.__name__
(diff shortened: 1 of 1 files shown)


andre-merzky (approved):
LGTM

andre-merzky:
That's a good use case motivation (using `partial`) - thanks.

eirrgang:
> That's a good use case motivation (using `partial`) - thanks.

`functool.partial` is a specific example of the more general case of local functions or callables, which, even if given a `__name__`, may not be named uniquely.

PR #2977, opened 2023-07-05
Provide a target for Sphinx `:py:mod:` role.

eirrgang:
Establish a docutils target for `radical.pilot`. Allows the root package module to be linked from inside or outside of the project (via intersphinx through the `objects.inv` file).

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -1,3 +1,5 @@
+.. py:module:: radical.pilot
+
 *************
 RADICAL-Pilot
 *************


eirrgang:
There is no module docstring in `radical/pilot/__init__.py` and there doesn't seem to be a better target for links to `radical.pilot`. It is not unusual for projects with mostly-prose documentation to just let the package module docutils target point to the top of the index page, and that seems appropriate here.

mturilli (approved):
Hi @eirrgang , thank you very much. LGTM.

PR #2985, opened 2023-07-13
increase raptor's heartbeat time

AymenFJA:
This PR should fix the issue of workers ranks dying if they do not join within a specific time period. Thus, this PR increases the master time to check on the worker.

related https://github.com/radical-cybertools/other_activities/issues/59

### DIFF
--- a/src/radical/pilot/raptor/master.py
+++ b/src/radical/pilot/raptor/master.py
@@ -47,8 +47,8 @@ class Master(rpu.Component):
         self._term       = mt.Event()  # termination signal
         self._thread     = None        # run loop
 
-        self._hb_freq    = 10          # check worker heartbetas every n seconds
-        self._hb_timeout = 15          # consider worker dead after 15 seconds
+        self._hb_freq    = 1000        # check worker heartbetas every n seconds
+        self._hb_timeout = 150         # consider worker dead after 15 seconds
 
         cfg              = self._get_config(cfg)
         self._session    = Session(cfg=cfg, uid=cfg.sid, _primary=False)
--- a/src/radical/pilot/raptor/master.py
+++ b/src/radical/pilot/raptor/master.py
@@ -47,7 +47,7 @@ class Master(rpu.Component):
         self._term       = mt.Event()  # termination signal
         self._thread     = None        # run loop
 
-        self._hb_freq    = 100         # check worker heartbetas every n seconds
+        self._hb_freq    = 1000        # check worker heartbetas every n seconds
         self._hb_timeout = 150         # consider worker dead after 150 seconds
 
         cfg              = self._get_config(cfg)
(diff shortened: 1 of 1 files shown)


src/radical/pilot/raptor/master.py
```diff
@@ -47,8 +47,8 @@ def __init__(self, cfg=None):
         self._term       = mt.Event()  # termination signal
         self._thread     = None        # run loop
 
-        self._hb_freq    = 10          # check worker heartbetas every n seconds
-        self._hb_timeout = 15          # consider worker dead after 15 seconds
+        self._hb_freq    = 1000        # check worker heartbetas every n seconds
```
mtitov, line 50: maybe set every 100s?

@andre-merzky  regarding to it, should we also move `time.sleep(self._hb_freq)` after `for uid in lost`-loop?

https://github.com/radical-cybertools/radical.pilot/blob/746d54b6c05caac193dc716e1688b97475381a07/src/radical/pilot/raptor/master.py#L555-L561

src/radical/pilot/raptor/master.py
```diff
@@ -47,8 +47,8 @@ def __init__(self, cfg=None):
...
+        self._hb_timeout = 150         # consider worker dead after 15 seconds
```
mtitov, line 51: Aymen, can you please also update comment here: `15` -> `n` (similar to the comment for previous line)

src/radical/pilot/raptor/master.py
andre-merzky: Yep, that makes sense.

andre-merzky (approved):
I would not mind if the timeouts are increased even more - they should be a magnitude larger than the slowest rank startup we expect on Rivanna for runs at full scale.

Either way, glad it works!

mturilli (approved):
In general, I do not like timeouts as they should be replaced with a suitable coordinate algo. The main issue is: how do we know what is a reasonable time beyond try and error? Anyway, for the specific issue, this should work.

andre-merzky:
> In general, I do not like timeouts as they should be replaced with a suitable coordinate algo. The main issue is: how do we know what is a reasonable time beyond try and error? Anyway, for the specific issue, this should work.

Sure - but the 'proper' solution needs a bit more work.  At the moment only rank0 registers with the master which then listens for heartbeats from all ranks for that worker.  The correct solution would be for all ranks to register individually.  Out of scope for this issue though, as you said.

src/radical/pilot/raptor/master.py
mtitov: @AymenFJA just to make clear, you increased the frequency of checks, thus you check the status every 1000s, but actual timeout value is 150s. I kinda thought to have at least one check during timeout time range, maybe then to have  `hb_freq = 500` and `hb_timeout = 1000`. @andre-merzky is this correct?

andre-merzky: Indeed - the check frequency should be usually lower than the heartbeat frequency.  The timeout value is the important one to increase in this context.

PR #2987, opened 2023-07-14
fixed GPUs discovery from SLURM env variables

mtitov:
https://github.com/radical-cybertools/other_activities/issues/74


mturilli (approved):
LGTM!

PR #2998, opened 2023-07-28
stub slurm test

andre-merzky:


mtitov:
I would propose to have it as a bash script that will submit a batch script, and thus having the following format for a command: `radical-pilot-test-slurm -a <account> -p <partition> [--exclusive]`. The last parameter is to test an exclusive approach (`--exclusive` option in SBATCH and `--exact` option in SRUN) and which is optional (so we can test without it first, and then with it)

I also would use our SRUN options here, such as `--nodes`, `--ntasks`, `--cpus-per-task`, `--mem 0`, `--nodelist` (since we need to ensure all these options)

bin/radical-pilot-test-slurm
```diff
@@ -0,0 +1,74 @@
...
+    exit 1
+else
+    echo "duration consistent with concurrent execution"
+    exit 0
+fi
+
+EOT
+
+echo sbatch /tmp/$$.slurm
+cat /tmp/$$.slurm
```
mtitov, line 72: should we actually submit this script? (as I understand this `cat` just to debug PR, but with this command `radical-pilot-test-slurm` we would want to submit batch script, right?)

mtitov (approved):
LGTM! (with a comment above)

PR #3012, opened 2023-08-18
Link config tutorial

mturilli:
Following Dewi's feedback, this makes more evident the connection between supported resources and the configuration tutorial


mtitov (approved):
LGTM!

PR #3016, opened 2023-08-22
Devel nodb 2

mtitov:


mturilli:
An initial look at the logs of staging with @mtitov shows that there seem to be an issue with the tutorial and the use of relative vs. full paths for staging data. If that is the case, we should fix the tutorial but also add a discussion of paths to the tutorial itself. @andre-merzky you may want to have a look/help with that.

mturilli:
@andre-merzky do we need to fix the unit test? https://github.com/radical-cybertools/radical.pilot/actions/runs/6151031997/job/16690246955?pr=3016

mturilli:
@andre-merzky @mtitov for `describing_tasks`, see errors in `agent_executing.0000.log`.

PR #3027, opened 2023-09-11
Upload session when testing notebooks

mturilli:


.github/workflows/run-rp-notebook.yml
```diff
@@ -53,4 +53,17 @@ jobs:
...
           jupyter nbconvert --to notebook --execute --inplace $TARGET_PATH
-
+          SIDCLIENT=$(ls -rt | grep rp.session)
+          # SIDAGENT="~/radical.pilot.sandbox/$SIDCLIENT"
+          # ls $SIDCLIENT
+          # ls $SIDAGENT
+          mkdir session
+          tar cvfj $SIDCLIENT.tar.bz2 $SIDCLIENT
+          # tar cvfj $SIDAGENT.tar.bz2 $SIDAGENT
+          cp -R $SIDCLIENT.tar.bz2 session
```
mtitov, line 63: we would need to put it into a subsequent set of actions
`- name: Collect artifacts...` and make it to run all the times (or just in case of failure/cancellation), since we want to collect it if run hangs and will fail or cancelled

mtitov (approved):
LGTM!

.github/workflows/run-rp-notebook.yml
mturilli: Thanks.

PR #3032, opened 2023-09-13
Fix/nodb tests

mtitov:
Fixed unit tests for NoDB devel branch

p.s. this PR is not focused on "analytics" part, that could be skipped


andre-merzky (commented):
Thank you!

PR #3033, opened 2023-09-13
Feature/task process id

mtitov:
Preserve process ID for tasks with `executable` mode

This feature will be used in the RECUP project to determine files (tracked by Darshan), which are associated with the corresponding task.


src/radical/pilot/agent/executing/popen.py
```diff
@@ -360,6 +360,11 @@ def _handle_task(self, task):
         # store pid for last-effort termination
         _pids.append(task['proc'].pid)
 
+        # keep process ID within metadata in the task description
+        if not task['description'].get('metadata'):
+            task['description']['metadata'] = {}
+        task['description']['metadata']['process_id'] = task['proc'].pid
+
```
andre-merzky, line 367: Note that this PID is the PID of, say, `mpirun`, not of the task process itself.  Is that what Darshan needs?

Otherwise this looks good to me.

src/radical/pilot/agent/executing/popen.py
```diff
@@ -520,9 +523,9 @@ def _check_running(self, to_watch, to_cancel):
     # --------------------------------------------------------------------------
     #
     # pylint: disable=unused-argument
-    def _get_prof(self, event, tid, msg=''):
+    def _get_prof(self, event, tid, msg=None):
 
-        return '$RP_PROF %s\n' % event
+        return '$RP_PROF %s%s\n' % (event, ' %s' % msg if msg else '')
```
andre-merzky, line 528: We probably want to quote the `msg` part, like:

```py
return '$RP_PROF %s%s\n' % (event, ' "%s"' % msg if msg else '')
```

but then this would also work and maybe be a bit simpler?
```py
def _get_prof(self, event, tid, msg=''):
    return '$RP_PROF %s "%s"\n' % (event, msg)
```

src/radical/pilot/agent/executing/popen.py
```diff
@@ -762,7 +767,10 @@ def _get_exec(self, task, launcher):
         ret  = ''
 
         for cmd in ru.as_list(launcher.get_exec(task)):
-            ret += '%s\n' % cmd
+            ret += '%s &\n' % cmd
+            # collect rank PID
+            ret += 'RP_RANK_PID=$!\n'
+            ret += 'wait $RP_RANK_PID\n'
```
andre-merzky, line 773: Assume `get_exec` returns a list of commands - now the pid for the *last* command will be stored in `RP_RANK_PID` - is that what we want?

src/radical/pilot/agent/staging_output/default.py
```diff
@@ -205,9 +206,21 @@ def _handle_task_stdio(self, task):
                         self._prof.prof(ts=float(ts), event=event,
                                         comp=comp, tid=tid, uid=_uid,
                                         state=state, msg=msg)
+                        # collect task related PIDs
+                        if 'RP_LAUNCH_PID' in msg:
+                            lpid_msg, epid_msg = msg.split(':')
+                            pids['launch_pid'] = int(lpid_msg.split('=')[0])
```
andre-merzky, line 212: The index should be `[1]` I believe?

src/radical/pilot/agent/executing/popen.py
mtitov: didn't want to send an empty string, should we skip it or you are in favor to set it anyway?

src/radical/pilot/agent/executing/popen.py
andre-merzky: Empty string is fine - as long it is quoted the shell is ok with it, and we need to quote anyway.  So I would be in favor of that if that works for you.  But I don't feel strongly about it, it's a minor nitpick...

mtitov:
validated on Frontier

andre-merzky (approved):
Left one comment, but in general looks correct and ready to merge.

src/radical/pilot/agent/staging_output/default.py
```diff
@@ -205,9 +206,21 @@ def _handle_task_stdio(self, task):
...
+                        if 'RP_LAUNCH_PID' in msg:
+                            lpid_msg, epid_msg = msg.split(':')
+                            pids['launch_pid'] = int(lpid_msg.split('=')[1])
+                            pids['exec_pid']   = int(epid_msg.split('=')[1])
+                        elif 'RP_RANK_PID' in msg:
+                            pids['rank_pid'].append(int(msg.split('=')[1]))
             except Exception as e:
                 self._log.error("Pre/Post profile read failed: `%s`" % e)
 
+            # keep process IDs within metadata in the task description
```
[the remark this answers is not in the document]
mtitov: updated, thus if there is no any PIDs then `pids` dict will be empty

Issue #3020, opened 2023-08-28
Corrupted RAPTOR logs and profiles


AymenFJA:
Noticed this behavior with our latest runs on Rivanna for a paper:

Examples of profiles:
```
task.cylon.w.000000,AGENT_EXECUTIN1691692208989.0348575,task_start,master.0000
```
```
1692208989.0201461,advance,master.000000.worker.0000,Thread-4,task.cylon.w.000000,AGENT_EXECUTING,
,
1692208989.0318933,exec_start,master.000000.worker.0000,Thread-4,task.cylon.w.000000,,
```
```
1692209217.1424799,exec_stop,master.000000.worker.0000,Thread-4,task.cylon.w.000000,,
0000,,
1692209217.1438785,exec_stop,master.000000.worker.0000,Thread-4,task.cylon.w.000000,,
```

Examples of logs:
```
wait for registration to com1692208982.508 : master.000000.worker.0000 : 685266 : 46947398254592 : DEBUG    : wait for registration to complete
```

andre-merzky:
#3034 addresses this issue.

PR #3034, opened 2023-09-18
one profile per rank

andre-merzky:
this is a workaround until we have the tracer available


andre-merzky:
This fixes #3020

AymenFJA (approved):
LGTM, thanks @andre-merzky

PR #3035, opened 2023-09-21
Option `--ppn` in MPIEXEC LM

mtitov:


mturilli:
@GKNB does this PR implements the behavior you expected?

andre-merzky:
> @GKNB does this PR implements the behavior you expected?

ping?

GKNB:
I think this branch is also consistent with the main devel branch now, and in the main devel branch I can see a --ppn flag is already added for each task and it runs successfully for the mini-app, so I think it works fine.

mtitov:
confirmed with the test run on Polaris

andre-merzky (approved):
LGTM - thanks!

PR #3038, opened 2023-09-25
Upload sessions of analytics tests

mturilli:
First iteration of analytics CI to enable quick debugging


.github/workflows/ci.yml
```diff
@@ -111,18 +111,44 @@ jobs:
           python -m pip install -r requirements-ci.txt
       - name: analyze example session
         timeout-minutes: 5
+        if: always()
```
mtitov, line 114: this is not needed here (we set it for consequent steps)

.github/workflows/ci.yml
```diff
@@ -111,18 +111,44 @@ jobs:
...
           . testenv/bin/activate
           ./examples/00_getting_started.py local.localhost
-          SID=$(ls -rt | grep rp.session)
-          echo "$SID: $SID"
+          SID=$(find . -name "rp.session*" -type d -exec basename {} "rp.session" \;)
+          echo "SID analyse example session: $SID"
           radical-analytics-inspect "$SID"
           mkdir artifacts
           ls -la
           cp -R *.png *.stats artifacts
```
mtitov, line 123: should we use a single directory for all outputs? thus not to have `artifacts`, but keep it within `sessions_analytics`? (-> having here `mkdir -p sessions_analytics` and the same in the next step?)

p.s. and we will skip the step `upload artifacts`

.github/workflows/ci.yml
```diff
@@ -111,18 +111,44 @@ jobs:
...
+        if: always()
+        run: |
+          SIDCLIENT=$(find . -name "rp.session*" -type d -exec basename {} "rp.session" \;)
+          echo "SIDCLIENT collect session : $SIDCLIENT"
+          SIDAGENT="$HOME/radical.pilot.sandbox/$SIDCLIENT"
+          CLIENTNAME="analytics_client_$SIDCLIENT"
+          AGENTNAME="analytics_agent_$SIDCLIENT"
+          mkdir sessions_analytics
+          # tar cvfj $CLIENTNAME.tar.bz2 $SIDCLIENT
+          tar cvfj $CLIENTNAME.tar.bz2 `find $SIDCLIENT -type d -print`
```
mtitov, line 134: try to change the order of the options -> `cvjf` (that is something tricky with keeping `f` the last one - just discovered it)

mtitov: that same would be for tutorials

.github/workflows/ci.yml
mturilli: I added it because we had a failure (no stats file) but the session was already created. The idea is that this step can fail but we can still upload a possibly insightful session in the later steps.

.github/workflows/ci.yml
mturilli: I think we may want to have a directory for each type of test so that it makes easier to look only for the data we need to check in presence of a failure. If all the sessions are together, I still need to look for the specific key_name in a (possibly) long list of other sessions.

okilic1:
We tested with @andre-merzky, and it works as intended.

.github/workflows/ci.yml
andre-merzky: Yeah, tar should need the `f` option last as that takes the `F`ilename argument which follows the set of options.

andre-merzky:
@mturilli :  Matteo, what is the plan for this PR?

andre-merzky:
@mturilli : Matteo, what is the plan for this PR?

andre-merzky:
@mturilli : Matteo, what is the plan for this PR?

andre-merzky:
@mturilli : Matteo, what is the plan for this PR?

mtitov (approved):

PR #3052, opened 2023-10-05
Clarify raptor tutorial

ejjordan:
It is not completely clear how one can use a local function in the existing tutorial. This is easily remedied.


docs/source/tutorials/raptor.ipynb
```diff
@@ -152,7 +152,7 @@
    "source": [
```
andre-merzky, line 152: <pre class="ql-syntax" spellcheck="false">Line #12.                               'args'     : [1, 2]})
</pre><p>Shouldn't the args be removed here?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/radical-cybertools/radical.pilot/pull/3052/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='6'/>

ejjordan: Yes, sorry for that. I made a local branch and tried to push directly to the parent repo. When I saw that I didn't have permissions (perhaps such development is discouraged?), I just used the editor on github. I just forgot to also remove this line.

On a related note though, I could not get the args to work. Perhaps some documentation on how to use this is in order?
`{'mode'     : rp.TASK_FUNCTION, 'function' : 'msg',  'args'     : [3]} # Fails`

andre-merzky: Oh, let me check that!  Thanks for noting this!

andre-merzky: <p>Sorry, I was a bit blind and should have seen that quicker.  But also it points to a lack of documentation.</p><p> </p><p>What happened is not that the <code>args</code> do not work, but rather that the <code>'function': 'msg' </code>is the problem.  The function <code>msg</code> is not defined in the namespace of the raptor worker, as it only defined in the client code.  So the symbolic reference (the string <code>'msg'</code>) cannot be resolved and the call fails.  In the earlier version you used (<code>msg(3)</code>), the function (and its arguments!) is/are serialized and passed to the raptor worker and then are available there.</p><p> </p><p>I will open a documentation ticket for this - but I think it has no immediate bearing on the proposed notebook change.</p>

ejjordan: I agree that this notebook change can go forward in parallel to any changes to documentation or implementation of raptor. I just wanted to raise the issue somewhere. It is generally preferable to bring such things up in issues, but I was not sure if there was a way to make it work that I was just missing.

Please let me know if anything more is needed from my end for this PR.

andre-merzky (approved):
LGTM - Thanks!

docs/source/tutorials/raptor.ipynb
andre-merzky: Ack!  This looks good to me now.  Thanks!

docs/source/tutorials/raptor.ipynb
```diff
@@ -169,10 +169,17 @@
...
    "outputs": [],
    "source": [
+    "# function for raptor to execute\n",
+    "@rp.pythontask\n",
+    "def msg(val: int):\n",
+    "    if(val %2 == 0):\n",
+    "        print('Regular message')\n",
+    "    else:\n",
+    "        print(f'This is a very odd message: {val}')\n",
+    "\n"
```
andre-merzky, line 179: This line is missing a comma.  Hmm, how did that happen?  Strange...

andre-merzky (changes requested):
It seems like a  json error has crept into the notebook - it does not parse right now...
https://github.com/radical-cybertools/radical.pilot/actions/runs/6467348997/job/17557212094?pr=3052

docs/source/tutorials/raptor.ipynb
ejjordan: This is my fault. Sorry for not actually running the notebook myself. Not it is fixed.

andre-merzky (approved):
Thanks - this will be merged as soon as the tests pass 🙂

PR #3059, opened 2023-10-10
fixed LM config initialization

mtitov:


andre-merzky (approved):
Oops - nice catch!  LGTM

PR #3060, opened 2023-10-10
Fix/raptor rpc

andre-merzky:


src/radical/pilot/task_manager.py
```diff
@@ -540,20 +540,20 @@ def add_pilots(self, pilots):
...
                     pilot_dict = pilot.as_dict()
-                    # real object: subscribe for state updates
                     pilot.register_callback(self._pilot_state_cb)
 
                 pid = pilot_dict['uid']
 
                 if pid in self._pilots:
                     raise ValueError('pilot %s already added' % pid)
-                self._pilots[pid] = pilot_dict
+                self._pilots[pid] = pilot
```
mtitov, line 556: should be kept `pilot_dict`

src/radical/pilot/pilot.py
```diff
@@ -756,7 +756,7 @@ def _control_cb(self, topic, msg_data):
 
     # --------------------------------------------------------------------------
     #
-    def rpc(self, cmd, *args, **kwargs):
+    def rpc(self, cmd, *args, rpc_addr=None, **kwargs):
```
mtitov, line 759: the ordering of inputs is not consistent with `Component`
btw, maybe we can not to have `*args`, that's brings some confusion, and instead to have `args=` explicitly

src/radical/pilot/agent/scheduler/continuous_ddmd.py
mtitov, line 1: should we exclude it from this PR?

src/radical/pilot/raptor_tasks.py
```diff
@@ -97,16 +95,12 @@ def rpc(self, rpc: str,
...
+        kwargs['raptor_cmd'] = cmd
 
-        if not args:
-            args = dict()
-
-        args['raptor_cmd'] = rpc
-
-        return self._tmgr.pilot_rpc(self._pilot, cmd, args)
+        return self._tmgr.pilot_rpc(self._pilot, 'raptor_rpc', *args,
+                                    rpc_addr=self.uid, **kwargs)
```
mtitov, line 103: in sync with comment above

mtitov (commented):
overall looks good, but inputs for rpc-methods need to keep aligned

src/radical/pilot/task_manager.py
mtitov: I was wrong here

src/radical/pilot/agent/scheduler/continuous_ddmd.py
andre-merzky: Yes, that was a mistake...  thanks

src/radical/pilot/pilot.py
andre-merzky: We discussed this out of band.  The current version allows a more native-like call invocation, thus leaving as is.

src/radical/pilot/raptor_tasks.py
andre-merzky: Thanks for spotting this, fixed.

andre-merzky:
> overall looks good, but inputs for rpc-methods need to keep aligned

Ack, made this consistent now - well spotted!

src/radical/pilot/agent/scheduler/continuous_ddmd.py
mtitov: ping for cleanup

src/radical/pilot/task_manager.py
```diff
@@ -660,7 +660,7 @@ def pilot_rpc(self, pid, cmd, args):
         if pid not in self._pilots:
             raise ValueError('tmgr does not know pilot %s' % pid)
 
-        return self._pilots[pid].rpc(cmd=cmd, args=args)
+        return self._pilots[pid].rpc(cmd=cmd, *args, rpc_addr=rpc_addr, **kwargs)
```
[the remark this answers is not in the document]
andre-merzky: right...

mtitov (approved):
LGTM!

PR #3068, opened 2023-10-21
improve multi pilot example

andre-merzky:
The example runs faster now and the multi-gen task submission was corrected.


mtitov (approved):
LGTM!

PR #3076, opened 2023-10-27
amarel cfg

andre-merzky:
as discussed: use `exclusive` and `exact` options on amatel.

@mtitov : Maybe we can change the check in the `srun` LM to something like

```py
if submitted_with_exclusive:
    self._exact == True
```

?

### DIFF
--- a/src/radical/pilot/agent/resource_manager/base.py
+++ b/src/radical/pilot/agent/resource_manager/base.py
@@ -61,7 +61,6 @@ class RMInfo(ru.TypedDict):
             'mem_per_node'     : int,           # memory per node (MB)
 
             'details'          : {None: None},  # dict of launch method info
-            'sys_arch'         : {str: None},   # system architecture flags
             'launch_methods'   : {str: None},   # dict of launch method info
     }
 
@@ -72,7 +71,7 @@ class RMInfo(ru.TypedDict):
             'threads_per_core' : 1,
             'gpus_per_node'    : 0,
             'threads_per_gpu'  : 1,
-            'sys_arch'         : {},
+            'details'          : {},
             'launch_methods'   : {}
     }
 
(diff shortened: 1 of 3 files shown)


mtitov:
> @mtitov : Maybe we can change the check in the `srun` LM to something like
> 
> ```python
> if submitted_with_exclusive:
>     self._exact == True
> ```
> ?

yeah, that works

mtitov (approved):
LGTM!

mtitov:
@andre-merzky should we try to automate this process by checking env variable`SBATCH_EXCLUSIVE` (https://slurm.schedmd.com/sbatch.html#OPT_SBATCH_EXCLUSIVE)? if it is set, then to use `exact`

andre-merzky:
> @andre-merzky ~should we try to automate this process by checking env variable`SBATCH_EXCLUSIVE` (https://slurm.schedmd.com/sbatch.html#OPT_SBATCH_EXCLUSIVE)? if it is set, then to use `exact`~
> 
> **edit:** it is an input env variable.. looks like not applicable for us

I added code to lookup the `system_architecture` flags.  How does that look to you?

src/radical/pilot/agent/resource_manager/base.py
```diff
@@ -61,6 +61,7 @@ class RMInfo(ru.TypedDict):
             'mem_per_node'     : int,           # memory per node (MB)
 
             'details'          : {None: None},  # dict of launch method info
+            'sys_arch'         : {str: None},   # system architecture flags
```
mtitov, line 64: let's use `details` for this?
```python
'details': {'exact': bool(self._rcfg.get('system_architecture', {}).get('exclusive'))}
```

all other information that we have inside `system_architecture` is used in other `RMInfo` parameters, thus I would go for to extract it and not to preserve the original state, especially `details` is already used for additional data for LM

PR #3080, opened 2023-11-02
remove MongoDB module load from the Perlmutter config

mtitov:


andre-merzky (approved):
Thanks, makes sense of course.

PR #3083, opened 2023-11-03
set original env as a base for LM env

mtitov:
Keep this PR and have extensive testing on supported platforms


andre-merzky (approved):
LGTM :-)

PR #3086, opened 2023-11-13
enable reporter in examples

andre-merzky:
Together with https://github.com/radical-cybertools/radical.utils/pull/395 this PR es expected to address https://github.com/radical-cybertools/radical.utils/issues/394


mtitov:
we can restart tests here after corresponding RU PR will be merged into devel

mtitov (approved):
LGTM!

PR #3089, opened 2023-11-22
current version requires RU v1.43

mtitov:
With the latest changes in `ru.Reporter`


andre-merzky (approved):
Thanks!

andre-merzky:
Some tests fail because `prepare_env` does not find that RU version as it is unreleased.  Chicken/Egg - will merge.

PR #3096, opened 2023-12-02
fixed ERF creation for JSRUN LM

mtitov:


mtitov:
We keep `JSRUN_ERF` outside of the `order` parameter, thus not to initialize it as a second option to the main JSRUN LM,
and if user wants to swap them then this is the way
```
mkdir -p ~/.radical/pilot/configs
cat > ~/.radical/pilot/configs/resource_ornl.json <<EOF
{
    "summit_jsrun": {
        "launch_methods": {"order": ["JSRUN_ERF"]}
    }
}
EOF
```

andre-merzky (approved):
LGTM

PR #3097, opened 2023-12-04
Fix/raptor worker registration

mtitov:


mtitov:
Issue with the RAPTOR tutorial was the following
- in the tutorial, we submit workers with  `raptor.submit_workers` (-> `tmgr.submit_workers`)
- in the worker class, I assumed that workers are submitted from the Master class only
https://github.com/radical-cybertools/radical.pilot/blob/c313ee3b3bcbfdf522b7a7eba883302b1a4c3ac0/src/radical/pilot/raptor/master.py#L343

@andre-merzky we should consider having methods unified

src/radical/pilot/raptor/master.py
```diff
@@ -441,6 +437,8 @@ def submit_workers(self, descriptions: List[TaskDescription]
 
         self.advance(tasks, publish=True, push=True)
 
+        # dump registry with all worker descriptions ("raptor.<worker_uid>.cfg")
+        self._reg.dump('raptor_master')
```
andre-merzky, line 441: Is this left intentionally or was that just for debugging?  If intentional, then we should add the master UID to the file name, otherwise dumps for multiple masters could collide.

src/radical/pilot/raptor/worker_default.py
```diff
@@ -47,12 +47,13 @@ def __init__(self, raptor_id : str):
                                                  cb=self._request_cb)
 
         # the master should have stored our own task description in the registry
-        self._reg.dump('raptor_worker')
-        self._descr = self._reg['raptor.%s.cfg' % self._uid]
+        self._descr = self._reg['raptor.%s.cfg' % self._uid] or {}
```
andre-merzky, line 50: Maybe we should consider it an error if the config is not found?

andre-merzky (approved):
Thanks, looks good but left two questions.

src/radical/pilot/raptor/master.py
mtitov: this one was instead of other dumps we had

> we should add the master UID to the file name, otherwise dumps for multiple masters could collide

agree, will update accordingly

src/radical/pilot/raptor/worker_default.py
mtitov: that's related to this https://github.com/radical-cybertools/radical.pilot/pull/3097#issuecomment-1839751426
-> two ways to submit workers, where in one of them worker description is  is not kept in the registry

src/radical/pilot/raptor/master.py
mtitov: done

src/radical/pilot/raptor/worker_default.py
andre-merzky: Ack, thanks.

andre-merzky (approved):
LGTM, thanks!

Issue #3094, opened 2023-11-27
Set `batch`/`interactive` modes as part of the `local` access schema


mtitov:
Continued from https://github.com/radical-cybertools/radical.pilot/pull/3077

RP will determine automatically from where client is launched: either from login node or from the placed batch job:
- Resource manager will have a corresponding method to determine `[non]batch` mode;
- Session will use that method to update resource configuration accordingly.

p.s. after that as the next step we'll create a lite-launcher for RP client

PR #3100, opened 2023-12-05
added mechanism to determine batch/non-batch RP starting

mtitov:
Fixes https://github.com/radical-cybertools/radical.pilot/issues/3094


andre-merzky (approved):
LGTM!

PR #3107, opened 2023-12-15
fixed tasks handling in `prof_utils`

mtitov:


src/radical/pilot/utils/prof_utils.py
```diff
@@ -662,6 +662,10 @@ def get_session_description(sid, src=None):
         if tmgr:
             tree[tmgr]['children'].append(uid)
 
+        if 'pilot' not in task:
+            # if task haven't finished, while session got terminated
+            continue
```
andre-merzky, line 667: I would suggest to not drop the task altogether but to only skip the pilot related parts.  We still may want to see the (short) state progression of those tasks, sort them by `etype` etc.

andre-merzky (changes requested):

src/radical/pilot/utils/prof_utils.py
mtitov: I've set it that way assuming that if `pilot` is not set, then most of other information is also missing

this is an example from the latest run, and `tmgr.0000.json` has the following info regarding that not complete task
```
state: "AGENT_EXECUTING"
type: "task"
uid: "task.000010"
```

and if I skip this `pilot` check only, then I have other KeyError-related issues

<details>

```
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
Cell In[5], line 1
----> 1 session = ra.Session(sid, 'radical.pilot')
      3 data = {'session': session,
      4         'pilot'  : session.filter(etype='pilot', inplace=False).get()[0],
      5         'tasks'  : session.filter(etype='task',  inplace=False)}
      7 data['sid'] = sid

File /opt/conda/lib/python3.9/site-packages/radical/analytics/session.py:115, in Session.__init__(self, src, stype, sid, _entities, _init)
    108     raise RuntimeError('radical.analytics requires the '
    109                        'radical.pilot module to analyze this '
    110                        'session - please install it.') from e
    112 self._profile, accuracy, hostmap = \
    113         rpu.get_session_profile(sid=sid, src=self._src)
    114 self._description = \
--> 115         rpu.get_session_description(sid=sid, src=self._src)
    117 self._description['accuracy'] = accuracy
    118 self._description['hostmap']  = hostmap

File /opt/conda/lib/python3.9/site-packages/radical/pilot/utils/prof_utils.py:669, in get_session_description(sid, src)
    666     tree[pid]['children'].append(uid)
    668 if 'resources' not in task:
--> 669     td = task['description']
    670     task['resources'] = {'cpu': td['ranks'] * td['cores_per_rank'],
    671                          'gpu': td['ranks'] * td['gpus_per_rank']}
    673 # we determine the entity type by the task mode

KeyError: 'description'
```

</details>

PR #3120, opened 2024-01-24
add docs for psij deployment

andre-merzky:


docs/source/getting_started.ipynb
mturilli, line 1: Suggested iteration: ```**Simplified stack for local usage:** When using RP on a local machine or cluster there is no remote file staging or job submission. In that situation, RP can fall back to a simplified software stack.  If this is the case, run `pip install psij-python` (or equivalents for `conda` or `spack`), and RP will transparently switch to PSIJ for local job submission.```

mturilli (approved):
Left a suggestion, otherwise LGTM. Thanks!

docs/source/getting_started.ipynb
andre-merzky: applied, thanks.

PR #3129, opened 2024-02-16
add sanity check on task ranks

### LINKED ISSUE #2655: Scheduler should fail on 0 ranks
Well, that should probably be caught earlier (in this case when submitting a raptor worker) - but either way, the scheduler should also fail.

andre-merzky:
This fixes #2655

### DIFF
--- a/src/radical/pilot/agent/scheduler/base.py
+++ b/src/radical/pilot/agent/scheduler/base.py
@@ -790,13 +790,12 @@ class AgentSchedulingComponent(rpu.AgentComponent):
         task['control']          = 'tmgr_pending'
         task['exception']        = repr(e)
         task['exception_detail'] = detail
-        task['target_state']     = 'FAILED'
+        task['target_state']     = rps.FAILED
         task['$all']             = True
 
         self._log.exception('scheduling failed for %s', task['uid'])
 
-        self.advance(task, rps.FAILED, publish=True, push=False,
-                           fwd=True)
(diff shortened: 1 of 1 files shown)


src/radical/pilot/agent/scheduler/base.py
```diff
@@ -790,6 +783,22 @@ def _schedule_waitpool(self):
...
 
 
+    # --------------------------------------------------------------------------
+    #
+    def _fail_task(self, task, e, detail):
+
+        task['control']          = 'tmgr_pending'
+        task['exception']        = repr(e)
+        task['exception_detail'] = detail
+        task['target_state']     = 'FAILED'
```
mtitov, line 793: `task['target_state'] = rps.FAILED`

src/radical/pilot/agent/scheduler/base.py
```diff
@@ -790,6 +783,22 @@ def _schedule_waitpool(self):
...
+        task['$all']             = True
+
+        self._log.exception('scheduling failed for %s', task['uid'])
+
+        self.advance(task, rps.FAILED, publish=True, push=False,
+                           fwd=True)
```
mtitov, line 799: btw, for `AgentComponent`s we always have `fwd=True`, unless it is set explicitly `fwd=False`

mtitov (approved):
LGTM! 
(have just small comments)

andre-merzky:
Thanks for both comments - applied the changes!

PR #3130, opened 2024-02-16
fix a pilot cancellation race

### LINKED ISSUE #2660: pilot cancel race condition
pmgr.launcher should not start pilots for which a cancel/kill command has previously been issued.

andre-merzky:
This fixes #2660

### DIFF
--- a/src/radical/pilot/pmgr/launching/base.py
+++ b/src/radical/pilot/pmgr/launching/base.py
@@ -121,6 +121,7 @@ class PMGRLaunchingComponent(rpu.ClientComponent):
         self._pilots    = dict()      # dict for all known pilots
         self._lock      = mt.RLock()  # lock on maipulating the above
         self._sandboxes = dict()      # cache of resource sandbox URLs
+        self._cancelled = list()      # list of cancelled pilots
 
         self._mod_dir   = os.path.dirname(os.path.abspath(__file__))
         self._root_dir  = "%s/../../" % self._mod_dir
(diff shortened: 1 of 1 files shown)


src/radical/pilot/pmgr/launching/base.py
```diff
@@ -248,16 +249,20 @@ def _kill_pilots(self, pids):
...
+            else:
+                pids = list(self._pilots.keys())
 
         with self._lock:
+
             for pid in pids:
 
                 self._log.debug('cancel pilot %s', pid)
                 if pid not in self._pilots:
                     self._log.warn('cannot cancel unknown pilot %s', pid)
```
AymenFJA, line 264: Can this really happen? Also, should this be an exception instead of a warning?

AymenFJA (approved):
I left one comment, other than that this L.G.T.M. 
Thanks.

src/radical/pilot/pmgr/launching/base.py
andre-merzky: It can happen in principle, but it is a very tight race.  Assume the application issues the following two statements:

```py
pilot = pmgr.submit_pilots(...)
pilot.cancel()
```

The two lines will create messages in two different channels: the submission goes via the normal ZMQ queues, the cancellation creates a command message on the control pubsub.  Since ordering between the two channels is undefined, the launcher could in principle receive the cancellation before ever launching the pilot.

Also, in a more complex code, one application thread could launch a pilot while another thread decides to terminate all pilots for some reason.  Since pilot uids can be manually assigned (hard coded), the threads would not even need to communicate uids and it is thus easy for them to go out of sync.

But thanks for asking, it's useful to double-check those assymptions!

src/radical/pilot/pmgr/launching/base.py
```diff
@@ -278,8 +283,17 @@ def _kill_pilots(self, pids):
...
-            pilots = [pilots]
+        pilots = ru.as_list(pilots)
+
+        # weed out pilots for which we have already received a cancel request
+        to_start = list()
+        for pilot in pilots:
+            if pilot['uid'] in self._cancelled:
+                to_start.append(pilot)
+                self.advance(pilot, rps.CANCELED, publish=True, push=False)
+                continue
```
mtitov, line 294: both actions are mixed together (starting and cancelling)
```python
            if pilot['uid'] in self._cancelled:
                self.advance(pilot, rps.CANCELED, publish=True, push=False)
            else:
                to_start.append(pilot)
```

mtitov: btw tests for this PR failed due to this mistake (pilot never started, since it is not in `_cancelled`)

andre-merzky: What the heck was I thinking??  Nothing apparently... :-P

PR #3136, opened 2024-02-22
disable env inheritance in PSIJ launcher

andre-merzky:


mturilli:
Needs to take into account when inheritance is actually needed.

mtitov:
Should be considered that disabling env inheritance is for the batch script, but it is recommended to reset it inside the execution environment, since corresponding launch methods might reuse (specifically the case for SBATCH and SRUN) options set by the batch script

Slurm example
```
...
#SBATCH --export  NONE

unset SLURM_EXPORT_ENV
```

andre-merzky:
This should be ready for another review.

mtitov (approved):
LGTM!

PR #3142, opened 2024-03-04
Fix/proxy url

### LINKED ISSUE #3143: `session._proxy` uninitialized when using external proxy service

andre-merzky:
Using an external proxy URL caused `self._proxy` to remain uninitialized.
This fixes #3143


mturilli (approved):
LGTM

PR #3145, opened 2024-03-10
Fix/staging

andre-merzky:
fix bulk sandbox creation, `js_hop` lookup

### DIFF
--- a/src/radical/pilot/tmgr/staging_input/default.py
+++ b/src/radical/pilot/tmgr/staging_input/default.py
@@ -60,6 +60,7 @@ class Default(TMGRStagingInputComponent):
         self._pilots_lock  = ru.RLock()
         self._connected    = list()  # list of pilot conected by ZMQ
         self._session_sbox = self._reg['cfg.session_sandbox']
+        self._tar_idx      = 0
 
         self.register_input(rps.TMGR_STAGING_INPUT_PENDING,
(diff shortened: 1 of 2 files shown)


src/radical/pilot/tmgr/staging_input/default.py
```diff
@@ -239,10 +240,13 @@ def work(self, tasks):
 
                     tmp_path = tempfile.mkdtemp(prefix='rp_agent_tar_dir')
                     tmp_dir  = os.path.abspath(tmp_path)
-                    tar_name = '%s.%s.tar' % (self._session.uid, self.uid)
-                    tar_tgt  = '%s/%s'     % (tmp_dir, tar_name)
+                    tar_name = '%s.%s.%04d.tar' % (self._session.uid, self.uid,
```
mtitov, line 243: wouldn't it be a too long name? (`sid + tar_id` seems enough, and `self.uid` is of the component, and not of the task)

mtitov (approved):
LGTM! (small comment regarding the name for a tarball)

src/radical/pilot/tmgr/staging_input/default.py
andre-merzky: Yes, the uid is not strictly needed, but helps to distinguish from the tarballs created by the launcher base (which also contains `self._uid`).

PR #3046, opened 2023-09-29
added scheduler capable to reconfigure tasks requirements

mtitov:

PR #3151, opened 2024-03-16
Feature/agent service ep

andre-merzky:
This PR adds an ZMQ service endpoint to the pilot agent.  We don't have a use for that just yet, but this is a step toward a simpler client implementation.


andre-merzky:
TODO: make configurable (on/off)

andre-merzky:
This branch now includes changes from `feature/sched_reconfig` as that branch also touches the pilot configuration.  So this PR should be merged *after* #3046...

src/radical/pilot/agent/agent_0.py
```diff
@@ -703,5 +713,63 @@ def _prepare_env(self, env_name, env_spec):
...
+
+            sbox = '%s/%s' % (os.environ['RP_PILOT_SANDBOX'], td['uid'])
+
+            task['state']             = rps.AGENT_STAGING_INPUT_PENDING
+            task['task_sandbox_path'] = sbox
+            task['task_sandbox']      = 'file://localhost/' + sbox
+            task['pilot_sandbox']     = os.environ['RP_PILOT_SANDBOX']
+            task['session_sandbox']   = os.environ['RP_SESSION_SANDBOX']
+            task['resource_sandbox']  = os.environ['RP_RESOURCE_SANDBOX']
+            task['pilot']             = os.environ['RP_PILOT_ID']
```
mtitov, line 754: the last comment here:

should these attributes be set the same way as for service tasks? -> env variable wouldn't be available for the agent, right?
https://github.com/radical-cybertools/radical.pilot/blob/3366307dffbd2d1994de9d7f78c462ef43795eb8/src/radical/pilot/agent/agent_0.py#L333-L344

andre-merzky: Ack, I made that code consistent now.

mtitov (approved):
LGTM! Thanks!

PR #3155, opened 2024-03-22
SRUN-Bridges2 

AymenFJA:
This PR adds `SRUN` to the `Bridges2` config file and makes it the default. 

As per discussion with @andre-merzky, `MPIRUN` suffers from high launching overheads at least at the moment.


andre-merzky (approved):
Thanks @AymenFJA !

PR #3165, opened 2024-04-10
Hotfix/mkdir

andre-merzky:
Another hotfix for staging semantics to get FACTS unstuck.

### DIFF
--- a/src/radical/pilot/agent/staging_output/default.py
+++ b/src/radical/pilot/agent/staging_output/default.py
@@ -319,6 +319,7 @@ class Default(AgentStagingOutputComponent):
 
             if action == rpc.COPY:
                 try:
+                    ru.rec_makedir(os.path.dirname(tgt.path))
                     shutil.copytree(src.path, tgt.path)
                 except OSError as exc:
                     if exc.errno == errno.ENOTDIR:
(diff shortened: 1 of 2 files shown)


src/radical/pilot/agent/staging_output/default.py
```diff
@@ -317,6 +317,13 @@ def _handle_task_staging(self, task, actionables):
             if action in [rpc.COPY, rpc.LINK, rpc.MOVE]:
                 assert tgt.schema == 'file', 'staging tgt expected as file://'
 
+            # implicitly create target dir if needed - but only for local ops
+            if action != rpc.TRANSFER:
+                tgtdir = os.path.dirname(tgt.path)
+                if tgtdir != task_sandbox.path:
+                    self._log.debug("mkdir %s", tgtdir)
```
mturilli, line 324: Should we elevate this to a warning with something like: "WARNING: implicit directory creation! Mkdir %s"?

mturilli (commented):
Looks good to me, left a comment.

src/radical/pilot/agent/staging_output/default.py
andre-merzky: I don't think this would help: unless the end user is actively looking for warnings in the agent logs, it would not really show up.  Also, if we provide some semantics then users begin to rely on it (as we have just seen in this case), and then a warning is not intuitive either.

Having said that, I will open a ticket to clearly document the implemented behavior - sorry for not doing that anyway...

andre-merzky: #3166 is now open for this.

mtitov (approved):
LGTM!

PR #3171, opened 2024-04-18
Demo upgrade python packaging

ejjordan:
Demonstrate upgrading python packaging.

To be complete the rest of the scripts would need to be moved and reworked. If this change is desired then that can be done rather quickly.

I did not investigate getting a hash for dirty install version identification, but I know it is possible.


andre-merzky:
Hey, thanks for that :-)  I am in fact working on exactly this right now.  The reason for using setuptools so long was that we packaged the module sdists in the install tree so that they could be staged to the pilot side and be used in setting up the pilot sandbox, ensuring the same version on both ends.  Well, the setuptools hook we used for that now broke for good and we have no reason to stick to it anymore, so pyproject.toml it is indeed, + versioneer I guess.  Thanks for your demo, we'll push something like that into the next release!

andre-merzky:
Hey @ejjordan  - we added a `pyproject.toml` to all our modules and those will appear in the next release (this week).  We still reply on setuptools under the hood though, as discussed.  Thanks for your input which prompted this change / addition :-)

PR #3181, opened 2024-05-09
cleanup defaults for virtenv setup

mtitov:
Propose to have defaults in resource config so we reuse current [client] virtual environment and installed RP in it. And if VE should be [re]created that should be set explicitly within the config file.

p.s. I've also cleaned out constants from `PMGRLaunching`, since they are set within `ResourceConfig`

@andre-merzky if these changes are accepted, can we have it within the upcoming release?

### DIFF
--- a/docs/source/tutorials/configuration.ipynb
+++ b/docs/source/tutorials/configuration.ipynb
@@ -1,7 +1,6 @@
 {
  "cells": [
   {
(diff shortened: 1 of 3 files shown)


andre-merzky (approved):
Yes, this should make it into the release.  Thanks!

PR #3185, opened 2024-05-30
ensure proper resource info for RA

andre-merzky:
This is needed by util plots for runs where pilot size was defined in nodes.

### DIFF
--- a/src/radical/pilot/utils/prof_utils.py
+++ b/src/radical/pilot/utils/prof_utils.py
@@ -641,6 +641,23 @@ def get_session_description(sid, src=None):
         details = pilot['description']
         details = ru.dict_merge(details, pilot['resource_details'])
 
+        # if the pilot was defined by number of nodes, make sure to fill in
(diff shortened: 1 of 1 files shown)


andre-merzky:
This closes radical-cybertools/radical.analytics/issues/195

mtitov:
isn't we already have this data, is it? in `pilot['resources'] -> {'cpu': ..., 'gpu': ...}`
comes from agent
https://github.com/radical-cybertools/radical.pilot/blob/f452c78aaeeed0044b5f8f11fdad9cffb8bffae1/src/radical/pilot/agent/agent_0.py#L207-L213

mtitov (approved):
LGTM!
(at first I missed the point, but I got that it is just to fill up pilot description if `cores/gpus` were not set)

andre-merzky:
> LGTM! (at first I missed the point, but I got that it is just to fill up pilot description if `cores/gpus` were not set)

Yes, precisely :-)  Thanks!

PR #3190, opened 2024-06-17
temp files to /tmp/

### LINKED ISSUE #3070: Tests should not write into user directories
`test_session.py` creates a config file in the user config dir.  DON'T DO THIS!

andre-merzky:
This fixes #3070


LourensVeen:
It seems like you're using pytest, which has built-in tempdir support. It automatically keeps the output from the past couple invocations, with some handy symlinks to the most recent one. Works really nicely and saves some custom code. Perhaps worth a look?

 https://docs.pytest.org/en/6.2.x/tmpdir.html

andre-merzky:
@LourensVeen : Oh, thanks for that pointer, did not know that one, yet! :-)

andre-merzky:
Thanks again for the link!  For the current PR we won't use it though as that affects files which we do not want to keep around really.

mturilli (approved):
LGTM, thanks!

PR #3191, opened 2024-06-18
added SRUN option not to kill all ranks for non-mpi task if a single rank fails

mtitov:


andre-merzky (approved):
Great!

andre-merzky:
Well - great apart from the minor fact that the tests fail ;-)

PR #3192, opened 2024-06-18
ensure that services are up

### LINKED ISSUE #3193: fix startup order for service tasks
we need to make sure that service tasks which are known on agent startup are executed *before* any other tasks

andre-merzky:
This fixes #3193

### DIFF
--- a/src/radical/pilot/agent/agent_0.py
+++ b/src/radical/pilot/agent/agent_0.py
@@ -426,10 +426,10 @@ class Agent_0(rpu.AgentComponent):
 
         # FIXME: reroute to agent daemonizer
 
-        if not self.session.cfg.agents:
+        if not self._cfg.agents:
             return
 
-        n_agents      = len(self.session.cfg.agents)
+        n_agents      = len(self._cfg.agents)
         n_agent_nodes = len(self._rm.info.agent_node_list)
 
         assert n_agent_nodes >= n_agents
@@ -447,9 +447,9 @@ class Agent_0(rpu.AgentComponent):
 
         bs_name = '%s/bootstrap_2.sh'
 
-        for idx, sa in enumerate(self.session.cfg.agents):
+        for idx, sa in enumerate(self._cfg.agents):
 
-            target  = self.session.cfg.agents[sa]['target']
+            target  = self._cfg.agents[sa]['target']
             bs_args = [self._sid, self.session.cfg.reg_addr, sa]
 
             if target not in ['local', 'node']:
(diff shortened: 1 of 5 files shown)


src/radical/pilot/agent/agent_0.py
```diff
@@ -160,25 +160,32 @@ def _configure_app_comm(self):
...
         self.register_input(rps.TMGR_STAGING_OUTPUT_PENDING,
                             rpc.AGENT_COLLECTING_QUEUE,
                             cb=self._proxy_output_cb)
 
-        # and register output
-        self.register_output(rps.TMGR_STAGING_OUTPUT_PENDING,
-                             rpc.PROXY_TASK_QUEUE)
-
         self.register_rpc_handler('prepare_env', self._prepare_env,
                                                  rpc_addr=self._pid)
```
mtitov, line 191: should these lines L189-L202 be moved before service tasks start? in case of `rp` env is used by service tasks

mtitov: https://github.com/radical-cybertools/radical.pilot/blob/c0ba7f1883cfd1fff2bf5eac2c336752c4d83ab0/src/radical/pilot/agent/agent_0.py#L189-L202

andre-merzky: that makes sense, will do.

andre-merzky: The RPC handler itself is actually not useful: that can only handle `pilot.prepare_env` calls *after* the pilot became active.  We also have to pass the `prepare_env` attributes from the pilot description - that is done now.

src/radical/pilot/agent/agent_0.py
```diff
@@ -194,9 +188,25 @@ def initialize(self):
         self.rpc('prepare_env', env_name='rp', env_spec=env_spec,
                                 rpc_addr=self._pid)
 
-        # start any services if they are requested
+        # ...and prepare all envs which are defined in the pilot description
+        for env_name, env_spec in self._cfg.get('prepare_env', {}).items():
```
mtitov, line 192: `self._cfg.get(` -> `self.session.cfg.get(` (or even `self.session.cfg.prepare_env`) to use one source for config data

mtitov (commented):
changes look good and have been tested! have only one comment as a refinement

src/radical/pilot/agent/agent_0.py
andre-merzky: I would consider the reverse: `self._cfg` should be normative for components.  Pushed that now...

mtitov (approved):
LGTM!

PR #3203, opened 2024-06-27
apply LM env blacklist

### LINKED ISSUE #3183: Flux environment setup failure
Error
```
$ cat task.000000.err 
/lustre/orion/scratch/matitov/chm155/flux/radical.pilot.sandbox/rp.session.frontier07830.matitov.019859.0001//pilot.0000//gtod: error while loading shared libraries: libfabric.so.1: cannot open shared object file: No such file or directory
/lustre/orion/scratch/matitov/chm155/flux/radical.pilot.sandbox/rp.session.frontier07830.matitov.019859.0001//pilot.0000//gtod: error while loading shared libraries: libfabric.so.1: cannot open shared object file: No such file or directory
/lustre/orion/scratch/matitov/chm155/flux/radical.pilot.sandbox/rp.session.frontier07830.matitov.019859.0001/pilot.0000/task.000000//task.000000.exec.sh: line 46: module: command not found
pre_exec failed
```

Task exec script
```
#!/bin/sh

export RP_TASK_ID="task.000000"
export RP_TASK_NAME="task.000000"
export RP_PILOT_ID="pilot.0000"
export RP_SESSION_ID="rp.session.frontier07830.matitov.019859.0001"
export RP_RESOURCE="ornl.frontier"
export RP_RESOURCE_SANDBOX="/lustre/orion/scratch/matitov/chm155/flux/radical.pilot.sandbox"
export RP_SESSION_SANDBOX="$RP_RESOURCE_SANDBOX/$RP_SESSION_ID/"
export RP_PILOT_SANDBOX="$RP_SESSION_SANDBOX/pilot.0000/"
export RP_TASK_SANDBOX="$RP_PILOT_SANDBOX/task.000000"
export RP_REGISTRY_ADDRESS="tcp://10.128.122.108:10002"
export RP_CORES_PER_RANK=1
export RP_GPUS_PER_RANK=1
export RP_GTOD="$RP_PILOT_SANDBOX/gtod"
export RP_PROF="$RP_PILOT_SANDBOX/prof"
export RP_PROF_TGT="$RP_PILOT_SANDBOX/task.000000/task.000000.prof"

rp_error() {
    echo "$1 failed" 1>&2
    exit 1
}

# rank ID
export RP_RANKS=1
export RP_RANK=$FLUX_TASK_RANK

rp_sync_ranks() {
    sig=$1
    echo $RP_RANK >> $sig.sig
    while test $(cat $sig.sig | wc -l) -lt $RP_RANKS; do
        sleep 1
    done
}

$RP_PROF exec_start ""

# pre-exec commands
$RP_PROF exec_pre ""
. /sw/frontier/init/profile || rp_error pre_exec
module reset || rp_error pre_exec
module load PrgEnv-gnu || rp_error pre_exec
module load rocm/6.0.0 || rp_error pre_exec
export TF_FORCE_GPU_ALLOW_GROWTH=true || rp_error pre_exec
export MIOPEN_USER_DB_PATH=$RP_PILOT_SANDBOX/miopen-cache || rp_error pre_exec
export MIOPEN_CUSTOM_CACHE_DIR=${MIOPEN_USER_DB_PATH} || rp_error pre_exec
mkdir -p ${MIOPEN_USER_DB_PATH} || rp_error pre_exec
source /ccs/proj/chm155/IMPECCABLE/miniconda/bin/activate st_mpi_base || rp_error pre_exec
cd /lustre/orion/scratch/matitov/chm155/flux/ST || rp_error pre_exec

# execute rank
$RP_PROF rank_start ""
python3 "smiles_regress_transformer_run_large.py" &

RP_EXEC_PID=$$
RP_RANK_PID=$!

wait $RP_RANK_PID
RP_RET=$?
$RP_PROF rank_stop "RP_EXEC_PID=$RP_EXEC_PID:RP_RANK_PID=$RP_RANK_PID"

# post-exec commands
$RP_PROF exec_post ""

$RP_PROF exec_stop ""
exit $RP_RET

```

andre-merzky:
This fixes #3183, better thank before ;-)
Note that https://github.com/radical-cybertools/radical.utils/pull/410/ is required for this to work.


src/radical/pilot/agent/launch_method/srun.py
```diff
@@ -207,5 +207,21 @@ def get_rank_cmd(self):
...
+    #
+    def get_env_blacklist(self):
+        '''
+        extend the base blacklist by SLURM specific environment variables
+        '''
+
+        ret = super().get_env_blacklist()
+
+        ret.extend([
+                    'SLURM_*',
```
mtitov, line 220: this should come for RM, since we can use different LMs within the slurm - we do have `rm_info` here, so maybe to add another attribute into it as `env_blacklist`? thus there are two sources for blacklists: RM and LM

andre-merzky: That's what we though, but I realized that this is incorrect.  The current version *does* support multiple launch methods.

Let me describe that with an example.

  - agent runs in an sbatch allocation, that sets `SLURM_NNODES=4`
  - we create a named_env called `test_env`
  - one task is started with `srun -n 2`, using that named env
  - second task is started with `mpirun -n 2`, also in that named env
  
Currently, the `prepare_env` for `test_env` will store all env, including all `SLURM_*` variables such as `SLURM_NNODES=4`.

When launching, task 1 (srun) will set `SLURM_NNODES=2`, but the named env will restore the original value (`SLURM_NNODES=4`) and the task is in trouble.

Task2 on the other hand will also have `SLURM_NNODES=4` restored, but it does not care as openmpi is not interpreting that env.  `mpirun` though will set, for example, `OMPI_RANKS=2` - that's ok as the named env does not contain any `OMPI_RANKS` setting and thus there is nothing overwritten.

With the current code in this PR, the following happens: on `prepare_env`, we prepare the env [**individually for each launch method**](https://github.com/radical-cybertools/radical.pilot/pull/3203/files#diff-24efcf4c13da0181a5726fc2b70567d3040a49decb3925b3ae322a78c74d697bL204).  So when first using `srun` to launch task 1, that code will get the blacklist from `srun` which will tell `don't store "SLURM_NNODES" because srun will set it itself, leading to a conflict`.  The named env will not contain that env, and the task get's the correct env from `srun`.

When starting task 2 with mpirun, the same code will ask the mpirun LM for the blacklist.  That will say `don't store "OMPI_RANKS" because conflict!` - which is fine: the agent env did not contain `OMPI_RANKS` in the first place, so all happens as before: `SLURM_NNODES` is kept in the env but does not pose any problems.

So essentially, what the blacklist says is: "this launch methods needs to set `THIS_LM_FOO`, so don't store that in any named env to avoid conflicts".  It does not really matter if the env contains that env setting or not - and that is why we don't need to ask the RM about it either, it would not make any difference.

mtitov: sorry, I don't disagree with the concept of blacklisting, just propose to keep corresponding lists where they belong, env variables that are set by a batch system should be within RMInfo, so for Slurm RM it will be `{... 'env_blacklist': ['SLURM_*']}`, while for PBSPro RM will `{... 'env_blacklist': ['PBS_*']}`

and then we use it within LM, and will have the following in base class `LaunchMethod.get_env_blacklist`

```python
        ret = [
                 # RP internal variables
                 'RP_*',

                 # resource details
                 'ROCR_VISIBLE_DEVICES',
                 'CUDA_VISIBLE_DEVICES',
         ]
        ret.extend(self._rm_info['env_blacklist'])
        return ret
```

Other LMs will inherit it and expand it with their related env variables (for now it is just rank-related)

andre-merzky: Sorry for being unclear, but that was my point: they *belong* into the LM!  The *SRUN LM* will set `SLURM_NNODES` and will thus need to have that blacklisted from overwriting, the `MPIRUN LM` will set `OMPI_RANKS` and thus needs to have that blacklisted instead.  It does not matter what the RM thinks needs blacklisting - the *LM* needs to decide what it needs to trickle through to the task.

In you example, it does not make  sense to unset `SLURM*` when using mpirun as LM - why would we do that?  

Remember, the goal is to recreate what a user would see when running a batch script.  in that case (`sbatch` script and then `mpirun`), the `SLURM_*`  would not be unset, and `mpirun` would work all right.  With `sbatch + srun`, the `SLURM_*` setting from srun would not be unset, but instead the `SLURM_*` settings from `srun` would overwrite them.  So that is what we recreate: let the `srun LM` report what vars it will overwrite (and thus need blacklisting from env caching).

mtitov:
Need to update dependency (`requirements.txt`) for RU -> `radical.utils>=1.62`?

Besides this all looks good!

src/radical/pilot/agent/launch_method/base.py
```diff
@@ -198,11 +198,16 @@ def get_task_named_env(self, env_name):
         src  = '%s.env'                 %  base
         tgt  = '%s.%s.sh'               % (base, self.name.lower())
 
+        blacklist = self.get_env_blacklist()
+
+        self._log.debug('=== blacklist: %s', blacklist)
```
mtitov, line 203: maybe to have it for a higher debug level (not a regular one)?

andre-merzky: Done

andre-merzky:
> Need to update dependency (`requirements.txt`) for RU -> `radical.utils>=1.62`?

Will do that during the release, thanks for reminding!

mtitov (approved):
LGTM!

> Will do that during the release, thanks for reminding!

btw, this could be added here as well, since RU 1.62 is already released (and it contains necessary changes)

PR #3216, opened 2024-07-21
Fix/task cancel

andre-merzky:
This fixes task cancellation which, in certain cases, was not correctly communicated to the agent scheduler.  This should resolve the inconsistent cancel behavior Hub was seeing in DDMD.


andre-merzky:
Note: only failing test times out :-/

src/radical/pilot/utils/component.py
```diff
@@ -335,18 +336,36 @@ def _control_cb(self, topic, msg):
...
+
+            else:
+                raise ValueError('message type not handled')
+
+        except Exception as e:
 
-        except:
             # could not be handled - fall through to legacy handlers
-            pass
+            self._log.debug('fall back to legacy msg handlers [%s]', repr(e))
```
mtitov, line 362: `debug_4`? (too many messages started to appear in the regular debug mode)

src/radical/pilot/utils/component.py
```diff
@@ -314,7 +314,8 @@ def control_cb(self, topic, msg):
         This callback can be overloaded by the component to handle any control
         message which is not already handled by the component base class.
         '''
-        pass
+
+        self._log.debug('ctrl msg ignored: %s:%s', topic, msg.get('cmd'))
```
mtitov, line 318: should this also be `debug_4`? (or some other sub-level for debug)

src/radical/pilot/session.py
```diff
@@ -888,7 +900,8 @@ def close(self, **kwargs):
             # terminate all components
             if self._role == self._PRIMARY:
                 self._ctrl_pub.put(rpc.CONTROL_PUBSUB, {'cmd': 'terminate',
-                                                        'arg': None})
+                                                        'arg': None,
+                                                        'fwd': True})
```
mtitov, line 904: do we actually need to forward this command? this part is responsible to terminate client-related components, while later, pilot-related (agent side) components are terminated by calling `pmgr.close`. Kinda being precise with calls/commands

mtitov (commented):
couple follow up comments, but overall looks good

btw, `return` is missed here, right? (and default value should be used for `uids`: `arg.get('uids', [])`)
https://github.com/radical-cybertools/radical.pilot/blob/c5c288cf06f9bcbf97d5415062359bfb34860520/src/radical/pilot/agent/agent_0.py#L614-L615

src/radical/pilot/session.py
andre-merzky: Alas no: the ticket this PR is addressing stumbled over failed output staging.  That would not be captured by the exit code, the RP task state however does capture that error.

andre-merzky: There needs to be some cleanup done wrt. agent termination again, and I like to keep this one as a safeguard for now.

src/radical/pilot/utils/component.py
andre-merzky: ack

src/radical/pilot/utils/component.py
andre-merzky: ack

andre-merzky:
> btw, return is missed here, right? (and default value should be used for uids: arg.get('uids', []))

Oops, of course!

mtitov (approved):
LGTM!

PR #3221, opened 2024-08-28
ensure full task info dict

### LINKED ISSUE #3220: Task info is not updated properly
@andre-merzky we started to have such errors in CI - https://github.com/radical-cybertools/radical.pilot/actions/runs/10497264247/job/29079702311

it seems that it comes from this update (probably some race condition)
https://github.com/radical-cybertools/radical.pilot/blob/7edf7f107dd6b85eb7be1838204aa9efbadd2223/src/radical/pilot/task_manager.py#L383-L384

with this, `tmgr.0000.json` contains the following info about tasks
```
...
    "tasks": {
        "task.000000": {
            "state": "TMGR_STAGING_OUTPUT",
            "type": "task",
            "uid": "task.000000"
        },
        "task.000001": {
            "state": "TMGR_STAGING_OUTPUT",
            "type": "task",
            "uid": "task.000001"
        },
...
```

andre-merzky:
This fixes #3220.

### DIFF
--- a/src/radical/pilot/task_manager.py
+++ b/src/radical/pilot/task_manager.py
@@ -7,7 +7,9 @@ import os
 import sys
 import time
 import queue
-import threading as mt
+import collections
+
+import threading     as mt
 
 import radical.utils as ru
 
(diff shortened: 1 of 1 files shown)


mtitov:
in some cases default run (`00_getting_started.py local.localhost`) has the following results in `tmgr.0000.json` for tasks
```json
            "state": "TMGR_STAGING_OUTPUT",
            "target_state": "DONE",
```
is this expected? (I thought "state" should be set as one of the final states)

mtitov (approved):
LGTM! Thanks!

PR #3243, opened 2024-10-22
Service Startup III

andre-merzky:
This PR supports external services: endpoint information are injected via `pilot.register_service(uid, info)`.


mtitov (approved):
LGTM!

PR #3245, opened 2024-10-23
fix RTD build

andre-merzky:


andre-merzky:
TODO: find out why this times out.  Get debug logs as artifacts.

PR #3250, opened 2024-10-29
we expect task metadata to be a dict - enforce and document

andre-merzky:


src/radical/pilot/task_description.py
```diff
@@ -584,7 +584,7 @@ class TaskDescription(ru.TypedDict):
         RAPTOR_ID       : str         ,
         RAPTOR_FILE     : str         ,
         RAPTOR_CLASS    : str         ,
-        METADATA        : None        ,
+        METADATA        : {str: None} ,
```
mtitov, line 587: should we also provide an empty dictionary as a default value instead of `None`?

andre-merzky: Yeah, probably because we do the same on other attributes.  It's probably stupid to carry that many empty entities around, but oh well... :-)  Fixed!

mtitov (approved):
LGTM!

Issue #3254, opened 2024-11-01
Update documentation - resource configuration 


mtitov:
Update "Configuration System" and "Using HPC Platforms" in regard of access schema ("local" vs "remote", while "remote" is an exception and is applicable only to specific platforms), which presents endpoints for batch system and file system.

AymenFJA:
A PR is open now here #3256

AymenFJA:
This addressed in #3256

PR #3256, opened 2024-11-13
fix docs

AymenFJA:
This addresses to [#3254](https://github.com/radical-cybertools/radical.pilot/issues/3254)

### DIFF
--- a/docs/source/tutorials/submission.ipynb
+++ b/docs/source/tutorials/submission.ipynb
@@ -63,7 +63,7 @@
     "session = rp.Session()\n",
     "\n",
     "pd_init = {'resource'     : 'tacc.frontera',\n",
-    "           'access_schema': 'ssh'\n",
+    "           'access_schema': 'ssh',\n",
     "           'project'      : 'myproject',\n",
     "           'queue'        : 'normal',\n",
     "           'runtime'      : 6000,\n",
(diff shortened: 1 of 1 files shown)


andre-merzky (approved):
LGTM!

docs/source/tutorials/submission.ipynb
```diff
@@ -64,6 +64,9 @@
     "\n",
     "pd_init = {'resource'     : 'tacc.frontera',\n",
     "           'access_schema': 'ssh'\n",
```
andre-merzky, line 66: This line misses a comma, causing the NB to fail.

andre-merzky (changes requested):
Left a minor change request

mtitov:
@andre-merzky @AymenFJA have a general proposal to rewrite this section/tutorial:
1. our focus is a "local" submission - either from the login node or from the compute node (i.e., inside the batch job, using interactive job or batch script). Thus, we don't talk about `access_schema` parameters here at all, because it is set by default. We can give examples of how to get interactive job or write a batch script (while considering that the RP pilot description is not anyhow different for these examples), but we should make a note that user is required to refer to the corresponding platform's user guide to follow their instructions.
2. "remote" submission, which is **optional** and **only** if platform supports it. If it is the case, then user need to have a corresponding configuration within the resource config for that platform, and the name for that set of config parameters (e.g., `ssh`, `gsissh`, etc.) should be used in the `access_schema` for `PilotDescription`.

Thus, this section is about submission within the platform, the second part is optional

andre-merzky (approved):
LGTM - thanks!

mtitov (approved):
Let's merge it, and further updates will be in a separate PR. These changes LGTM! Thanks!

PR #3259, opened 2024-11-19
separate ve activation and env settings

andre-merzky:
A recent python  or pip or venv or whatever change screwed with our bootstrapper: `distutils` were not installed by default.  In this PR the bootstrapper ensures that it *is* installed before being used.


mtitov (approved):
LGTM! Thanks!

PR #3264, opened 2024-11-20
Feature/three

andre-merzky:
I intent to use radical.3 as test host for xgfabric, thus adding this resource config entry.  This fixes issue https://github.com/radical-cybertools/other_activities/issues/116


mtitov (approved):
LGTM!

PR #3270, opened 2024-12-02
safer srun termination

### LINKED ISSUE #3267: When SRUN killed, some of its ranks keep running (Frontier)
Issue (reported to OLCF):
```
when one of the SRUNs was killed, its ranks still were running on some nodes 
(didn't check all allocated nodes). How can I be sure that with terminating 
SRUN all its ranks are terminated as well? (that SRUN had 8K ranks)
```

Reply:
```
Especially with a large srun (8k nodes) it can take time for everything to 
react to the kill (and there's always a chance some node within that group 
hangs when trying to kill the processes for several of reasons).

The normal way to kill an srun is with two subsequent (within 1s of each 
other) SIGINT (i.e. Ctrl-C) commands. I wonder if you could modify your 
kill statement to operate that way (basically issue os.killpg(pgrp.signal.SIGINT), 
sleep a few milliseconds, then the os.killpg again).

Alternatively you can add

--quit-on-interrupt

to the srun command to remove the need for the second SIGINT (in which case 
you'd just change your one call from SIGKILL to SIGINT).

Finally, you could also try the scancel command, although you'd need the 
step id of the particular srun command (and you'd run something like a 
"system(scancel JOB_ID.STEP_ID)". I'm trying to see if the step id can 
be found easily from the wrapper script.

I think one of those routes may help by permitting srun to go through it's 
normal termination procedure rather that issuing the more heavy-handed SIGKILL.
```

andre-merzky:
This fixes #3267

### DIFF
--- a/src/radical/pilot/agent/executing/popen.py
+++ b/src/radical/pilot/agent/executing/popen.py
@@ -317,9 +317,17 @@ class Popen(AgentExecutingComponent):
                     # process group (which should include the actual launch
                     # method)
                     try:
-                        # kill the whole process group
+                        # kill the whole process group.
+                        # Try SIGINT first to allow signal handlers, then
+                        # SIGTERM to allow clean termination, then SIGKILL to
+                        # enforce termination.
                         pgrp = os.getpgid(task['proc'].pid)
+                        os.killpg(pgrp, signal.SIGINT)
+                        time.sleep(0.1)
+                        os.killpg(pgrp, signal.SIGTERM)
+                        time.sleep(0.1)
                         os.killpg(pgrp, signal.SIGKILL)
+
                     except OSError:
                         # lost race: task is already gone, we ignore this
                         # FIXME: collect and move to DONE/FAILED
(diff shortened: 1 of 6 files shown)


mtitov:
@andre-merzky quick check - should we add some handling of the termination signal within the launching script? do we need to preserve `launch_collect` tracing event? or such tasks will finish with `task_run_cancel_stop` only?

andre-merzky:
> @andre-merzky quick check - should we add some handling of the termination signal within the launching script? do we need to preserve `launch_collect` tracing event? or such tasks will finish with `task_run_cancel_stop` only?

A launch script handler would, IMHO, only make sense if we want to still execute `post_launch` commands - which I don't think should happen.  We can't do much about the exec script anyway, and that's where the interesting parts are.  So indeed, I would expect `task_run_cancel_stop` to be the next event in this case.

mtitov:
Thanks Andre!

mtitov (approved):
LGTM!

andre-merzky:
Oh, I need to fix the tests...  Will do ASAP.

PR #3273, opened 2024-12-03
Fix/task cancel

### LINKED ISSUE #3271: cancelation from client side not happening?

andre-merzky:
This fixes #3271, and should be merged after #3270.


andre-merzky:
Oops, wrong base

PR #3282, opened 2024-12-10
add `DOWNLOAD` staging action

### LINKED ISSUE #3266: Download input staging
In ROSE, there is a case where we want to use RP input_staging to download a file from the internet or from an endpoint.

I think we can extend input_staging and RP ACTIONS as follows:

```python
        data_deps = {'source': f"https://xxxx/xx.txt",
                     'target': f"task:///{file_name}", 'action': rp.DOWNLOAD}
```

andre-merzky:
This fixes #3266

### DIFF
--- a/examples/05_task_input_data.py
+++ b/examples/05_task_input_data.py
@@ -91,13 +91,16 @@ if __name__ == '__main__':
             # Here we don't use dict initialization.
             td = rp.TaskDescription()
             td.executable     = '/usr/bin/wc'
-            td.arguments      = ['-c', 'input.dat']
+            td.arguments      = ['-c', 'input_1.dat', 'input_2.dat']
           # td.input_staging  = ['input.dat']
 
           # this is a shortcut for:
-            td.input_staging  = {'source': 'client:///input.dat',
-                                 'target': 'task:///input.dat',
-                                 'action': rp.TRANSFER}
+            td.input_staging  = [{'source': 'client:///input.dat',
+                                  'target': 'task:///input_1.dat',
+                                  'action': rp.TRANSFER},
+                                 {'source': 'https://1.1.1.1/',
+                                  'target': 'task:///input_2.dat',
+                                  'action': rp.DOWNLOAD}]
             tds.append(td)
             report.progress()
         report.ok('>>ok\n')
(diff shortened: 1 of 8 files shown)


examples/05_task_input_data.py
```diff
@@ -91,13 +91,16 @@
...
           # td.input_staging  = ['input.dat']
 
           # this is a shortcut for:
-            td.input_staging  = {'source': 'client:///input.dat',
-                                 'target': 'task:///input.dat',
-                                 'action': rp.TRANSFER}
+            td.input_staging  = [{'source': 'client:///input.dat',
+                                  'target': 'task:///input_1.dat',
+                                  'action': rp.TRANSFER},
+                                 {'source': 'https://1.1.1.1/',
```
AymenFJA, line 101: For this line: `source': 'https://1.1.1.1/` is it missing the `input_2.dat` in the end so it can be like:
`https://1.1.1.1/input_2.dat`? Or are we expecting only the remote path without the file name in `source` and then concatenating `source` + `target` when we download or something like that?

Also, for `https://1.1.1.1/input_2.dat`, when I try to fetch the file I get a file not found from the server!

AymenFJA (commented):
I left 2 comments @andre-merzky. Thanks.

src/radical/pilot/utils/staging_helper.py
```diff
@@ -107,6 +113,11 @@ def link(self, src, tgt, flags):
         self.mkdir(os.path.dirname(tgt), flags)
         os.link(src, tgt)
 
+    def download(self, src, tgt, flags):
+        tgt = ru.Url(tgt).path
+        self.mkdir(os.path.dirname(tgt), flags)
+        ru.sh_callout('wget -r %s -O %s' % (src, tgt))
```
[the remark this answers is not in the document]
andre-merzky: Ack, we can use python for that, will change!

examples/05_task_input_data.py
AymenFJA: Got it and make sense, thanks.

AymenFJA (approved):
LGTM,  thanks @andre-merzky

PR #3289, opened 2024-12-16
target state is interpreted *after* output staging

### LINKED ISSUE #3288: output staging only completed *after* final state
See https://github.com/Parsl/parsl/issues/3722, reproducible in RP devel w/o raptor.

andre-merzky:
This fixes #3288


AymenFJA (approved):
LGTM, thanks @andre-merzky

PR #3291, opened 2024-12-17
registry dump hack

andre-merzky:
temporary fix for impeccable runs

### DIFF
--- a/src/radical/pilot/session.py
+++ b/src/radical/pilot/session.py
@@ -359,6 +359,18 @@ class Session(object):
         self._cfg.reg_addr = self._reg_service.addr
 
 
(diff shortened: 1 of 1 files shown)


mtitov:
we also would need to add the following JSONs (for TMGR and PMGR) into the watcher
https://github.com/radical-cybertools/radical.pilot/blob/8368576593017c9400c4cfa989cffc41091bbf24/src/radical/pilot/pilot_manager.py#L235-L243
https://github.com/radical-cybertools/radical.pilot/blob/70c6bc62f432a41e588c545770275bceb9abd3a1/src/radical/pilot/task_manager.py#L241-L249

PR #3295, opened 2024-12-21
trigger dump via command

andre-merzky:
This PR adds support for
```sh
radical-pilot-command rp.session.thinkie.merzky.020078.0069 dump name=foo
```
which will result in
```sh
$ ls rp.session.thinkie.merzky.020078.0069/*json
rp.session.thinkie.merzky.020078.0069/pmgr.0000.foo.json
rp.session.thinkie.merzky.020078.0069/rp.session.thinkie.merzky.020078.0069.reg.foo.json  rp.session.thinkie.merzky.020078.0069/rp.session.thinkie.merzky.020078.0069.reg.init.json  rp.session.thinkie.merzky.020078.0069/tmgr.0000.foo.json
```

If `name=foo` is ommited, the normal registry and manager dumps occur, and the session then can be used for analytics.
Note that the first argument must point to a client session sandbox - the `<SID>.reg.init.json` is used to extract the ZMQ endpoints of the `CONTROL` pubsub over which the request message is sent.


src/radical/pilot/session.py
```diff
@@ -254,11 +254,32 @@ def _init_primary(self):
                                           url=bcfg['addr_pub'],
                                           log=self._log,
                                           prof=self._prof)
+        ru.zmq.Subscriber(channel=rpc.CONTROL_PUBSUB, url=bcfg['addr_sub'],
```
mtitov, line 257: should it be assigned to `self._ctrl_sub` and then stopped (`self._ctrl_sub.stop()`) in the `close` method?

(*) and same for `self._ctrl_pub` - being stopped explicitly?

andre-merzky: Done.  The pub channel though does not need closing (and does not, in fact, have a `close()` method).

bin/radical-pilot-command
```diff
@@ -0,0 +1,47 @@
...
+
+__copyright__ = "Copyright 2024, http://radical.rutgers.edu"
+__license__   = "MIT"
+
+
+import sys
+import time
+
+import radical.utils   as ru
+import radical.pilot   as rp
```
mtitov, line 11: not used in the code

mtitov: or maybe keep it, and then use it for channel name: instead of `'control_pubsub'` use `rp.CONTROL_PUBSUB`?

andre-merzky: Thanks, I did so (use the constant)

mtitov (approved):
LGTM! Thanks!

PR #3297, opened 2025-01-14
added CI with spack manager

mtitov:


andre-merzky (commented):
thanks for taking care of this!  Two questions:
  - I don't see this being executed under `Checks`?
  - Do we need to do something to not have it time out?

andre-merzky:
> @andre-merzky
> 
> > I don't see this being executed under Checks?
> 
> I keep it scheduled (twice a week) and on-demand, so no triggers for pull requests, since changes in RP will not be reflected in spack right away
> 
> > Do we need to do something to not have it time out?
> 
> I didn't set any timeout for now, would like to keep it as it is and later we'll have an average runtime and will be able to set upper threshold

Ack on both - and sorry for the first question, because, duh!

PR #3302, opened 2025-01-22
ensure scheduler sub-process terminates

andre-merzky:


src/radical/pilot/agent/scheduler/base.py
```diff
@@ -272,6 +273,10 @@ def initialize(self):
         self._p.daemon = True
         self._p.start()
 
+        self._pwatcher = ru.PWatcher(uid='%s.pw' % self._uid, log=self._log)
+        self._pwatcher.watch(os.getpid())
+        self._pwatcher.watch(self._p.pid)
+
```
mtitov, line 279: quick check - do we need this part considering that the base component already tracks itself?

mtitov (commented):
General question: should we have "our" `mp.Process` (within `radical.utils`) that will include `PWatcher`? So, whenever we create another process using `Process` we don't need to have an extra step to initiate `PWatcher` in it

src/radical/pilot/agent/scheduler/base.py
andre-merzky: It's not about watching itself (that's not really possible - if `self` goes away, so does the watcher. Instead, this is tying the self process to the lifetime of `self._p` - if the latter dies, self will also be killed.

andre-merzky:
> General question: should we have "our" `mp.Process` (within `radical.utils`) that will include `PWatcher`? So, whenever we create another process using `Process` we don't need to have an extra step to initiate `PWatcher` in it

Possibly - but for now I think it is beneficial to have the watchers explicitly in the code.  You need some of those anyway, as we watch both directions: parent watches child and vice versa.  Only the first part could easily be put into a Process class - the second would require some magic wrapping of the child process - not worth the effort IMHO.

src/radical/pilot/agent/scheduler/base.py
mtitov: ok, got it! thanks!

mtitov (approved):
LGTM! Thanks!

PR #3304, opened 2025-01-27
fixed task handling when cancelled

mtitov:
when task is cancelled it shouldn't be handled within the `Popen._check_running` method

(*) revert back this change https://github.com/radical-cybertools/radical.pilot/commit/2e6f18653f69059fe5db846639812347e0ad8956


andre-merzky (approved):
my bad, thanks!

PR #3307, opened 2025-01-30
Extend functionality for `startup_timeout` (TaskDescription)

mtitov:
@andre-merzky would the following approach work? can we publish a command from task exec script (using pub_addr `RP_CONTROL_PUB_ADDRESS`)

- `if task['description'].get('startup_timeout', 0.)`
    - then add publishing command for rank 0 (`'cmd': 'startup_done'`) into `...exec.sh` after `exec_start`


src/radical/pilot/agent/executing/base.py
```diff
@@ -158,51 +170,51 @@ def _to_watcher(self):
...
     # --------------------------------------------------------------------------
     #
     def handle_timeout(self, task):
 
-        to = task['description'].get('timeout', 0.0)
+        startup_to = task['description'].get('startup_timeout', 0.)
+        exec_to    = task['description'].get('timeout',         0.)
 
-        if to > 0.0:
+        if startup_to > 0. or exec_to > 0.:
```
andre-merzky, line 213: `if startup_to or exec_to:` ?

src/radical/pilot/agent/executing/popen.py
```diff
@@ -68,7 +68,8 @@ def initialize(self):
     #
     def get_task(self, tid):
 
-        return self._tasks.get(tid)
+        with self._check_lock:
```
andre-merzky, line 71: I don't understand what the lock is achieving here, the `.get()` should be safe either way? If the task was removed just before the lock is set, the result is the same.

src/radical/pilot/task_description.py
```diff
@@ -426,13 +425,17 @@ class TaskDescription(ru.TypedDict):
...
+            starting from the launch time, if startup_timeout is not set, or
+            starting from execution start time, if startup_timeout is set and
+            was tracked. The task will then end up in `CANCELED` state.
+
+        startup_timeout (float, optional): Any value larger than 0 will abort
+            the task process if it hasn't started the actual execution process,
+            e.g., batch system didn't provide allocated resources or the service
+            startup takes too long. This option is very important for service
+            tasks. The service task will end up in `FAILED` state, regular task
+            will end up in `CANCELED` state.
```
andre-merzky, line 438: Maybe we should always move to `CANCELED` on timeout for better consistency?

andre-merzky (commented):
Overall this looks like an reasonable approach to me.  The signal sending part is missing though?

src/radical/pilot/agent/executing/base.py
```diff
@@ -158,51 +170,51 @@ def _to_watcher(self):
...
             # avoid busy wait
-            if not to_list:
+            if not to_tasks:
                 time.sleep(1)
                 continue
 
-            # sort by timeout, smallest first
-            to_list.sort(key=lambda x: x[1])
-
+            to_list = sorted(to_tasks.values(), key=lambda x: x[1])
```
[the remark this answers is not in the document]
mtitov: right, brought it back (had several changes back and forth, and left that comment behind)

src/radical/pilot/agent/executing/base.py
```diff
@@ -158,51 +170,51 @@ def _to_watcher(self):
...
             # cancel all tasks which have timed out
-            offset = 0
             for task, cancel_time in to_list:
                 now = time.time()
                 if now > cancel_time:
-                    self._prof.prof('task_timeout', uid=task['uid'])
-                    self.cancel_task(task=task)
-                    offset += 1
+                    if cancel_time:
```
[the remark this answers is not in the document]
mtitov: we've already disccused this part, keeping it here for records - use `cancel_time` equals to `0.` to reset cancellation after `td.startup_timeout` cleared and there is no `td.timeout` (i.e., execution timeout)

src/radical/pilot/agent/executing/base.py
mtitov: Updated accordingly

src/radical/pilot/agent/executing/popen.py
mtitov: yeah, agree, brought it back

src/radical/pilot/task_description.py
mtitov: with service tasks it's a little different - while task is failed (killing service process), Agent_0 raises `RuntimeError`, thus agent terminates itself. Should I keep description as it is, but adding info about agent?

andre-merzky: Good point.  I would argue that the agent should not fail on service failures - those are reported to the application after all and can be acted upon.  That would be out of scope for this PR though - should we open an issue for that?

mtitov:
@andre-merzky latest fix was tested on Frontier, communication is set between tasks and agent, so agent received all `task_startup_done` commands which were sent

andre-merzky:
Sorry for the delay!

andre-merzky (approved):

PR #3308, opened 2025-02-02
MPIEXEC: rankfile collides with hostnames

andre-merzky:
On delta, `mpiexec` does not like a rankfile specified and at the same time `-H` being used.  And in fact the information seems redundant as the hostnames are listed in the rankfile anyway.


mtitov (approved):
LGTM!

p.s., we actually didn't have a use case to test `-rf` related options, thus this works perfectly
p.p.s., our tests started to fail since Jan 30.., couldn't find what has changed since then (something on GitHub side)

PR #3314, opened 2025-02-19
make sure tasks tmpdir exists

andre-merzky:


src/radical/pilot/agent/executing/base.py
```diff
@@ -91,8 +91,10 @@ def initialize(self):
         self.prof      = '$RP_PILOT_SANDBOX/prof'
 
         # if so configured, let the tasks know what to use as tmp dir
-        self._task_tmp = self.session.rcfg.get('task_tmp',
-                                               os.environ.get('TMP', '/tmp'))
+        self.tmpdir    = os.environ.get('TMPDIR', '/tmp') + '/' + self.sid
```
AymenFJA, line 94: Some minor suggestions:
```python
self.tmpdir = os.environ.get('TMPDIR', '/tmp')
self.tmpdir = os.path.join(tmpdir, self.sid)
```

Besides that, looks good to me.

AymenFJA (approved):

mtitov (approved):
LGTM!

src/radical/pilot/agent/executing/base.py
andre-merzky: Well, `/tmp` in that code is also hardcoding a slash.  So one would need

```py
self.tmpdir = os.environ.get('TMPDIR', os.path.join(os.path.abspath(os.sep), 'tmp'))
self.tmpdir = os.path.join(self.tmpdir, self.sid)
```

Do we want this? ;-)

PS.: Mikhail asked me: do you want to beat Aymen with his own weapons :-P

PR #3317, opened 2025-02-25
allow oversubscribe on localhost

andre-merzky:
The raptor_mpi tests hung because the GH CI VM did not have sufficient resources.  This PR now allows to oversubscribe resources on `local.localhost` (at least with `mpirun`, to be extended) to resolve this.


mtitov (approved):
LGTM! Thank you!

PR #3319, opened 2025-03-04
Fix/double schedule

andre-merzky:


mtitov (approved):
LGTM!

p.s., and would be good to merge it before `fix/cancel`

PR #3322, opened 2025-03-06
Fix task cancellation problem where (a) cancel raced in the executor and scheduler and (b) cancellation updates were not forwarded

andre-merzky:


src/radical/pilot/utils/component.py
```diff
@@ -1263,7 +1268,10 @@ class AgentComponent(BaseComponent):
     def advance(self, things, state=None, publish=True, push=False, qname=None,
                       ts=None, fwd=True, prof=True):
 
+        things = ru.as_list(things)
```
mtitov, line 1271: this one is a duplication, we convert it into list in L1278

src/radical/pilot/task.py
```diff
@@ -157,6 +157,18 @@ def _update(self, task_dict, reconnect=False):
...
+        if current in [rps.FAILED, rps.DONE]:
+            self._log.debug('task %s is final, ignore update', self.uid)
+            return
+
+        # when in CANCELED state, we only allow updates for `DONE` tasks - in
+        # that case the cancel command raced the task execution, and the
+        # execution actually won, so we don't want to waste that work
+        if current == rps.CANCELED and target != rps.DONE:
+            self._log.debug('task %s was CANCELED, ignore update', self.uid)
+            return
```
mtitov, line 170: should we still make an update? we need to ensure that if there are slots, we need to update them (probably some other attributes as well)

src/radical/pilot/utils/component.py
```diff
@@ -1090,8 +1103,8 @@ def advance(self, things, state=None, publish=True, push=False, qname=None,
 
             uid = thing['uid']
 
-          # if thing['type'] not in ['task', 'pilot']:
-          #     raise TypeError("thing has unknown type (%s)" % uid)
+            if thing['type'] not in ['task', 'pilot']:
+                raise TypeError("thing has unknown type (%s)" % uid)
```
mtitov, line 1107: we also have `'service_task'` as type (in agent_0)

mtitov:
@andre-merzky regarding tests - analytics keep failing because there are no bootstrap profile events..
```
% cat bootstrap_0.prof 
#time,event,comp,thread,uid,state,msg
1741810951.934896,sync_abs,bootstrap_0,MainThread,pilot.0000,PMGR_ACTIVE_PENDING,fv-az1694-632:10.1.0.63:1741810951.934896:1741810951.934896:1741810951.934896
```

src/radical/pilot/utils/component.py
andre-merzky: Thanks, cleaned up.

src/radical/pilot/task.py
andre-merzky: Ack, done.

mtitov (approved):
LGTM! Thank you!!

PR #3323, opened 2025-03-12
Fix/install

andre-merzky:
I am not sure when we lost installing that file...

### DIFF
--- a/src/radical/pilot/utils/prof_utils.py
+++ b/src/radical/pilot/utils/prof_utils.py
@@ -492,8 +492,7 @@ def get_session_profile(sid, src=None):
 
     if os.path.exists(src):
         # we have profiles locally
-        profiles  = glob.glob('%s/*.prof'    % src)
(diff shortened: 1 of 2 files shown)


AymenFJA (approved):
LGTM. Thanks.

src/radical/pilot/utils/prof_utils.py
```diff
@@ -493,7 +493,7 @@ def get_session_profile(sid, src=None):
     if os.path.exists(src):
         # we have profiles locally
         profiles  = glob.glob('%s/*.prof'    % src)
-        profiles += glob.glob('%s/**/*.prof' % src)
+        profiles += glob.glob('%s/**/*.prof' % src, recursive=True)
```
mtitov, line 496: the updated command will include the one above, right? so we can just keep `profiles = glob.glob('%s/**/*.prof' % src, recursive=True)`

andre-merzky: You are right - I actually expected that *not* to be the case... :-P  Fixed!

PR #3324, opened 2025-03-12
make it simpler to assign session IDs to experiments.

andre-merzky:
Example:

    export RP_SID_BASE=flux_256nodes_64ktasks_2partitions
    ./my_experiment.py
    ./my_experiment.py

will result in these two session IDs:
    flux_256nodes_64ktasks_2partitions.0000
    flux_256nodes_64ktasks_2partitions.0001

etc, thus making sessions easily recognicable and easy to match to experiments.

### DIFF
--- a/src/radical/pilot/session.py
+++ b/src/radical/pilot/session.py
@@ -229,7 +229,8 @@ class Session(object):
 
         # if user did not set a uid, we need to generate a new ID
         if not self._uid:
-            self._uid = ru.generate_id('rp.session', mode=ru.ID_PRIVATE)
+            base = os.environ.get('RP_SID_BASE', 'rp.session')
+            self._uid = ru.generate_id(base, mode=ru.ID_PRIVATE)
 
         # we still call `_init_cfg` to complete missing config settings
         # FIXME: completion only needed by `PRIMARY`


src/radical/pilot/session.py
```diff
@@ -229,7 +229,8 @@ def _init_primary(self):
 
         # if user did not set a uid, we need to generate a new ID
         if not self._uid:
-            self._uid = ru.generate_id('rp.session', mode=ru.ID_PRIVATE)
+            base = os.environ.get('RP_SID_BASE', 'rp.session')
```
mtitov, line 232: quick check on naming - env variables which start with `RP_` are internal ones, but the ones that start with `RADICAL_PILOT_` are set by users (aka external) - is this one an internal variable?

p.s., one more thing - `BASE` -> `PREFIX`?

andre-merzky: Ack on the var name, will change.  I think `BASE` is ok as it is not a prefix to the 'normal' session ID, but a completely different base name.

andre-merzky: Thanks - I also fixes the actual `generate_uid` call - that embarrassingly did not yet yield what the doc string promised...

mtitov (approved):
LGTM!

PR #3326, opened 2025-03-17
Fix/rf option

andre-merzky:
fix detection of rankfile support for newer openmpi versions


mtitov (approved):
LGTM! 

p.s., just need to fix tests

PR #3334, opened 2025-03-31
move python version to 3.8

andre-merzky:


mtitov:
as part of this update I would propose to remove these pieces in CI configs (initially these parts were set to use 3.7, and we do have 3.8 within the list, no need to have it here as well)

https://github.com/radical-cybertools/radical.pilot/blob/dbe53148244ef9fe7a039f65e843b72f70b4cc62/.github/workflows/ci-release.yml#L17-L19

https://github.com/radical-cybertools/radical.pilot/blob/dbe53148244ef9fe7a039f65e843b72f70b4cc62/.github/workflows/ci.yml#L53-L55

andre-merzky:
> as part of this update I would propose to remove these pieces in CI configs (initially these parts were set to use 3.7, and we do have 3.8 within the list, no need to have it here as well)

Ack, done.

mtitov (approved):
LGTM!

PR #3337, opened 2025-04-02
pass network configuration through to RU

andre-merzky:
This accompanies https://github.com/radical-cybertools/radical.utils/pull/444 and allows to set the default network interface per RP's resource configs.


mtitov (approved):
LGTM! Thank you!

PR #3344, opened 2025-04-29
service task must also have a final/ready state

AymenFJA:
My suggestion for `service` tasks in RP is that it should have a state indicating that the service is ready to accept requests.

This is aligned with the definition of service.

### DIFF
--- a/src/radical/pilot/agent/agent_0.py
+++ b/src/radical/pilot/agent/agent_0.py
@@ -435,6 +435,8 @@ class Agent_0(rpu.AgentComponent):
                     raise RuntimeError('Unable to start service')
             else:
                 self._service_start_evt.wait()
+            
+            self.advance(task, rps.SERVICE_ACTIVE, publish=True, push=False)
 
             info = self._reg.get('services.%s' % td.uid)
             self._log.info('agent service started: %s - %s', td.uid, info)
--- a/src/radical/pilot/states.py
+++ b/src/radical/pilot/states.py
@@ -12,10 +12,11 @@ NEW      = 'NEW'
 DONE     = 'DONE'
 FAILED   = 'FAILED'
 CANCELED = 'CANCELED'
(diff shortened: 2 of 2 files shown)


AymenFJA:
This will require: 

```shell
rose@feature/hacakthon
radical.flow@feature/hackathon
radical.pilot@feature/hackathon

```

andre-merzky (commented):
Aymen - I will have to contradict myself: during the call I assumed that the `SERVICE_READY` state lives in flow.  In RP, I am a bit hesitant to embrace it, for several reasons:

As discussed, it cannot be a final state in RP.

The service task execution is a bit shoehorned into the agent at the moment.  Specifically, it stalls the agent while the service is starting up, precisely because it waits for the service to get ready.  That will create problems sooner or later, and I do not yet know how we are going to resolve this, and if and how we are able to ensure that state transition.

RP uses the same state model for *all* task types - so having a state which only applies to service tasks will break some pieces which ensure correct state transition sequence for all tasks.

Would you be able to park that functionality in Flow for the time being?

mtitov:
@andre-merzky maybe we can have a general state `AGENT_RUNNING`, which will mean that the task executable is actually running - it will be applicable to all tasks, but at this moment we will activate it for service tasks only (`AGENT_EXECUTING` and `AGENT_RUNNING` will be overlapping)

mtitov:
@AymenFJA @andre-merzky just to confirm that we agreed to not create a new state (`SERVICE_ACTIVE`) but to rely on published commands as `service_up` and corresponding info data.

@AymenFJA can put here a link to a corresponding commit at `radical.async` when you'll add a callback that Andre proposed to use as a solution for service tasks? (just to keep track of how it was resolved)

AymenFJA:
A workaround was introduced in this PR: https://github.com/radical-cybertools/radical.flow/commit/c831429a65d0e963f294f53d59e9afb2f7fda699

Issue #3011, opened 2023-08-14
Misleading RAPTOR message pyfuncs and suppressed error message.


AymenFJA:
The RAPTOR code, specifically in https://github.com/radical-cybertools/radical.pilot/blob/9190fa38783191167156365ac2e7bcd508ab2c15/src/radical/pilot/raptor/worker.py#L358

If this line fails, our RAPTOR logger will print: 
https://github.com/radical-cybertools/radical.pilot/blob/9190fa38783191167156365ac2e7bcd508ab2c15/src/radical/pilot/raptor/worker.py#L361

Two problems here: 

1. The message of the `get_func_attr` is suppressed, and we just report a hard-coded message (we need to at least report the correct message, not the entire exception)
2. I tested it with a Python function with some module dependencies that were unavailable in the namespace where the function was deserialized, and I printed the exception. Something like this:
```python
except Exception as e:
          self._log.warn('function is not a PythonTask [%s] ', uid)
          self._log.warn(e)
```

This is the modified output message from our `raptor.worker`: 
```
1691803277.803 : master.000000.worker.0000 : 28773 : 140004894422784 : WARNING  : function is not a PythonTask [task.000000]
1691803277.803 : master.000000.worker.0000 : 28773 : 140004894422784 : WARNING  : No module named 'mpi_funcs'
1691803277.804 : master.000000.worker.0000 : 28777 : 139872066201344 : DEBUG    : orig args: [] : {}
```

As you can see, the first line is what RAPTOR reported and marked this function as not `PythonTask` which is not correct as this task was serialized by our function serializer. 

The 2nd line shows the actual error of why our serialization failed, and that is because `No module named 'mpi_funcs'` which my function requires during the execution.

PR #3361, opened 2025-06-14
catch deserialization error and fail the task

AymenFJA:
This PR addresses RAPTOR issue #3011.

When a task fails on a `deserialization` error, we get a `Not a PythonTask` error. The problem is that the error is overloaded; it can be true sometimes, and sometimes it can be due to an internal deserialization failure that we couldn't catch. What makes it harder is that there is no specific `deserialization` exception from `dill` that we can catch, for example, it can raise `ModuleNotFoundError` or so. To solve this issue, we create standard classes and we wrap any Exception generated by `dill` during `serialization` and `deserialization` to control what to do next, for example, `fail the task` instead of hanging forever.

This issue was raised again when building integration tests for RADICAL-Async with RAPTOR.
This PR does that.

Previous behaviour:
```
function is not a PythonTask [task.0002]
```

Current behaviour:

```
1749869376.621 : flow.worker.000000   : 3988586 : 127041735206592 : WARNING  : failed to deserialize function for [task.0002]: Failed to deserialize object from data: No module named 'tests'. 

Task returns as FAILED
```

### DIFF
--- a/src/radical/pilot/raptor/worker.py
+++ b/src/radical/pilot/raptor/worker.py
@@ -388,7 +388,7 @@ class Worker(object):
         # check if we have a serialized object
         if not to_call:
             self._log.debug('func serialized: %d: %s', len(func), func)
-            
+
             try:
                 to_call, _args, _kwargs = PythonTask.get_func_attr(func)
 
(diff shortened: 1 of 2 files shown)


andre-merzky (approved):
This all looks rather reasonable to me.  Thanks!

PR #3369, opened 2025-08-20
check if proxy is still alive before closing

andre-merzky:


andre-merzky:
@mtitov : ping for review.

PR #3379, opened 2025-10-08
complement missing events in event pairs

andre-merzky:


mtitov (approved):
LGTM! Thank you!
(we don't need to polish this code at this moment, but in general, some of the recommendations from co-pilot are reasonable)

PR #3390, opened 2025-11-28
fixed LM Flux

mtitov:


src/radical/pilot/agent/launch_method/flux.py
```diff
@@ -299,11 +300,12 @@ def _part_thread(self, part_id, nodes, q_in, q_out):
 
                 self._log.debug('%s: submitted %d tasks', part.uid, len(tasks))
 
-            except Exception:
-                self._log.exception('LM flux submit failed')
+            except Exception as e:
+                self._log.exception(f'LM flux submit failed: {e}')
```
andre-merzky: `log.exception` logs the message anyway, so this is redundant.

andre-merzky (approved):
Looks good to me, thanks!  I left a comment about a log message, happy to leave it to your discretion if you want to change that.

PR #3391, opened 2026-01-16
fixed LM preparation for RM

mtitov:
Failed LM initialization should raise an exception, so RM will skip it

https://github.com/radical-cybertools/radical.pilot/blob/03b139c2e98c21d4199f5508d9626d241c3a1745/src/radical/pilot/agent/resource_manager/base.py#L431-L441


andre-merzky (approved):
LGTM, thanks!

